PR #5, opened 2017-11-30
Feature/add debounce throttle utils

AndyOGo:
Adds `debounce` and `throttle` helper function.

Maybe we should consider to optionally utilize `requestAnimationFrame` instead off `setTimeout`. 🤔

### DIFF
--- a/src/js/throttle.js
+++ b/src/js/throttle.js
@@ -1,0 +1,48 @@
+import debounce from './debounce';
+
+/**
+ * Creates a throttled function that only invokes `func` at most once per
+ * every `wait` milliseconds. The throttled function comes with a `cancel`
+ * method to cancel delayed `func` invocations and a `flush` method to
+ * immediately invoke them. Provide an options object to indicate whether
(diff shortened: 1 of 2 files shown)


src/js/throttle.js
```diff
@@ -0,0 +1,48 @@
...
+ * @param {DebounceFunc} func - The function which will be throttled.
+ * @param {number} [wait=0] - The time in milliseconds the execution of this function will be deferred.
+ * @param {object} [options] - The options object.
+ * @param {boolean} [options.leading=true] - The function is executed at the beginning (first call).
+ * @param {boolean} [options.trailing=true] - The function will be executed at the end (last call, maybe deferred).
+ * @returns {DebouncedFunction} Returns the new throttled function.
+ * @example
+ *
+ * // Avoid excessively updating the position while scrolling.
+ * jQuery(window).on('scroll', _.throttle(updatePosition, 100));
```
LucaMele: ? jQuery?

LucaMele:
I would start easy and just du a debounce that clears itself if called before finishing

PR #8, opened 2017-11-30
start improving mobile layout

AndyOGo:
improves look and feel of footer on small mobiles.

# Todo:

- [x] height set to zero by JS needs to be ineffective on screens greater `xs`
- [ ] ~~ideally dropdown js would only be executed at `xs`~~ -> tracked in #15 
- [x] use variables

### DIFF
--- a/src/components/m-dropdown/mixins/dropdown.scss
+++ b/src/components/m-dropdown/mixins/dropdown.scss
@@ -24,7 +24,7 @@
   height: $size;
   margin-top: -($size / 2);
 
-  transition: all 0.3s ease;
+  transition: transform 0.3s ease;
 
   .is-open & {
     transform: rotate(180deg);
--- a/src/components/m-dropdown/mixins/dropdown.scss
+++ b/src/components/m-dropdown/mixins/dropdown.scss
@@ -1,0 +1,44 @@
+@import "../settings/dropdown";
+
+@mixin dropdown-toggle() {
+  position: relative;
+
+  &:hover,
+  &:active,
+  &:focus {
+    cursor: pointer;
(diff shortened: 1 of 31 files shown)


src/components/m-dropdown/mixins/dropdown.scss
```diff
@@ -0,0 +1,52 @@
...
+  &:hover,
+  &:active,
+  &:focus {
+    cursor: pointer;
+  }
+}
+
+@mixin dropdown-icon($size: $dropdown__icon-size, $right: $dropdown__toggle-x-padding) {
+  position: absolute;
+  top: 50%;
```
LucaMele: could use calc instead of top:50% and margin-top -(size/2). Much cleaner with calc

src/components/m-dropdown/mixins/dropdown.scss
```diff
@@ -0,0 +1,52 @@
...
+  right: $right;
+
+  display: block;
+  width: $size;
+  height: $size;
+  margin-top: -($size / 2);
+
+  transition: all 0.3s ease;
```
LucaMele: please dont use all

src/components/m-footer-legals/index.scss
```diff
@@ -8,44 +8,94 @@
...
+  // delimiter
+  &::after {
+    position: absolute;
+    left: 0;
+    top: 50%;
+
+    display: block;
+    width: $footer-delimiter-width;
+    height: $footer-delimiter-height;
+    margin-top: ($footer-delimiter-height / -2);
```
LucaMele: ditto

PR #21, opened 2017-12-04
moved button from atom to molecule

### LINKED ISSUE #19: Button should be a molecule as it includes an atom

AndyOGo:
fixes #19


src/components/m-button/index.js
```diff
@@ -8,6 +8,6 @@ class Button extends BaseComponentGlobal {
   }
 }
 
-window.customElements.define('axa-button', Button);
+window.customElements.define('axm-button', Button);
```
LucaMele: attention here

PR #24, opened 2017-12-04
Feature/add device state detection

### LINKED ISSUE #15: JS is unaware of active breakpoint
We should be able to detect the current breakpoint even in JS, and also changes of it.

- define breakpoints at one place - in **CSS** (single source of truth)
- add custom `breakpoint:change` event

Should work like it is described here https://davidwalsh.name/device-state-detection-css-media-queries-javascript

AndyOGo:
fixes #15 

# Todo

- [x] implement single node device state magic
- [x] update drop down menu at mobile footer
- [x] implement `pub/sub` module
- [x] add `CustomEvent` polyfill

### DIFF
--- a/src/components/a-device-state/index.scss
+++ b/src/components/a-device-state/index.scss
@@ -25,7 +25,7 @@
       @include respond-landscape() {
         @include respond-pixel-ratio($pixel-ratio) {
           @include respond-up($breakpoint-key) {
-            content: 'breakpoint:#{$breakpoint-key}, orientation:portrait, pixel-ratio:#{$pixel-ratio}';
+            content: 'breakpoint:#{$breakpoint-key}, orientation:landscape, pixel-ratio:#{$pixel-ratio}';
           }
         }
       }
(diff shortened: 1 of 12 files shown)


src/components/a-device-state/index.scss
```diff
@@ -0,0 +1,50 @@
...
+@import "../../styles/settings/grids";
+@import "../../styles/settings/pixel-ratios";
+@import "../../styles/settings/zindex";
+@import "../../styles/mixins/respond";
+
+.a-device-state::after {
+  $breakpoints-keys: map-keys($grid-breakpoints);
+
+  position: absolute;
+  top: -999em;
```
LucaMele: em?

AndyOGo: true, `px` will work too

AndyOGo:
@LucaMele 
I added a polyfill for `CustomEvents` and according to SO is really need for IE 11 too:
// https://stackoverflow.com/questions/23349191/event-preventdefault-is-not-working-in-ie-11-for-custom-events

PR #30, opened 2017-12-05
FF fix

LucaMele:


src/app/app.js
```diff
@@ -64,19 +64,22 @@ document.addEventListener('DOMContentLoaded', () => {
...
-    let lastEnabled = null;
-    const toggleButtons = section.querySelectorAll('.js--toggle');
-    const arr = Array.from(toggleButtons);
-    [lastEnabled] = arr;
-    arr.forEach((button) => {
-      button.addEventListener('click', () => {
-        disable(lastEnabled, section);
-        enable(button, section);
-        lastEnabled = button;
+  // let the components render first, thats why this hack
```
AndyOGo: what about using event delegation here instead? by `on` module

src/components/a-checkbox/index.js
```diff
@@ -145,4 +145,6 @@ class Checkbox extends BaseComponentGlobal {
   }
 }
 
-window.customElements.define('axa-checkbox', Checkbox);
+document.addEventListener('DOMContentLoaded', () => {
```
AndyOGo: I think it's time for a `ready` function

LucaMele:
I cant find any reference for the workaround. Logically make sense that dom should be ready before defining a new custom element. I dont have time to analyse now why ff behaves like that. For me, make sense to wait till dom is loaded so this is good enough

src/components/a-checkbox/index.js
LucaMele: Will do

src/app/app.js
LucaMele: i think is ok with settimeout as its not critical code. Its just executed before the Components cause their DOM Ready listener is later in the queue. I can comment it better thought so is clear why

LucaMele:
yes we are

PR #41, opened 2017-12-06
Bugfix/deterministic lifecycle order

AndyOGo:
Hierarchical Lifecycle methods call is **NOT** deterministic in sense of markup structure test case.
See #42


AndyOGo:
Just observing behaivior of a failing test. can be closed.

PR #43, opened 2017-12-06
adding publish of enabled context

### LINKED ISSUE #42: Hierachical Webcomponents Lifecycle order is not deterministic
**Edit:**
~~Markup should determine hierarchical components bootstrap order, but it **doesn't**!~~~
The spec allows to bootstrap components as they are defined, independent of the markup's structure. Though in the following case I need this order to be deterministic in sense of markup structure, not definition order.
 
Given following example I expect:

```html
<axa-parent>
  <axa-child></axa-child>
</axa-parent>
```

**Expected:**
```js
Parent -> constructor
Parent > connectedCallback
Child -> constructor
Child > connectedCallback
```

Instead the definition of a web-component dictates the order of it's bootstrapping process.
**Observed:**
```js
index.js:579 Child -> constructor
index.js:588 Child > connectedCallback
index.js:542 Parent -> constructor
index.js:551 Parent > connectedCallback
```

screen shot 2017-12-06 at 14 06 04

AndyOGo:
tries to fix #42 

# Todo

- [x] published messages without subscriptions need to be queued
- [x] published context enabled message

# Fails

~~This fails in scoped local variable sharing... it's build/bundle issue tracked in~~ #44
Mitigated by using events, which needs more memory the more often `pub/sub` module is redundantly loaded.

### DIFF
--- a/src/components/_abstract/component-types.js
+++ b/src/components/_abstract/component-types.js
@@ -1,5 +1,5 @@
 import getAttributes from '../../js/get-attributes';
-import { publish } from '../../js/pubsub';
+import { publish, subscribe } from '../../js/pubsub';
 
 const memory = {};
 
@@ -12,6 +12,8 @@ const THROWED_ERROR = 'throwed';
 export class BaseComponent extends HTMLElement {
   constructor(styles = '', template) {
     super();
+
+    this._makeContextReady = this._makeContextReady.bind(this);
     this._initialise(styles, template);
   }
 
@@ -35,6 +37,21 @@ export class BaseComponent extends HTMLElement {
   connectedCallback() {
     this._appendStyles();
     this.render();
+
+    if (this.contextCallback) {
+      this._makeContextReady();
+    }
+  }
+
+  /**
+   * disconnectedCallback - description
+   *
+   * @return {type}  description
+   */
+  disconnectedCallback() {
+    if (this.unContextEnabled) {
+      this.unContextEnabled();
+    }
   }
   /**
    * _appendStyles - description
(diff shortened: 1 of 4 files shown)


AndyOGo:
This is impossible to implement cleanly cause of redundant modules shadowing local variables #44

src/js/pubsub.js
```diff
@@ -2,7 +2,12 @@ import fire from './fire';
 import on from './on';
 import debounce from './debounce';
 
-const cache = {};
+// @TODO: this local variable isn't shared between redundant module instance
```
AndyOGo: @LucaMele 
I have tracked this issue in #44

LucaMele: Hello Andy, thank you. I dont think thought that local variables should be shared as its a closed module/component. If we would have shadow dom, we would have the same thing.

LucaMele: It have to be thourgh an external module/window

LucaMele: we need to invent a mechanism for this. I consider it as a feature.

LucaMele: Best having like a global module. Similar how the app.js works, just not for the SG but for the WCs

AndyOGo: @LucaMele 
I'm not sure if we talk about the same thing.
We have module **A** with a local variable `localVar`.

If that module is bundled redundantly into various components, instead of a shared `vendor` bundle, then multiple instance of module **A** exist, each with their own scope and own local variables, without access to each other.

LucaMele: Exactly. This is not as issue but just a limitation of having completely self contained modules. It makes absolutely sense as a single module should work always on its own (holistic component) and can be integrated in a System over connections. 

The concept is different from having one bundle with everything. Here we have self contained components. That means, no sharing.

We need to have a kind of vendor module/file where other modules can share properties. This would be a nice feature. We need to discuss it.

LucaMele: So all those vars should be living in one external file so they wont be multiple bundled when integrating them in a bundled system

LucaMele: and that system should be avaiable inside the components. Hope it make sense what im saying.

src/components/_abstract/component-types.js
```diff
@@ -115,6 +133,19 @@ export class BaseComponent extends HTMLElement {
...
   enableContext() {
     this.__isContext = true;
+
+    // publish context/enabled with contextual node name
+    publish('context/enabled', this.nodeName.toLowerCase());
+  }
+
+  _makeContextReady() {
+    if (this.contextNode) {
+      setTimeout(() => {
```
LucaMele: please clear it before

PR #59, opened 2017-12-13
Feature/dev helps

LucaMele:


stack/tasks/create-a-m-o.js
```diff
@@ -0,0 +1,234 @@
...
+  3 for ORGANISM 📙
+    `,
+  );
+};
+
+const writeIndexJs = (path, _name) => {
+  fs.writeFileSync(
+    `${path}/index.js`,
+    `import { BaseComponentGlobal } from '../_abstract/component-types';
+// import the styles used for this component
```
AndyOGo: maybe we can find a nice solution with template tag function to handle leading whitespace

AndyOGo: ohooo
https://www.npmjs.com/package/outdent

PR #109, opened 2017-12-20
refactored all templates

### LINKED ISSUE #107: all templates second arg isn't children but rather a fragment object
To be semantically correct, all out templates receive a `DocumentFragment` object, not children per se. In fact all child nodes can be accessed by `fragment.children`.

So I would change all templates to reflect this detail and make them semantically correct.

Another option to access the real children would be:

```js
export default (props, { children }) => bel`<div></div>`
```

AndyOGo:
fixes #107 

@LucaMele 
What do think about this change, regarding variable naming. I think it should say that this object are the `children` of Web Components root node and that they are of type `DocumentFragment`.
So I went for `childrenFragment`, though just `fragment` or `children` as before would be shorter...

### DIFF
--- a/src/components/_abstract/component-types.js
+++ b/src/components/_abstract/component-types.js
@@ -87,13 +87,13 @@ export class BaseComponent extends HTMLElement {
 
     if (template) {
       try {
-        const children = document.createDocumentFragment();
+        const childrenFragment = document.createDocumentFragment();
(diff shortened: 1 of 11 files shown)


AndyOGo:
@LucaMele 
I fixed some conflicts and rebased.

PR #113, opened 2017-12-21
fixed generating nav for every level

### LINKED ISSUE #112: Fix Header semantic issues
- [ ] `<nav>` is repeated in mobile menu

AndyOGo:
fixes #112 
improves #114 

@LucaMele 
I just found one semantic issue.
May I ask you to double check?

### DIFF
--- a/src/components/a-link/mixins/_link.scss
+++ b/src/components/a-link/mixins/_link.scss
@@ -13,15 +13,19 @@
 }
 
 @mixin link-color($style: default) {
-  $color-link: get-color($link-colors-map, $style, text);
-
-  color: $color-link;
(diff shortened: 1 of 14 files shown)


LucaMele:
ok i think im not understading why. Where is the semantic issue in having a nav?

AndyOGo:
@LucaMele 
the issue isn't the `<nav>` tag per se, but the fact that it is repeated recursively in this case.
It should rather be used as a wrapper for the menu at a whole, not for each sub menu.

LucaMele:
Ah if its repeated yes is not correct. But can be fixed by wrapping chilren witth an article or section then is valid again

Issue #101, opened 2017-12-19
Chrome render bug


AndyOGo:
That's really weird 😱 

<img width="932" alt="screen shot 2017-12-19 at 21 00 07" src="https://user-images.githubusercontent.com/914443/34176267-a786bdf0-e4ff-11e7-912f-b4e2dce0b5a5.png">

**How I can reproduce it**
![chrome-render-bug](https://user-images.githubusercontent.com/914443/34176276-b6189a32-e4ff-11e7-963f-e6e331623bc1.gif)

AndyOGo:
A dejavu at mobile too???:
![mobile-navi-deja-vu](https://user-images.githubusercontent.com/914443/34835910-94591c88-f6f6-11e7-95f6-60d6562dfbce.gif)

AndyOGo:
Hardware Acceleration Issues in Chrome seems to be a hot topic:
- https://productforums.google.com/forum/#!topic/chrome/qtWSyiqOIz8
- https://www.reddit.com/r/chrome/comments/6qppi8/chrome_hardware_acceleration_issues/
- https://www.tonymacx86.com/threads/fix-chrome-graphics-issues-but-without-turning-off-graphics-acceleration.221328/

AndyOGo:
We are affected by following scenarious:
- toggling a hardware-accelarted parent node's `display` between `none` and e.g. `block`
- positioning sibling nodes offscreen by `position: fixed`

AndyOGo:
Woou and they are still interactive, which means `disconnectedCallback` was not triggered 😱 
![dropdown-still-there](https://user-images.githubusercontent.com/914443/36670588-1be733da-1af9-11e8-875a-37ffca04abae.gif)

AndyOGo:
yeah, I haven't seen this a long time

PR #164, opened 2018-01-11
fix stroke animation for issimple menu

### LINKED ISSUE #64: Make header for Pattern Library mobile ready
Only the o-header is mobile ready. Also the main header of the styleguide should work on mobile

AndyOGo:
fixes #64

### DIFF
--- a/src/components/m-dropdown/mixins/dropdown.scss
+++ b/src/components/m-dropdown/mixins/dropdown.scss
@@ -156,7 +156,7 @@
 }
 
 @mixin dropdown-icon($size: $dropdown__icon-size, $right: $dropdown__frame-x-padding, $color: $dropdown__icon-color) {
-  @include hardware-accelerate();
+  // @include hardware-accelerate(); Chrome has an issue
 
   position: absolute;
   top: 50%;
@@ -176,7 +176,7 @@
 }
 
 @mixin dropdown-content($zindex: $zindex-dropdown) {
-  @include hardware-accelerate();
+  // @include hardware-accelerate(); chrome has an issue
 
   position: absolute;
   top: 100%;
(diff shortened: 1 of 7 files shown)


AndyOGo:
@LucaMele 
I fixed the `context` bug, it was a hard one...
And I started adding some nice functional utils for debugging...

AndyOGo:
@LucaMele 
Wait we should double-check if that is really fixed (I would really want written tests here...)
Or may this is related to #101 
mobile-navi-deja-vu

src/components/m-dropdown/mixins/dropdown.scss
```diff
@@ -156,7 +156,7 @@
 }
 
 @mixin dropdown-icon($size: $dropdown__icon-size, $right: $dropdown__frame-x-padding, $color: $dropdown__icon-color) {
-  @include hardware-accelerate();
+  // @include hardware-accelerate(); Chrome has an issue
```
AndyOGo: I really want them for mobile, though I'm fine commenting them out right now

LucaMele: agree. But curretnly works for mobile as well so its okish. :(

LucaMele: so can i merge?

AndyOGo: sure merge it, but I'm really eager and considered to find a cross-browser solution here...

LucaMele: :) ok
just keep in mind that css 3d transform is also still an experimental feature and could bring other problems in future. I have used it as well lots of time and like the idea of using the gpu, but please consider this as not high prio.

May be worst reading this article here: https://www.smashingmagazine.com/2016/12/gpu-animation-doing-it-right/

LucaMele: same as WC at the end ;)

Issue #183, opened 2018-01-22
Interactive examples/demos


AndyOGo:
Every component should have a full-fledged interactive demo, including but no limited to:

- [ ] checkboxes enabing/disabling boolean attributes
- [ ] providing JSON data if needed
- [ ] callback allerts?
- [ ] Enums/radios/dropdowns for value ranges

LucaMele:
closing as new plib version is live. If issue is with v2 still relevant, please re-open

Issue #147, opened 2018-01-10
What do we need to be "stable" and not "beta" anymore


LucaMele:
Lets define what is mandatory for our first stable version

AndyOGo:
- [ ] first class props #39 
   - data down
   - actions up
- [ ] incremental rendering during lifecycle (`attributesChangedCallback`, `connectedCallback`, `disconnectedCallback`, context change)
- [ ] feature complete components
- [ ] complete, automatic documentation
- [ ] React POC
- [ ] Angular POC

LucaMele:
I agree on almost all  of andy apart of incremental build (which is important but not as much to be mandatory for a stable Version).

I will add these as well:

- [ ] Dist components have everything they need apart of polyfills (font in base64 for example)
- [ ] Cross OS is tested and works everywhere
- [ ] automating device testing and local testing concept at minimum (best would be running example)

LucaMele:
closing as new plib version is live. If issue is with v2 still relevant, please re-open

PR #195, opened 2018-01-25
fixed demo for m-button with react

zuzana:
### Why?

Fixed demo for toggling state of button. 
I don't know exactly why wrapping text in `<span>` or any other component works. 

### How to test?

1. Prepare
        
        rm -rf node_modules
        npm i 
        npm run serve

1. Navigate to [http://localhost:3000/#react](http://localhost:3000/#react)
1. Try toggling button

### Issue with babel plugin external-helpers

 I had issue with `external-helpers` plugin. I solved it according [this comment](https://github.com/storybooks/storybook/issues/1320#issuecomment-310777396)

      Web Component AXA-BUTTON has an error while loading its template:
      ReferenceError: babelHelpers is not defined
      
      Stack Trace: ReferenceError: babelHelpers is not defined
          at template (http://localhost:3000/components/m-button/index.js:857:74)
          at HTMLElement.render (http://localhost:3000/components/m-button/index.js:1624:23)
          at HTMLElement.connectedCallback (http://localhost:3000/components/m-button/index.js:1553:12)
          at http://localhost:3000/components/m-button/index.js:1850:25
          at HTMLDocument.once (http://localhost:3000/components/m-button/index.js:1827:12)


AndyOGo:
@zuzana 
Thanks for your contribution.
Sorry we haven't mentioned it anywhere, though may I ask you to not include distribution/generated code into the PR (`dist/` and `docs/` folder).
Those stuff is only necessary for packaging and gets automatically added by every release by running `npm run release`

src/demos/demo.react.jsx
```diff
@@ -24,7 +24,7 @@ class MyEventDemoReact extends React.Component {
   render() {
     return (
       <axa-button onClick={this.handleClick}>
-        {this.state.isToggleOn ? 'ON' : 'OFF'}
+        <span>{this.state.isToggleOn ? 'ON' : 'OFF'}</span>
```
AndyOGo: @zuzana 
The real issue is that we don't support incremental rendering of any web component right now.
Adding `<span>` isn't the proper way to deal with it. All necessary features are specified here #147. And for easy testing support of interactive demos #183

AndyOGo: Correct place to fix all of those is:
https://github.com/axa-ch/patterns-library/blob/master/src/components/_abstract/component-types.js

LucaMele: Agree for when it goes production ready. As we are still beta, this quick fix will enhance the demo and therefore for me is ok to merge. The dist folder however should be reset and we will make a deploy as soon as this is merged in order to update the dist and docs folder.

zuzana: changes to dist folder reverted.

LucaMele:
@zuzana, you can merge

PR #202, opened 2018-01-25
Feature/refactore typo mixins

### LINKED ISSUE #201: refactor typography mixins
Turned out that at CH we don't exactly follow the `font-weight` restriction provided at:
https://design.axa.com/web-guidelines/typography

Hence I think it makes more sense to:

- [x] allow for custom `font-weight` declaration of `normal`, `semibold` and `bold`
- [x] reduce the sass map to the only needed data
- [x] breakpoints should only output changing values

AndyOGo:
fixes #201


src/styles/functions/_typo.scss
```diff
@@ -19,11 +16,42 @@
     $key: #{$key}--#{$face};
   }
 
-  @if not map-has-key($typo-map, $key) {
-    @error "Key #{$key} not found in $typo-map.";
+  @return $key;
+}
+
+@function get-typo($name, $breakpoint: sm, $face: null) {
```
LucaMele: please add description of what u just told me :)

Issue #158, opened 2018-01-11
Components in dist folder should have all they need


LucaMele:
For example at the moment the font is loaded. It should be in base64 incorporated ONLY with disted components

AndyOGo:
This will bloat our CSS but I agree, with dist it should be all together.

PR #204, opened 2018-01-26
make global css work

LucaMele:


postcss.config.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+  plugins: [
+    require('autoprefixer')({ // eslint-disable-line
+
+    }),
+    require('./stack/tasks/postcss-font-base64.js')({ // eslint-disable-line
```
AndyOGo: I would provide it as a separate module in the future. For now is okay.

src/index.html
```diff
@@ -23,6 +23,7 @@
     <script src="app/webcomponents-loader.js"></script>
     <script src="app/app.js"></script>
     <link rel="stylesheet" href="./index.css">
+    <link rel="stylesheet" href="./sg.css">
```
AndyOGo: awesome

src/sg.scss
```diff
@@ -0,0 +1 @@
+@import "app/preview";
```
AndyOGo: awesome again

postcss.config.js
LucaMele: i though that as well. Totally agree

src/index.html
LucaMele: :)

LucaMele:
fixes #158

Issue #213, opened 2018-01-30
iOS does not trigger click events


AndyOGo:
iOS/Mobile Safari refuses to trigger click events bound by delegation.

AndyOGo:
dito

PR #214, opened 2018-01-30
Bugfix/ios footer titles not clickable

AndyOGo:
improves #213

### DIFF
--- a/src/components/m-dropdown/mixins/dropdown.scss
+++ b/src/components/m-dropdown/mixins/dropdown.scss
@@ -1,4 +1,5 @@
 @import "../settings/dropdown";
+@import "../../../styles/mixins/force-ios-click";
 @import "../../../styles/mixins/hardware-accelerate";
 @import "../../../styles/mixins/target";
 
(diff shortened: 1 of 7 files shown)


src/components/m-dropdown/mixins/dropdown.scss
```diff
@@ -148,6 +148,9 @@
 @mixin dropdown-toggle() {
   position: relative;
 
+  // this is important to enable click events on iOS
+  cursor: pointer;
```
LucaMele: mmmmm maybe better separate this inanother mixing (single responsability)

AndyOGo: well, every toggle button needs it anyway. Cause even at desktop the expected style is `pointer` on hover.

AndyOGo: or do mean to extract those 2 lines into a mixin and include it here?

AndyOGo:
@LucaMele 
Sorry we need this asap, i will merge now

Issue #54, opened 2017-12-12
Module build failed: Error: Unknown plugin "external-helpers"


AndyOGo:
I'm currently migrating the footer in axa.ch repo and face this bug.
So I just installed it locally into the aem repo, though we should document this some how.

> ERROR in ./~/@axa-ch/patterns-library/src/components/m-footer-links/js/footer-links.js
[scripts:watch] Module build failed: Error: Unknown plugin "external-helpers" specified in "/Users/axawinterthur/dev/aem-all_ext/frontend-lib/node_modules/@axa-ch/patterns-library/.babelrc" at 0, attempted to resolve relative to "/Users/axawinterthur/dev/aem-all_ext/frontend-lib/node_modules/@axa-ch/patterns-library"
[scripts:watch] ReferenceError: Unknown plugin "external-helpers" specified in "/Users/axawinterthur/dev/aem-all_ext/frontend-lib/node_modules/@axa-ch/patterns-library/.babelrc" at 0, attempted to resolve relative to "/Users/axawinterthur/dev/aem-all_ext/frontend-lib/node_modules/@axa-ch/patterns-library"
[scripts:watch]     at /Users/axawinterthur/dev/aem-all_ext/frontend-lib/node_modules/babel-core/lib/transformation/file/options/option-manager.js:180:17
[scripts:watch]     at Array.map (native)
[scripts:watch]     at Function.normalisePlugins (/Users/axawinterthur/dev/aem-all_ext/frontend-lib/node_modules/babel-core/lib/transformation/file/options/option-manager.js:158:20)
[scripts:watch]     at OptionManager.mergeOptions (/Users/axawinterthur/dev/aem-all_ext/frontend-lib/node_modules/babel-core/lib/transformation/file/options/option-manager.js:234:36)
[scripts:watch]     at OptionManager.init (/Users/axawinterthur/dev/aem-all_ext/frontend-lib/node_modules/babel-core/lib/transformation/file/options/option-manager.js:368:12)
[scripts:watch]     at File.initOptions (/Users/axawinterthur/dev/aem-all_ext/frontend-lib/node_modules/babel-core/lib/transformation/file/index.js:216:65)
[scripts:watch]     at new File (/Users/axawinterthur/dev/aem-all_ext/frontend-lib/node_modules/babel-core/lib/transformation/file/index.js:139:24)
[...]

Issue #219, opened 2018-02-01
use umd format


zuzana:
I would like to be able to extend class from patterns-library. 
Rollup configuration is using `iife` format. 
Was there a particular reason to switch back to `iife`?
If not, let's use `umd` that should not affect current functionality and enable us to extend exported classes. 
PR #220

LucaMele:
The main scope of the dist bundle is that it just works if the js is included. If we merge this, components wont work anymore if just the javascript is included. If we activate umd, we must as well have a js that loads requireJs or similar.

I would not reccomend to update to umd.

We should rather have es modules exported as well so that in the dist folder we have index.js for just iife and index.es.js for example for es module export.

What do u think about it?

The code snipped on the readme that shows an usage example should still work after this merge.

https://github.com/rollup/rollup/wiki/ES6-modules

zuzana:
I'm fine with whatever enables me to extend class. 
I need to re-render items on AXADropdown component after they have been changed so that's why I tried to solve it with extend.

LucaMele:
Ok so maybe for now we just go with extending from src folder and when the es modules are done we can extend from there

LucaMele:
And please let me know who the re-rendering did work for u :)

LucaMele:
ref: 44

Issue #44, opened 2017-12-06
Implement common module. Activate ES Modules in pkg.module prop and IFEE modules that global namespace


AndyOGo:
ATM we bundle up isolated bundle which include identical modules redundantly.

This breaks access to local variables between same module instances.

```js
// this variable is scope within each redundant import/bundle of this module
// and not accessible to other instances of this same module
const localVar = {}

export function foo() {
  localVar.key = 'value';
}

export function bar() {
  return localVar.key;
}
```

AndyOGo:
I think DOM data warehouse could be a feasible workaround, though perfect would be to clean up the build and extract common chunks into a separate vendor file.

LucaMele:
I think a module should contain and not contaminate variables so therefore the behaviour is correct IMHO. What is missing is a mechanism that can be used everywhere like u said global js or similar. We will discuss it. I dont see it as a bug, just a missing feature.

AndyOGo:
Okay.

Let's look at an example:

**shared module `push.js`:**
```js
const localVar = [];

export default function push(value) {
  localVar.push(value)

  return localVar;
}
```

**consumer foo:**
```js
import push from `push`;

const allVars = push('foo');

console.log(allVars); // => ['foo']
```

**consumer bar:**
```js
import push from `push`;

const allVars = push('bar');

console.log(allVars); // => ['bar']
// expected
// => ['foo', 'bar']
```

## Our current bundling result in this

The bug is that shared modules are bundled more than once.

**bundled consumer foo:**
```js
const localVar = [];

function push(value) {
  localVar.push(value)

  return localVar;
}

const allVars = push('foo');

console.log(allVars); // => ['foo']
```

**bundled consumer bar:**
```js
const localVar = [];

function push(value) {
  localVar.push(value)

  return localVar;
}

const allVars = push('bar');

console.log(allVars); // => ['bar']
// expected
// => ['foo', 'bar']
```

LucaMele:
as discussed:

dist should have:
common.js -> IFEE that uses global namespace for common. Ready for <script src=""> imports
common.module.js -> ready for webpack/rollup/etc imports

index.js -> IFEE that uses global namespace for common. Ready for <script src=""> imports
index.module.js -> ready for webpack/rollup/etc imports

AndyOGo:
@LucaMele 
To clarify I talk about the de facto standard **module** prop within `package.json` file.
[...]

PR #226, opened 2018-02-02
Feature/es module ready and cleanup

LucaMele:
please test locally after running: 

`npm cache clean --force && rm package-lock.json && rm -R node_modules/ && npm install && npm run serve`


LucaMele:
fiexs: #219 #44 

Instead of es modules, we go UMD as they are more compatible and there is an issue with preset-env. It will work with es-2015 preset, but is deprecated for babel. I would say we go like this and if we really need es modules instead of UMD, we can add them later.

The umd files will be in the dist folder after the release

LucaMele:
Added as well a cleaner way to include the SVG

and fixes #54

PR #236, opened 2018-02-12
Feature/form components

TheDadi:
introduce checkbox, radio, input, label, form-group w.i.p


src/components/a-checkbox/_template.js
```diff
@@ -0,0 +1,30 @@
...
+  return bel`<label class="${checkBoxClasses}">
+    <input
+      id="${id}"
+      class="a-checkbox__input" 
+      type="checkbox"
+      name="${name}" 
+      checked="${checked}" 
+      value="${value}" 
+      disabled="${disabled}">
+    <i class="a-checkbox__icon"></i>${childrenFragment}
```
AndyOGo: Shouldn't this use `<axa-icon></axa-icon>`?

src/components/a-form-input/_template.js
```diff
@@ -0,0 +1,52 @@
+import bel from 'bel';
+import classnames from 'classnames';
+
+export default function ({
```
AndyOGo: Hmh, I think it would be cool to support HTML5 input attributes as well.
May the easiest would be through rest `...args` or in this case `...props`.
I'm curious if **bel** understand object spread for attributes 🤔

src/components/a-form-input/index.scss
```diff
@@ -0,0 +1,104 @@
+@import "../../styles/settings/colors";
+@import "../../styles/settings/base";
+@import "../../styles/mixins/typo";
+@import "../../styles/mixins/control-transition";
+
+.a-form-input-wrapper {
```
AndyOGo: @LucaMele 
@TheDadi 
I'm curious how we want to deal with class naming for the root element of a web component.
Having two blocks within the same sass file seems kinda wrong to to me.
On the other side some class like `.a-form-input__root` indicates an element by BEM syntax.
Or we could even use the overline character to indicate a parent (not a child like by `__`): `.a-form-input‾‾root` (okay this looks kinda ugly).

So I guess, we should just conclude on a naming convention - a separte block is also fine with me.

src/components/a-form-input/index.scss
```diff
@@ -0,0 +1,104 @@
...
+.a-form-input {
+  @include control-transition();
+  @include typo-tag();
+
+  position: relative;
+
+  display: inline-flex;
+
+  width: 100%;
+  height: 40px;
```
AndyOGo: please use vars for value.
In this case I would do:

````scss
$form-input-height: 40px !default;
````

File: src/components/a-form-input/index.scss [hunk 5]
```diff
@@ -0,0 +1,104 @@
...
+
+  &:hover,
+  &:active,
+  &:focus {
+    &:not([disabled]):not(.a-form-input--error) {
+      border-color: $control-border-color--active;
+    }
+  }
+
+  &[disabled] {
```
AndyOGo: please add the `:disabled` pseudo state too

src/components/a-form-input/index.scss
```diff
@@ -0,0 +1,104 @@
...
+    background-color: $color-bg--gray;
+  }
+
+  &--error {
```
AndyOGo: Please write class names in full length, only use & for adjoining classes and stuff like that

src/components/a-form-input/index.scss
```diff
@@ -0,0 +1,104 @@
...
+    border-color: $control-border-color--error;
+  }
+
+  &--valid {
```
AndyOGo: dito

src/components/a-form-input/index.scss
```diff
@@ -0,0 +1,104 @@
...
+    max-width: calc(100% - 48px);
+  }
+
+  &--inline {
```
AndyOGo: dito

src/components/a-form-label/index.scss
```diff
@@ -0,0 +1,18 @@
...
+@import "../../styles/mixins/typo";
+
+.a-form-label {
+  @include typo-secondary-text--semibold();
+
+  display: block;
+  padding-bottom: 10px;
+
+  &--disabled {
+    color: $color-prim-gray-dusty;
```
AndyOGo: Please write the full class

src/components/a-form-label/index.scss
```diff
@@ -0,0 +1,18 @@
+@import "../../styles/settings/colors";
+@import "../../styles/settings/base";
+@import "../../styles/mixins/typo";
+
+.a-form-label {
+  @include typo-secondary-text--semibold();
+
+  display: block;
+  padding-bottom: 10px;
```
AndyOGo: We stick to `margin-top` for vertical rhythm.

src/components/m-form-group/index.scss
```diff
@@ -0,0 +1,46 @@
+@import "../../styles/settings/colors";
+@import "../../styles/settings/base";
+@import "../../styles/mixins/respond";
+
+// fix for web-components
+/* stylelint-disable */
+axa-form-group {
```
AndyOGo: We would like to avoid tag selectors, is it really necessary?

src/components/m-form-group/index.scss
```diff
@@ -0,0 +1,46 @@
...
+.m-form-group {
+  &:not(:first-child):not(.m-form-group--inline) {
+    margin-top: 20px;
+
+    @include respond-up(md) {
+      margin-top: 25px;
+    }
+  }
+
+  &--inline {
```
AndyOGo: Please expand to full class

src/styles/mixins/_control-base.scss
```diff
@@ -0,0 +1,83 @@
+@import "../../styles/settings/colors";
+@import "../../styles/settings/base";
+@import "../../styles/mixins/control-icon";
+@import "../../styles/mixins/visual-hidden";
+@import "../../styles/mixins/control-transition";
+
+@mixin control-base($block, $border-radius: 0) {
+  .#{$block} {
```
AndyOGo: I would remove the `$block` and use the intrinsic `&` provided by sass.
Like I did it here:
https://github.com/axa-ch/patterns-library/blob/master/src/styles/mixins/_visibility.scss
And utilized here:
https://github.com/axa-ch/patterns-library/blob/master/src/styles/trumps/_visibility.scss

src/styles/settings/_base.scss
```diff
@@ -30,3 +30,12 @@ $link-color: $color-link !default;
 $link-decoration: none !default;
 $link-hover-color: $color-link--hover !default;
 $link-hover-decoration: underline !default;
+
+$control-border-width: 1px !default;
```
AndyOGo: May we move this into a separate `_control.scss` settings file.

AndyOGo:
In general I'm very curious about generic form/control state management. Stuff like https://github.com/final-form/final-form

- [ ] touched/pristine
- [ ] disabled/enabled
- [ ] focused/blured/active
- [ ] correct/valid/error/invalid
- [ ] complete/incomplete

Regexs for pattern matching, etc.

src/components/a-form-input/index.scss
```diff
@@ -44,40 +47,44 @@
...
+
   &::placeholder {
     color: $color-prim-gray-dusty;
   }
 
   &:hover,
   &:active,
   &:focus {
-    &:not([disabled]):not(.a-form-input--error) {
+    &:not([disabled]):not(:disabled):not(.a-form-input--error) {
```
AndyOGo: Hmh, does this work if `:disabled` pseudo isn't supported

src/styles/mixins/_control-base.scss
```diff
@@ -4,8 +4,10 @@
 @import "../../styles/mixins/visual-hidden";
 @import "../../styles/mixins/control-transition";
 
-@mixin control-base($block, $border-radius: 0) {
-  .#{$block} {
+@mixin control-base($border-radius: 0) {
```
AndyOGo: Awesome

src/components/a-form-input/_preview.html
```diff
@@ -0,0 +1,7 @@
+<axa-form-input id="test" value="Pre-filled content" name="test-name"></axa-form-input>
+<axa-form-input id="test" value="Pre-filled content" name="test-name" inline></axa-form-input>
+<axa-form-input id="test" value="Pre-filled content" name="test-name" disabled></axa-form-input>
```
LucaMele: also for webcomponents id cannot be duplicated on a page. I would suggest, to not allow id to be set from outside. If we need one later, we can generate on the go a uuid. in `component-types` there is a static function for uuid generation that can be used (`BaseComponent.uuidv4()`).

here especially in the input-form we wont need an id as you packed wisely the input field inside its label

src/components/a-form-input/index.scss
```diff
@@ -0,0 +1,111 @@
+@import "../../styles/settings/colors";
+@import "../../styles/settings/controls";
+@import "../../styles/mixins/typo";
+@import "../../styles/mixins/control-transition";
+
+.a-form-input-wrapper {
```
LucaMele: this does not fully satisfy our BEM naming convention. Sorry for have not done a proper documentation yet. Maybe call me and I will inform you. But we generally use BEM where the element is separated with `__` from the block name and modifier with `--`

LucaMele: here would be i guess a-form-input the block and wrapper the element.

src/components/a-checkbox/_preview.html
```diff
@@ -1,2 +1,5 @@
-<axa-checkbox id="test" value="a" name="test-name" ></axa-checkbox>
```
LucaMele: This has visible layout issues that need to be resolved before merge:

bildschirmfoto 2018-02-16 um 09 18 27

src/components/a-radio/_preview.html
```diff
@@ -1,2 +1,5 @@
-<axa-radio id="test" value="a" name="test-name"></axa-radio>
```
LucaMele: This visible layout issues that need to be fixed before merging:

bildschirmfoto 2018-02-16 um 09 18 09

src/components/a-radio/_template.js
```diff
@@ -0,0 +1,30 @@
...
+  return bel`<label class="${radioClasses}">
+    <input
+      id="${id}"
+      class="a-radio__input" 
+      type="radio"
+      name="${name}" 
+      checked="${checked}" 
+      value="${value}" 
+      disabled="${disabled}">
+    <i class="a-radio__icon"></i>${childrenFragment}
```
LucaMele: We have axa-icon for this. Plus the i tag is a no-go for icons: https://html.spec.whatwg.org/#the-i-element

LucaMele: and as it contains an Atom `axa-icon` it will be a molecule and not an atom

src/components/a-form-label/_preview.html
```diff
@@ -0,0 +1,3 @@
+<axa-form-label htmlfor="id">Test Label</axa-form-label>
```
LucaMele: i like the idea of splitting label and input. But my option if we do it then we should also provide to bind the label with name and not just id. Generally is better have name as binding in order to avoid incorrect HTML in case of id duplication

LucaMele:
Thank you very much @TheDadi  for your effort and contribution

TheDadi:
@LucaMele you must have some kind of wrong sources cause the controls are styled nicely in every browser. :)

TheDadi:
image

TheDadi:
image

TheDadi:
image

LucaMele:
Weird. I did `rm -R node_modules && npm i && npm run serve` and still cant see them correctly, in any browser

src/components/a-choice/index.js
```diff
@@ -0,0 +1,55 @@
...
+      'a-choice--disabled': getAttribute(this, 'disabled'),
+    });
+
+    this.className = choiceClasses;
+    // Your DOM interaction here, but keep it decoupled.
+    // If you don't have any, just remove this function
+  }
+
+  disconnectedCallback() {
+    super.disconnectedCallback();
```
LucaMele: u can delete the whole method

src/components/a-choice/index.js
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * REF: https://www.w3.org/TR/custom-elements/#custom-element-conformance
+   */
+  connectedCallback() {
+    super.connectedCallback();
+
+
+    const choiceClasses = classnames(this.initialClassName, 'a-choice', {
+      'a-choice--error': getAttribute(this, 'error'),
```
AndyOGo: @TheDadi 
If you need multiple attributes you can replace this by a single call to `getAttributes`.
**Check-out** may missing camelcase fix #277

AndyOGo:
@TheDadi 
A big list of updates are arriving:
- rerendering #340 
- first class props #356 
- React Integration POC #355 

Unfortunately those breaking changes will need a bigger refactoring

Issue #237, opened 2018-02-13
Font spacing is bigger on iOS than in chrome


AndyOGo:
May it's caused by not text adjusting on mobile devices?

PR #238, opened 2018-02-13
try to improve text inflation

AndyOGo:
improves #237 according to https://developer.mozilla.org/en-US/docs/Web/CSS/text-size-adjust#Browser_compatibility

### DIFF
--- a/src/styles/_reboot.scss
+++ b/src/styles/_reboot.scss
@@ -38,8 +38,14 @@
 html {
   font-family: sans-serif; // 2
   line-height: 1.15; // 3
+  -webkit-text-size-adjust: none; // 4
   -webkit-text-size-adjust: 100%; // 4
+  -moz-text-size-adjust: none; // 4
+  -moz-text-size-adjust: 100%; // 4
+  -ms-text-size-adjust: none; // 4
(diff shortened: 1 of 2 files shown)


src/styles/_reboot.scss
```diff
@@ -38,8 +38,14 @@
 html {
   font-family: sans-serif; // 2
   line-height: 1.15; // 3
+  -webkit-text-size-adjust: none; // 4
```
LucaMele: vendor prefix shoukld be done by the post css which we have. please try without and ispoect the css

LucaMele: vendor prefix shoukld be done by the post css which we have. please try without and ispoect the css

AndyOGo: right, I will take a deep look at postCSS support for `text-size-adjust`

AndyOGo: Okay, `text-size-adjust` support seems to be there https://github.com/postcss/autoprefixer/issues/264

Expect Firefox for Android, which isn't include in the default browser list
https://github.com/postcss/autoprefixer/issues/496

PR #262, opened 2018-02-23
Feature/add more docslets

AndyOGo:
adding more doclets to prepare for better documentation.

### DIFF
--- a/src/js/array-remove.js
+++ b/src/js/array-remove.js
@@ -2,8 +2,8 @@
  * Removes an item from an array.
  * **Note:** This does change the array itself.
  *
- * @param array - The array to update.
- * @param element - The item to be removed.
+ * @param {Array} array - The array to update.
+ * @param {Element} element - The item to be removed.
  */
 export default function remove(array, element) {
   const index = array.indexOf(element);
--- a/src/js/array-remove.js
+++ b/src/js/array-remove.js
@@ -1,3 +1,10 @@
+/**
+ * Removes an item from an array.
+ * **Note:** This does change the array itself.
+ *
+ * @param array - The array to update.
+ * @param element - The item to be removed.
+ */
 export default function remove(array, element) {
   const index = array.indexOf(element);
 
--- a/src/js/class-list.js
+++ b/src/js/class-list.js
@@ -1,5 +1,12 @@
 const reTrim = /^\s+|\s{2,}|\s+$/g;
 
+/**
+ * Generates a bullet-prove Regex to match a CSS class name.
+ *
+ * @param {String} className - The CSS class name to match.
+ * @param {String} modifier - Any valid Regex modifier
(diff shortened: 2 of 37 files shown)


src/js/array-remove.js
```diff
@@ -1,3 +1,10 @@
+/**
+ * Removes an item from an array.
+ * **Note:** This does change the array itself.
+ *
+ * @param array - The array to update.
```
LucaMele: should not be this: `@param {Array}`

src/js/class-list.js
```diff
@@ -1,5 +1,12 @@
 const reTrim = /^\s+|\s{2,}|\s+$/g;
 
+/**
+ * Generates a bullet-prove Regex to match a CSS class name.
```
LucaMele: 💪

src/js/get-attribute.js
```diff
@@ -1,5 +1,13 @@
 const reJson = /^\s*(?:true|false|null|undefined|-?[0-9]+\.?[0-9]*|[[{](?:.|[\r\n])*[\]}])\s*$/;
 
+/**
+ * Get an HTML attribute's value by it's name.
+ * **Note:** The attribute's value get parsed as JSON if it's valid JSON.
+ *
+ * @param {Element} node - The HTML element to get an attribute from.
+ * @param {String} name - The attribute's name.
+ * @returns {*} - Returns the value of the attribute.
```
LucaMele: maybe add the direct result of a JSON parse

src/js/get-attributes.js
```diff
@@ -1,5 +1,11 @@
 import getAttribute from './get-attribute';
 
+/**
+ * Get all attributes of a node.
+ *
+ * @param {Element} node - The HTML element to get all attributes of.
+ * @returns {Object|null} -Returns a hash of all attributes or `null`.
```
LucaMele: space after -

LucaMele: and theoretically null is an object as well.. but better leave both :)

src/js/ui-events.js
```diff
@@ -4,7 +4,26 @@ import outer from './outer';
...
+ * - **non-interactive**
+ *
+ * Certain actions trigger interactive mode and others leave it.
+ */
 class UiEvents {
+  /**
+   * Default options of UIEvents
+   *
+   * @type {Object}
+   * @property {string} containerClass - A CSS class selector, if the container is not the WebComponent's node itself.
```
LucaMele: maybe keep consistency with using String and string.

src/js/which-transition-event.js
```diff
@@ -1,3 +1,8 @@
+/**
+ * Find which transition event is supported.
+ *
+ * @returns {String|null} - Returns the supported transition event or `null`.
```
LucaMele: would be better if the functionn returns `''` instead of `null` to keep type consistency

PR #277, opened 2018-02-26
Feature/add camel case

### LINKED ISSUE #274: HTML attributes should be camelCased for JS access
Currently using `getAttributes()` for `<axa-foo bar-baz="hello">` returns a non-camel-cased hash, like:

```js
{
  "bar-baz": "hello"
}
```

It would be much better if property access, i.e. dot-notation would be supported, like `attributes.barBaz`.

AndyOGo:
fixes #274

### DIFF
--- a/src/js/camelize.js
+++ b/src/js/camelize.js
@@ -1,5 +1,5 @@
-const reWord = /(?:^\w|[A-Z]|\b\w|\s+|[-_]+)/g;
-const reSeparator = /[-_]+/g;
+const regexWord = /(?:^\w|[A-Z]|\b\w|\s+|[-_]+)/g;
+const regexSeparator = /[-_]+/g;
 
 /**
  * Camelcase any given string.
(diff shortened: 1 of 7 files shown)


AndyOGo:
@LucaMele 
@TheDadi 
We may need to refactor other components now 🤔

src/js/camelize.js
```diff
@@ -0,0 +1,23 @@
+const reWord = /(?:^\w|[A-Z]|\b\w|\s+|[-_]+)/g;
```
LucaMele: still think regexWord would be a better name and less confusing

AndyOGo:
@LucaMele 
This changes props by attributes as follows

Attribute|New Prop
---|---
`data-toggle`|`dataToggle`
`data-atomic-category`|`dataAtomicCategory`
`first-left`|`firstLeft`

**Candidates for migration**

- `indextitle`
- `indexurl`
- `offcanvas`

Issue #20, opened 2017-12-01
is mobile click delay really fixed through CSS?


AndyOGo:
According to [BS4 reboot](https://getbootstrap.com/docs/4.0/content/reboot/#click-delay-optimization-for-touch):
> Traditionally, browsers on touchscreen devices have a delay of approximately 300ms between the end of a “tap” – the moment when a finger/stylus is lifted from screen – and the click event being fired. This delay is necessary for these browsers to correctly handle “double-tap to zoom” gestures without prematurely triggering actions or links after the first “tap”, but it can make your site feel slightly sluggish and unresponsive.

>Most mobile browsers automatically optimize away this 300ms delay for sites that use the width=device-width property as part of their responsive meta tag (as well as for sites that disable zooming, for instance with user-scalable=no, though this practice is strongly discouraged for accessibility and usability reasons). The biggest exceptions here are IE11 on Windows Phone 8.1, and iOS Safari (and any other iOS WebView-based browser) prior to iOS 9.3.

>On touch-enabled laptop/desktop devices, IE11 and Microsoft Edge are currently the only browsers with “double-tap to zoom” functionality. As the responsive meta tag is ignored by all desktop browsers, using width=device-width will have no effect on the 300ms delay here.

>To address this problem in IE11 and Microsoft Edge on desktop, as well as IE11 on Windows Phone 8.1, Bootstrap explicitly uses the touch-action:manipulation CSS property on all interactive elements (such as buttons and links). This property essentially disables double-tap functionality on those elements, eliminating the 300ms delay.

>In the case of old iOS versions (prior to 9.3), the suggested approach is to use additional scripts such as FastClick to explicitly work around the delay.

>For further details, see the compatibility table for suppressing 300ms delay for touchscreen interactions.

LucaMele:
closing as new plib version is live. If issue is with v2 still relevant, please re-open

PR #290, opened 2018-02-28
Feature/add tap event

### LINKED ISSUE #287: Add cross interface tap event type
To improve mobile responsiveness to `click` events we should introduce a custom cross-interface `tap` event type.

- [x] Should not be ignored by `passive` events - except if specified
- [ ] should be able to `preventDefault` click actions
- [x] Should support plug and play of touch interfaces
- [x] needs to works with **Touch Events**, **Pointer Events** and classic **Mouse Events**
- [x] We should never need to use any `cursor: pointer` hacks to enable them
- [ ] Scroll needs to be prevented in case of a valid tap
- [x] mitigates click delay #20 
- [ ] **IMPORTANT** remove this iOS shame to make outer click events work:
   - delete [`outer-ios-SHAME.js`](https://github.com/axa-ch/patterns-library/blob/master/src/js/outer-ios-SHAME.js)
   - remove it's usage from [`outer.js`](https://github.com/axa-ch/patterns-library/blob/master/src/js/outer.js#L19)

AndyOGo:
fixes #287, #20

### DIFF
--- a/src/js/tap.js
+++ b/src/js/tap.js
@@ -11,6 +11,7 @@ const pointerup = 'touchend pointerup MSPointerUp mouseup';
 const regexPointer = /touch|pointer/;
 const TAP_THRESHOLD = 150;
 const TAP_PRECISION = 60 / 2;
+const getPointer = event => event.targetTouches ? event.targetTouches[0] : event;
 
 // eslint-disable-next-line consistent-return
 function addEventListenerProxy(eventType, ...args) {
@@ -24,7 +25,8 @@ function addEventListenerProxy(eventType, ...args) {
 
   const { ownerDocument: { documentElement } } = target;
   let wasPointer = false;
-  let pointerId;
+  let cachedPointerId;
+  const isSamePointer = event => !event.pointerId || typeof cachedPointerId === 'undefined' || event.pointerId === cachedPointerId;
   let downTimestamp;
   let cachedX;
   let cachedY;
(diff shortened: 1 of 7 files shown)


src/js/tap.js
```diff
@@ -0,0 +1,141 @@
...
+const regexPointer = /touch|pointer/;
+const TAP_THRESHOLD = 150;
+const TAP_PRECISION = 60 / 2;
+const getPointer = event => event.targetTouches ? event.targetTouches[0] : event;
+
+// eslint-disable-next-line consistent-return
+function addEventListenerProxy(eventType, ...args) {
+  const eventTarget = this;
+
+  if (!hasSupport) {
```
LucaMele: unnecessary lines if u add `eventType = 'click'` on line 23

src/js/tap.js
```diff
@@ -0,0 +1,141 @@
...
+    eventType = 'click'; // eslint-disable-line no-param-reassign
+  }
+
+  if (eventType !== 'tap') {
+    return _addEventListener.call(eventTarget, eventType, ...args);
+  }
+
+  console.log(`>>> proxy ${eventType}`);
```
LucaMele: ?

src/js/tap.js
```diff
@@ -0,0 +1,141 @@
...
+
+    if (isValidThreshold && isValidPrecision) {
+      fire(target, eventType, event);
+    }
+  }
+}
+
+// eslint-disable-next-line consistent-return
+function removeEventListenerProxy(eventType, ...args) {
+  if (!hasSupport) {
```
LucaMele: dtto

AndyOGo:
@LucaMele 
The custom `tap` event type works now, but due to our multiple loading of modules. It is bound X times.
This needs to be fixed by properly bundling/loading modules.

tap-bound-multiple-times

LucaMele:
bildschirmfoto 2018-06-06 um 09 38 38

AndyOGo:
@LucaMele 
It's apparently an infinite loop, though the real problem is that this module is define multiple times (once for each web component). Which causes it to be patched multiple times und thus it logs very often for a single call.
It's the same problem we have with `on`.
Will be fixed until we have proper builds including code-splitting.

LucaMele:
it makes browser to freeze thought.

AndyOGo:
yes I now and max call stack...

AndyOGo:
**Breaking Changes:** Please migrate before merging
https://github.com/axa-ch/patterns-library/releases/tag/v2.0.1-beta.217

PR #297, opened 2018-03-01
added opt-in for gpu of button

### LINKED ISSUE #293: GPU accelerated rendering
I really like the idea of enabling the GPU to increase render performance, by utilising CSS properties like `transform: translate3d(0, 0, 0)`. **BUT:** we faced various glitches triggered by enabling this feature.

- messed up `z-index` stacking context #263 
- render issues #101
- iOS momentum scrolling renders a black background on overscroll #280
- `position: fixed` and overscroll and GPU accelerated rendering is like a misery

So what we could do is:
- using 2D-Transforms at the beginning and only if we face performance issues, opt-in for hardware-accelerated rendering.
- Only enable GPU rendering temporarily, e.g. for an animation.
- It would also be possible to have always an opt-in modifier like `--gpu`

A good reason to avoid it is our `m-button` component which will break `z-index` by default 😱

AndyOGo:
fixes #293 

**Could introduce breaking changes**

### DIFF
--- a/src/components/m-button/_template.js
+++ b/src/components/m-button/_template.js
@@ -10,6 +10,7 @@ export default function ({
   ghost,
   classes,
   motion,
+  gpu,
   arrow,
 }, childrenFragment) {
   const buttonClasses = classnames('m-button', classes, {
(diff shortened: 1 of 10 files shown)


AndyOGo:
@LucaMele 
This may introduce breaking changes.
So we should make sure, that we provide a migration path.

PR #303, opened 2018-03-02
Feature/bundle with webpack

### LINKED ISSUE #302: Add Webpack for bundling
Webpack 4 is out 🎉

# Todo

- [x] One bundle for each component
- [x] Common chunks should be hoisted to `common.js`...
- [ ] Enable sourcemaps for ESnext and SCSS

AndyOGo:
fixes #302


AndyOGo:
@LucaMele 
I finished bundling of all components.
Now a few things are still left, which I found within `stack/tasks/bundles/`

- [ ] **Web-Components Polyfill Loader**
   Those files are copied from it's `node_modules/@webcomponents/webcomponentsjs` folder
- [ ] **ES6-Ployfill**
   `src/app/es6-polyfills.js` isn't directly imported anywhere - is it need just for the **app** or all components?
- [ ] **Demos** folder, has the React button example. I don't know the purpose of it, being in a separate folder. May I ask you to explain me the magic?
- [ ] **Prod-All** - Is there something special I must consider?
- [ ] **HTML File Script Tag Injection** can be replaced by [html-webpack-plugin](https://github.com/jantimon/html-webpack-plugin)

LucaMele:
ES6-Ployfill is needed for all components and is imported in the dist file. It contains the necessary polyfills *only* for ie11

LucaMele:
the react magic was to have an environment where we can showcase react integration and i wanted to keep it separatly from everywhere

LucaMele:
not really as that was the "bundeling" only relevant for prod (no watchers).

AndyOGo:
@LucaMele 
Thanks for your quick response.

So I will inject the `es6-polyfill` indirectly. Hope we could do the same for the webcomponents-loader 🤔 

Then I will also add `src/app/` and `src/demos` to webpack's entry configuration. May I move all polyfills to a separate `vendor` folder - just to keep them nicely separated and I wouldn't need to exclude any files in webpack's config.

AndyOGo:
@LucaMele 
UMD bundles are easy, but code-splitting + UMD bundles doesn't seem to be supported right now:
https://github.com/webpack/webpack/issues/2471

AndyOGo:
@LucaMele 
Rollup has just released code-splitting, but it's still experimental:(
https://medium.com/rollup/rollup-now-has-code-splitting-and-we-need-your-help-46defd901c82

Issue #314, opened 2018-03-08
Sub Navigation does not close after first click


AndyOGo:
![sub-menu-not-closing-after-first-click](https://user-images.githubusercontent.com/914443/37139887-e9f24356-22af-11e8-95c4-d454b1695819.gif)

LucaMele:
I will fix it right now

LucaMele:
thanks

AndyOGo:
Maybe it's related to the removal of `js-header-navigation-close` class

AndyOGo:
BTW Why is there so much space between the links?
Is it intented?

LucaMele:
yes.. and on dev it behaves also again totally different than here. This is not good

LucaMele:
no its how it came out :)

AndyOGo:
@LucaMele 
I just checked locally the configuration passed to `UIEvents` from the **Stroke** and **Header Navigation**.
In fact they are not equal. `useDefaultEvent` is `false` for the patterns library, but true for axa.ch

<img width="1440" alt="screen shot 2018-03-08 at 13 39 58" src="https://user-images.githubusercontent.com/914443/37151598-6a8989fe-22d6-11e8-8c0e-1c7011a7a65e.png">

**VS**
<img width="1440" alt="screen shot 2018-03-08 at 13 40 09" src="https://user-images.githubusercontent.com/914443/37151610-712d0c0e-22d6-11e8-90ef-480a5b31107f.png">

PR #315, opened 2018-03-08
try to fix menu behaivour

LucaMele:


LucaMele:
@AndyOGo please check & test. I will redeploy dev asap this is merged. I dont like that the axa header behaves slightly different then this and will try to find out why.

LucaMele:
fixes #314

AndyOGo:
The fundamental difference of axa.ch is that if a menu-item is clicked a whole pageload is triggered. So it doesn't matter if the sub-navigation is closed or not by JS.

But in case of an SPA and client side routing, it matters.

LucaMele:
yes true. Im referring that the same menu settings with classes behavies diffrent on axa.ch and here (so both not in SPA modus)

LucaMele:
if we check the dev menu, its differently broken

LucaMele:
im not speaking about submenu, im speking about top level menu. i understand that if u have anchor tags it does not do a page reload (which is this bug fix here)

AndyOGo:
@LucaMele 
But we have hash changes, which we don't have at axa.ch

LucaMele:
ok im not eplining myself. Haschanges comes from sub menu. The top menu is the issue. Here in the p-lib the top menu is fine and works. Same top menu in axa.ch triggers a page reload as soon as u press it. Not the submenu. the top menu. This hasnt to do anything with this fix. The stroke js prevent default it and thats why i did work so far on axa.ch. But in the patterns library that prevent default was not needed. This is inconrgouency thats scares me.

AndyOGo:
stroke-not-moving

LucaMele:
shit ok will fix it. Is the return

AndyOGo:
Alright.
The top navigation should only prevent the default action if it has a sub navigation.

Let's go back to the basics, how is navigation's design originally intended.

- Sub Navigation entries are shown on `click` - in fact clicking the same item again closes it. So it's intended to be a toggle.
- A stroke is displayed if:
   - The current menu item is active or one of it's children is active
   - An optional sub-navigation is actiaved

**Regarding Stroke's Animations**

The interactive part consists of three different animations - one for `entering` an interactive state, one for `moving` between items and one for `leaving` the interactive state

has an active item | enter | move | leave
---|---|---|---
Nope | Animate the `height` from `0` to `4px` | Animate the `left` to the next items position | Animate the `height` from `4px` to `0`
Yes | Hide the stroke at the active element | nothing | Show the stroke at the active element 

**How do events related to those interactive states**

- first `click` -> `enter`
   - another item `click` -> `move`
   - second `click` on same item -> `leave`
   - clicking `outside` of the navigation -> `leave`
   - hitting `ESC` key -> `leave`

To make it even more complex the transition between those states is also handled by `transitionend` events. And the correct Stroke's position in case of `resize`, `orientation` events is also handled properly.

LucaMele:
fixes: #320, #316, #314

Issue #335, opened 2018-03-14
Incremental rendering duplicates DOM nodes


AndyOGo:
## Are you reporting a bug or a feature?

 - [x] Bug

## Expected Behavior

Web-Components with children (slots) have proper incremental rendering.

- [ ] Content / Slots have changed by `appendChild`, `replaceChild`, `insertBefore`, `innerText`, `VDOM`, etc.
   ```html
   <axa-foo><div>Some Content</div></axa-foo>
   <!-- Next the same component instance, but with different content -->
   <axa-foo><div>Other Content</div></axa-foo>
   ```
- [ ] Attributes have changed by `setAttribute`, `remove Attribute`, setting properties, etc.
   ```html
   <axa-foo></axa-foo>
   <!-- Next the same component instance, but with different attribute -->
   <axa-foo some-attr="bar"></axa-foo>
   ```

## Actual Behavior

Each Component does only render once.

## Specifications

Every Browser

**Using `<slot />` Tag**
![incremental-dublication](https://user-images.githubusercontent.com/914443/37393851-99e014b6-2772-11e8-9fda-3b348d8b83d3.gif)

AndyOGo:
Setting `innerText` on a Web-Component's instance does not trigger Lifecycle events 😱 

![no-lifecycle-events](https://user-images.githubusercontent.com/914443/37397298-89e8b3e8-277b-11e8-91d2-f4390897dda2.gif)

LucaMele:
ok.. seems like lots of work :)

AndyOGo:
Yep, indeed. I think we have 3 options:

- Relying on `DOMSubtreeModified` mutation event (deprecated)
- Relying on `MutationObserver`
- Monkey Patching DOM API's with getters/setters (ES5 only)

All aren't nice IMHO.

AndyOGo:
https://developers.google.com/web/fundamentals/web-components/shadowdom#lightdom

AndyOGo:
seems that [`<slot />` is intended for ShadowDOM only](https://developers.google.com/web/fundamentals/web-components/shadowdom#slots)?
>A component can define zero or more slots in its shadow DOM.

AndyOGo:
setting properties or calling `setAttribute` form outside does not trigger `attributeChangedCallback` 😱

PR #337, opened 2018-03-14
added lifecycle test

AndyOGo:
Experiment for #335 

Not meant to be merged.
 
 # How Has This Been Tested?
 
In chrome's console:

```js
var test = document.getElementById('axa-test');
// this triggers attributeChangeCallback only if foo is defined in observeAttributes
test.setAttribute('foo', 'bar');
// property changes do not trigger attributeChangeCallback, even if foo is defined in observeAttributes
test.foo = 'baz';
// not a single lifecycle event triggers, DOM gets corrupted
test.innerText = 'foo bar';
// not a single lifecycle event triggers, DOM gets corrupted
test.innerHTML = '<span>foo bar</span>';
// not a single lifecycle event triggers, DOM gets corrupted
test.appendChild(document.createElement('div'))
```

### DIFF
--- a/src/components/m-test/_example.html
+++ b/src/components/m-test/_example.html
@@ -1,0 +1,1 @@
+<!--Please create here a HTML example by using just default HTML tags-->
--- a/src/components/m-test/_preview.html
+++ b/src/components/m-test/_preview.html
@@ -1,0 +1,1 @@
+<axa-test id="axa-test">Some Text</axa-test>
--- a/src/components/m-test/_template.js
+++ b/src/components/m-test/_template.js
@@ -1,0 +1,5 @@
+import bel from 'bel';
+
+export default ({ }, childrenFragment) => bel`
+  <article class="m-test__article">${childrenFragment}</article>
+`;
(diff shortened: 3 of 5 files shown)


LucaMele:
why is jenkins failing?

AndyOGo:
I don't know, may I have lint issue... will check

AndyOGo:
@LucaMele 
Monkey Patching works quite well, the only issue I see is that the whole DOM subtree is rebuild every time, which could lead to performance implications. So far I have monkey patch `innerText`, `innerHTML` and `appendChild`, we may need more... or `textContent` for cross browser compatibility.

AndyOGo:
@LucaMele 
I just added an interval to rerender every second automatically for testing purpose.
At the moment it will rerender whether something has changed or not.

AndyOGo:
@LucaMele 
Since this re-renders the whole DOM subtree we will face the following issues:

- References to DOM nodes inside the subtree get out-of-sync
- Performance

AndyOGo:
Since we just have merge real incremental rendering this can be closed :)

PR #338, opened 2018-03-14
add small dropdown

LucaMele:
Fixes # (issue).

 - add a small option for the dropdown

## Type of change

 # How Has This Been Tested?

 Locally

### DIFF
--- a/src/components/m-dropdown/_template.js
+++ b/src/components/m-dropdown/_template.js
@@ -3,17 +3,17 @@ import raw from 'bel/raw';
 
 const arrowIcon = '<axa-icon id="angle-bracket-down" classes="m-dropdown__icon"></axa-icon>';
 
-const nativeSelect = ({ items }) => bel`<div class="m-dropdown__select-wrap" tabindex="0">
-    <select class="m-dropdown__select">
+const nativeSelect = ({ items, size }) => bel`<div class="m-dropdown__select-wrap ${size === 'sm' ? 'm-dropdown__select-wrap--small' : ''}" tabindex="0">
+    <select class="m-dropdown__select ${size === 'sm' ? 'm-dropdown__select--small' : ''}">
       ${items && items.map(({ name, url }) => bel`
         <option data-url="${url}">${name}</option>
       `)}
     </select>
-    <div class="m-dropdown__select-icon">${raw(arrowIcon)}</div>
+    <div class="m-dropdown__select-icon ${size === 'sm' ? 'm-dropdown__select-icon--small' : ''}">${raw(arrowIcon)}</div>
   </div>`;
 
-const enhancedSelect = ({ title, items }) => [
-  bel`<button type="button" class="m-dropdown__toggle js-dropdown__toggle">
+const enhancedSelect = ({ title, items, size }) => [
+  bel`<button type="button" class="m-dropdown__toggle js-dropdown__toggle ${size === 'sm' ? 'm-dropdown__toggle--small' : ''}">
     ${title}${raw(arrowIcon)}
   </button>`,
   bel`<ul class="m-dropdown__content">
--- a/src/components/m-dropdown/_template.js
+++ b/src/components/m-dropdown/_template.js
@@ -3,17 +3,17 @@ import raw from 'bel/raw';
 
 const arrowIcon = '<axa-icon id="angle-bracket-down" classes="m-dropdown__icon"></axa-icon>';
 
-const nativeSelect = ({ items, size }) => bel`<div class="m-dropdown__select-wrap ${size === 'sm' ? 'm-dropdown__select-wrap--small' : ''}" tabindex="0">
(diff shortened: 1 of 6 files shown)


src/components/m-dropdown/index.scss
```diff
@@ -12,16 +12,30 @@
...
+  line-height: $dropdown__select-height-small;
+}
+
 // native select
 .m-dropdown__select-wrap {
   @include dropdown-frame();
   @include dropdown-select-wrap();
 }
 
+.m-dropdown__select-wrap--small {
```
AndyOGo: Won't we have just one modifier at the root node and increase CSS specificity for children?

src/components/m-dropdown/index.js
```diff
@@ -27,9 +28,11 @@ class AXADropdown extends BaseComponentGlobal {
     super.render();
 
     const inFlow = this.hasAttribute('in-flow');
+    const size = getAttribute('size');
 
     this.className = classnames(this.initialClassName, 'm-dropdown js-dropdown', {
       'm-dropdown--in-flow': inFlow,
+      'm-dropdown--small': !!size,
```
AndyOGo: Won't we use a more consistent syntax, because size isn't a boolean here, like in the `m-button`:
```js
{
  [`m-button--${size}`]: size,
}
```

src/components/m-dropdown/index.scss
LucaMele: no please not increase specifity! Is EVIL! ahahaha. thats why we have bem. So i do not agree on this

src/components/m-dropdown/index.js
LucaMele: so to have instead of `--small` `--sm`? yes makes sense i will change it

AndyOGo: yep

LucaMele: i like it.. much better agree

src/components/m-dropdown/index.scss
AndyOGo: Well, what should I say. In general I agree with you.
Though in this case we talk about controlled specificity increase for valid reasons, like:
- styles should and can't be easily overwritten here
- all changes needed for `small` modifier can be grouped together, which increase code visibility.

AndyOGo: BTW. I just realized that this modifier should implemented by a reusable mixin.

LucaMele: Agree for the reason that they are grouped and so easier maintainable, but still benefit is higher if we do not increase specifity especially if a consumer overrides the class. If they do it with high specificity, they have to use `!important`. So i think for the purpose of the Patterns Library is better not to do it and have a bit more classes instead. Clean Code vs Usability for Consumers

LucaMele: the small modifier? so to have an iclude for the height? and one for the line-height?

AndyOGo: Mhm, indeed boths points are valid. In that case we should define a convention and refactor all other components if necessary

src/components/m-dropdown/_template.js
```diff
@@ -3,17 +3,17 @@ import raw from 'bel/raw';
 
 const arrowIcon = '<axa-icon id="angle-bracket-down" classes="m-dropdown__icon"></axa-icon>';
 
-const nativeSelect = ({ items }) => bel`<div class="m-dropdown__select-wrap" tabindex="0">
-    <select class="m-dropdown__select">
+const nativeSelect = ({ items, size }) => bel`<div class="m-dropdown__select-wrap ${size === 'sm' ? 'm-dropdown__select-wrap--small' : ''}" tabindex="0">
```
AndyOGo: This could also be updated to ${size && `m-dropdown__select-wrap--${size}`}

LucaMele: i already did it in the changes im doing :) thanks anyway

src/components/m-dropdown/_template.js
```diff
@@ -3,17 +3,17 @@ import raw from 'bel/raw';
 
 const arrowIcon = '<axa-icon id="angle-bracket-down" classes="m-dropdown__icon"></axa-icon>';
 
-const nativeSelect = ({ items }) => bel`<div class="m-dropdown__select-wrap" tabindex="0">
-    <select class="m-dropdown__select">
+const nativeSelect = ({ items, size }) => bel`<div class="m-dropdown__select-wrap ${size ? `m-dropdown__select-wrap--${size}` : ''}" tabindex="0">
+    <select class="m-dropdown__select ${size ? `m-dropdown__select--${size}` : ''}">
```
AndyOGo: This could be even more reduced by using `&&` instead of trinary operator, like:
```js
`${size && `m-dropdown__select-wrap--${size}`}`
```

LucaMele:
:)

PR #348, opened 2018-03-19
restructure global scss folder

### LINKED ISSUE #209: Foldder structure in global scss
I think we made a mistake.

Our global scss is structed as follows:
- `/animations`
- `/functions`
- `/mixins`
- `/settings`
- `/trumps`

But we wanted to organise everything at component level. So maybe it would be better doing it like:
- `/scss/typo/`
   - `typo.functions.scss`
   - `typo.mixins.scss`
   - `typo.settings.scss`

AndyOGo:
Fixes #209 

 - make global SCSS more organised like components. The general rule is
   - are more than 1 file related and split across folders? Then move it to it's own folder
   - left it, since it is more like a utility or super generic
 - fixes `icon-size` mixin to just `size`
 - keeps old entry points and adds deprecation `@warn`ings

**Grouped so far:**

- grid
- typo
- vertical-rhythm

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) - in this case break is mitigated
 
 # How Has This Been Tested?
 
locally.
 
 # Checklist:
 
all good


LucaMele:
I dont have time unfortunatly to test it this Week

AndyOGo:
@LucaMele 
I have kept the old files and replaced their contents with imports from new locations.
Plus I added SCSS warnings that they stop importing from old locations.
So should be save for old imports:)

PR #351, opened 2018-03-19
refactored classlist functions to support multiple class names

### LINKED ISSUE #306: CSS class-list should handle more than one class
`add`, `has`, `toogle` and `remove` should be able to deal with mora than just a single class.

AndyOGo:
Fixes #306 .

 - added support for multiple classes
 - needs to refactor existing code

## Type of change

 # How Has This Been Tested?
 
locally
 
 # Checklist:
 
all done

### DIFF
--- a/src/components/m-header-navigation/js/stroke.js
+++ b/src/components/m-header-navigation/js/stroke.js
@@ -87,8 +87,7 @@ class Stroke extends UiEvents {
     if (has(node, this.options.activeClass)) {
       this._activeNode = node;
 
-      add(node, this.options.activeOpenClass);
-      add(node, this.options.activeMoveClass);
+      add(node, this.options.activeOpenClass, this.options.activeMoveClass);
     }
 
     requestAnimationFrame(() => {
@@ -135,8 +134,7 @@ class Stroke extends UiEvents {
     this._offMoving();
 
     if (this._activeNode) {
-      remove(this._activeNode, this.options.activeMoveClass);
-      remove(this._activeNode, this.options.activeOpenClass);
+      remove(this._activeNode, this.options.activeMoveClass, this.options.activeOpenClass);
       this._activeNode = null;
     }
 
@@ -149,8 +147,7 @@ class Stroke extends UiEvents {
       remove(this._list, this.options.interactiveClass);
 
       requestAnimationFrame(() => {
-        remove(this._stroke, this.options.moveClass);
-        remove(this._stroke, this.options.enterClass);
+        remove(this._stroke, this.options.moveClass, this.options.enterClass);
         remove(_node, this.options.activeOpenClass);
       });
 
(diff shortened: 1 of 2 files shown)


src/components/m-header-navigation/js/stroke.js
```diff
@@ -87,8 +87,7 @@ class Stroke extends UiEvents {
     if (has(node, this.options.activeClass)) {
       this._activeNode = node;
 
-      add(node, this.options.activeOpenClass);
-      add(node, this.options.activeMoveClass);
+      add(node, this.options.activeOpenClass, this.options.activeMoveClass);
```
AndyOGo: @LucaMele 
@TheDadi 
Is it really worth this update?
I checked all other components, only the stroke needs to `add`/`remove` multiple classes at once on the same DOM node.
Valid Reason for me are:
- Pattern-Library consumers won't need to worry if they need to attach multiple classes
- It is more coerce the `Element.classList` API https://developer.mozilla.org/en-US/docs/Web/API/Element/classList

AndyOGo: @LucaMele 
@TheDadi 
Knock knock

LucaMele: :)

src/js/class-list.js
```diff
@@ -12,56 +12,142 @@ function getReClass(className, modifier) {
...
-  if (!has(node, className)) {
-    node.className += ` ${className}`;
+export function add(node, ...classNames) {
+  const hasResult = has(node, ...classNames);
+
+  if (hasResult === true) {
+    return;
+  }
+
+  let classesToAdd = [...classNames];
```
LucaMele: u might want to cache deconstruction before u do has(node, ...classNames)

src/js/class-list.js
```diff
@@ -12,56 +12,142 @@ function getReClass(className, modifier) {
...
+  const { className: nodeClassName } = node;
+  let hasAtLeastOne = false;
+  let hasAllClasses = true;
 
-  return regexClass.test(node.className);
+  const hasArray = classNames.map(mapHas);
+
+  return hasAllClasses || hasAtLeastOne ? hasArray : false;
+
+  function mapHas(className) {
```
LucaMele: grrrr hoisting

src/js/class-list.js
```diff
@@ -12,56 +12,142 @@ function getReClass(className, modifier) {
...
-    const regexClass = getReClass(className, 'g');
+export function remove(node, ...classNames) {
+  const hasResult = has(node, ...classNames);
 
-    node.className = node.className.replace(regexClass, ' ').replace(regexTrim, ' ');
+  if (hasResult === false) {
+    return;
   }
+
+  let classesToRemove = [...classNames];
```
LucaMele: dtto

PR #356, opened 2018-03-21
Bugfix/fix props vs attrs

### LINKED ISSUE #39: first class props
How do we support properties other than `string`?

**Requirements:**

- [ ] Data down, actions up
- [ ] support prop types of `Number`, `String`, `Boolean`, `Object`, `Array` `Null`
- [ ] `Function` as property

**Note:**
We already parse the given property string with `JSON.parse()`. This way we can support `true`, `false`, `1234`, `'string'`, `[]`, `{}`, `null`

**Nice Reads:**
- http://robdodson.me/interoperable-custom-elements/
- React discussion: https://github.com/facebook/react/issues/7901

### LINKED ISSUE #321: HTML Attributes VS DOM properties
Currently we utilize the attribute DOM APIs which in fact read from markup like `<div data-foo="bar">`.

In contrast the property API read from the generated DOM like `node.property`

# Todo

- [ ] `<axa-foo attr="bar">` creates DOM node from HTML and sets an attribute
- [ ] DOM created through JS
   ```js
   const axaFoo = document.createElement('axa-foo');
   axaFoo.attr = 'bar'; // sets the property of the DOM node (not necessarily the HTML attribute)
   ```
- [ ] Is `attributesChangedCallback` Lifecycle method called?
   - for both - attribute and property changes?
   - for attribute only?
   - for property onyl?

# Resources

- https://developers.google.com/web/fundamentals/web-components/customelements#properties_and_attributes
- https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#reflecting-content-attributes-in-idl-attributes

AndyOGo:
Fixes #321 and foundations for #39 .

 - Adds DOM property getters and setters for related HTML attributes
 - Properly checks for Arrays in templates

The idea is that **first-class** props can be set by DOM properties. In case of HTML's `setAttribute` is used, we parse the attributes value once for JSON and store the result as **first-class** DOM property. This way it will be much easier to bridge with frameworks like React or Angular and `getAttribute/s` won't be needed any more, just simple prop access.
 
## Type of change

 # How Has This Been Tested?
 
Locally
 
 # Checklist:
 
 - [x] I have made corresponding changes to the documentation


src/components/a-icon/_preview.html
```diff
@@ -1 +1,34 @@
-<axa-icon icon="logo-AXA" classes="a-icon"></axa-icon>
+<axa-icon id="axa-icon-test" icon="logo-AXA" classes="a-icon"></axa-icon>
+
+<script>
```
LucaMele: ahahahaha like this!

AndyOGo:
**BROKEN** right now

```sh
%cWeb Component %cAXA-HEADER-MOBILE%c#2 has an error while loading its template:
RangeError: Maximum call stack size exceeded

Stack Trace: RangeError: Maximum call stack size exceeded
    at HTMLElement._makeContextReady (http://localhost:3000/components/m-header-mobile/index.js:1385:16)
    at HTMLElement.connectedCallback (http://localhost:3000/components/m-header-mobile/index.js:1134:14)
    at HTMLElement.connectedCallback (http://localhost:3000/components/m-header-mobile/index.js:2033:127)
    at HTMLElement.didRenderCallback (http://localhost:3000/components/m-header-mobile/index.js:2057:14)
    at HTMLElement.render (http://localhost:3000/components/m-header-mobile/index.js:1284:14)
    at HTMLElement.connectedCallback (http://localhost:3000/components/m-header-mobile/index.js:1131:12)
    at HTMLElement.connectedCallback (http://localhost:3000/components/m-header-mobile/index.js:2033:127)
    at HTMLElement.didRenderCallback (http://localhost:3000/components/m-header-mobile/index.js:2057:14)
    at HTMLElement.render (http://localhost:3000/components/m-header-mobile/index.js:1284:14)
    at HTMLElement.connectedCallback (http://localhost:3000/components/m-header-mobile/index.js:1131:12)
```

src/js/abstract/base-component.js
```diff
@@ -92,17 +154,19 @@ export default class BaseComponent extends HTMLElement {
...
-    if (this._isConnected && this._hasRendered) {
-      this.render();
+  attributeChangedCallback(name, newValue, oldValue) {
+    if (ENV !== PROD) {
+      lifecycleLogger(this.logLifecycle)(`+++ attributeChangedCallback -> ${this.nodeName}#${this._id} | ${name} from ${oldValue} to ${newValue}\n`);
     }
+
+    const key = camelize(name);
+
+    this[key] = toProp(newValue);
```
AndyOGo: @LucaMele 
I would be keen to know, how `removal` is detected. Guess just an empty string `''`.
May we should add something like prop-types but in separate PR, as soon as it becomes an issue. And I would realy like to have automatic tests for this...

src/js/abstract/base-component.js
```diff
@@ -182,7 +246,7 @@ export default class BaseComponent extends HTMLElement {
           });
         }
 
-        const items = template(getAttributes(this), this.childrenFragment);
+        const items = template(this._props, this.childrenFragment);
```
LucaMele: i have a mixed feeling trying to reactify standard web components. I like this solution, dont get me wrong, but im just a bit scared :) U can ignore this comment :)

src/js/abstract/base-component.js
LucaMele: this is a bit dangerous to be honest.. What if we override a DOM property like style or so on?

src/js/abstract/base-component.js
```diff
@@ -83,7 +121,31 @@ export default class BaseComponent extends HTMLElement {
...
+          lifecycleLogger(this.logLifecycle)(`\n!!! observedAttributes start -> ${this.nodeName}#${this._id}`);
+        }
+
+        observedAttributes.forEach((attr) => {
+          const key = camelize(attr);
+
+          if (this.hasAttribute(attr)) {
+            const value = getAttribute(this, attr);
+
+            this[key] = value;
```
LucaMele: this is a bit dangerous to be honest.. What if we override a DOM property like style or so on?

LucaMele: i understand we need it for react thought just want to be sure

src/js/abstract/base-component.js
AndyOGo: Absolutely, may we add a blacklist for `['style', `className`, `id`, etc]`which throws?
Or I could a check if the property exists already, then throw 😇

LucaMele: yes better with check + throw

src/js/abstract/base-component.js
AndyOGo: I agree. We need to document it and could check if a `prop` exists already, if so `throw`

LucaMele:
hope that super[key] works on all browsers...

LucaMele:
OK then it should work

PR #395, opened 2018-04-06
remove email

LucaMele:
Security remove email

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "node": ">=8.8.1",
     "npm": "~5.4.2"
   },
-  "author": "luca.mele@axa-winterthur.ch",
+  "author": "Luca Mele",
   "license": "MIT",
   "devDependencies": {
     "autoprefixer": "^7.2.3",


AndyOGo:
@LucaMele 
Great, though your email is still in git's history.
You can pick an previous commit like https://github.com/axa-ch/patterns-library/blob/v2.0.1-beta.27/package.json#L56

We would need to rewrite the whole history, which can be done by `git filter-branch` or there is an awesome tools which I used for my affected repos [BFG Repo-Cleaner](https://rtyley.github.io/bfg-repo-cleaner/). This updates all commit, all branches, all refs, all reflogs, you name it.
Though we need to cautious that all PR's are merged before we do this.

LucaMele:
@AndyOGo is fine for the moment. I dont have time for this and till i revieve an email that says also like this is not ok, i leave it. Git history is more important at the moment.

LucaMele:
thanks

AndyOGo:
Though you also received a warning. Yeah thats fine.
May their automatic tool is only checking the latest commit on `HEAD` anyway...

PR #419, opened 2018-05-01
some cleanup

LucaMele:
Fixes # (issue).

Cleansup the component structure


src/components/a-input/index.js
```diff
@@ -18,17 +18,17 @@ class AXAFormInput extends BaseComponentGlobal {
...
+      'a-input--inline': inline,
+      'a-input--error': error,
+      'a-input--disabled': disabled,
     });
   }
 }
 
 wcdomready(() => {
-  window.customElements.define('axa-form-input', AXAFormInput);
+  window.customElements.define('axa-input', AXAFormInput);
```
AndyOGo: The variable name is still `AXAFormInput` should be just `AXAInput`

src/components/a-input/index.scss
```diff
@@ -4,7 +4,7 @@
 @import "../../styles/control/control-transition";
 @import "./settings/form-input";
```
AndyOGo: this file also has the `form` prefix

src/components/o-accordion/index.scss
```diff
@@ -1,3 +1,3 @@
-.m-accordion {
+.o-accordion {
```
AndyOGo: And the file which uses this class?

LucaMele: does not exists

LucaMele: i remove it

AndyOGo: wait, but it should. Thought custom element are by default `inline` and in most cases we want them to be `display: block`.

LucaMele: ah yes true sorry i revert :)

src/components/a-input/index.js
```diff
@@ -18,17 +18,17 @@ class AXAFormInput extends BaseComponentGlobal {
...
   }
 }
 
 wcdomready(() => {
-  window.customElements.define('axa-form-input', AXAFormInput);
+  window.customElements.define('axa-input', AXAInput);
 });
 
-export default AXAFormInput;
+export default AXAInput;
```
AndyOGo: 👍

src/components/o-accordion/index.scss
LucaMele: no hold on the aricle underneath has display block as default so is it not enough that?

src/components/m-accordion/index.scss
```diff
@@ -1,3 +0,0 @@
-.m-accordion {
```
AndyOGo: please restore, this class is used here:
[`src/components/m-accordion/index.js#L19`](https://github.com/axa-ch/patterns-library/blob/develop/src/components/m-accordion/index.js#L19)

src/components/a-input/index.scss
```diff
@@ -98,7 +98,7 @@
   }
 }
 
-.a-form-input__valid-icon {
+.a-input__valid-icon {
   display: none;
 
   width: $form-input-icon-size;
```
AndyOGo: This variable name should also just start with `$input`

Issue #439, opened 2018-05-25
Make input-wrapping components controllable by React


pavestru:
## Are you reporting a bug or a feature?

 - [x] Bug (or is it a feature?)

## Expected Behavior

The recommended way to implement forms with React is to [use controlled components](https://reactjs.org/docs/uncontrolled-components.html).  That means that the state of React component controls the state of forms input elements, like checked, value or disabled. Changing React state from outside changes state of form input elements.

## Actual Behavior

Currently custom elements like axa-checkbox wrap form elements and create a facade for their attributes. In case of axa-checkbox, it wraps an `<input type="checkbox" />` with attribute `checked`. However, these attributes do not reflect well between the custom element and its own wrapped inputs. For example, in case of axa-checkbox, when set `checked={true}` from outside, it gets reflected to input's `checked="checked"` correctly once, but changing axa-checkbox's checked attribute further from outside does not change the `checked` attribute of the input element.

However, interacting with the element itself directly (e.g. by clicking) changes its internal state and is able to trigger event listeners.

Observed in components:
 - axa-choice
 - axa-checkbox
 - (possibly others)

## Steps to Reproduce the Problem

Notice when clicking first time on AxaCheckbox, its input's internal checked state is changed to unchecked first (independently of the app state) and from the 2nd click on is in sync with state:

![controlled_axa-checkbox](https://user-images.githubusercontent.com/10186479/40559944-be737d12-6058-11e8-8029-2f92b3ed4a00.gif)

```js
// App.js
import React, { Component } from "react";
import "@axa-ch/patterns-library/dist/components/a-checkbox/index.js";

class App extends Component {
  state = {
    checked: false
  };
  componentDidMount() {
    this.checkbox.addEventListener("change", this.onCheckBoxChange, {
      passive: false
    });
  }
  onCheckBoxChange = e => {
    const checked = e.target.checked;
    this.setState({ checked });
  };
  render() {
    return (
      <div>
        State: {JSON.stringify(this.state.checked)}
        <axa-checkbox
          checked={this.state.checked}
          name="test-name2"
          ref={el => (this.checkbox = el)}
        >
          AxaCheckbox
        </axa-checkbox>
        <label>
          <input
            type="checkbox"
            checked={this.state.checked}
            onChange={this.onCheckBoxChange}
          />
          Native checkbox
        </label>
      </div>
    );
  }
}

export default App;
```

## Specifications

 - Browser: Chrome 66 on macOS

[...]

PR #440, opened 2018-05-25
Work-around for axa-choice checked attribute controllable by React

pavestru:
Workaround to #439 for axa-choice - does not fix the issue

 - adds `controlled` attribute to axa-choice
 - adds `controlled` and `uncontrolled` modifiers to `a-choice__input`
 - in case of controlled is set to true, styles are applied based on axa-choice's checked attribute instead of its wrapped input element's checked attribute
 - even though the same was applied to `disabled` attribute, it is discouraged to use controlled disabled attribute, because it is still buggy (it is not possible to enable input once disabled)

## Type of change

- [x] Bug fix (non-breaking change which is just a workaround and not a real fix)
 
 # How Has This Been Tested?
 
Tested on Chrome 66, macOS
 
 # Checklist:
 

src/components/a-choice/_template.js
```diff
@@ -7,15 +7,16 @@ export default function ({
   name,
   checked = false,
   disabled = false,
+  controlled = false,
 }, childrenFragment) {
   return html`<label class="a-choice__wrapper">
     <input
       id=${inputId}
-      class="a-choice__input" 
+      class="a-choice__input a-choice__input--${controlled ? 'controlled' : 'uncontrolled'}"
```
AndyOGo: This is indeed a bug.

Though I would not fix the symptom by adding overruling classes.
Rather I assume it to be broken at the our DOM-diffing level.

pavestru:
Yes, @TheDadi , you're right in that that it's not a good solution. It's just my attempt at a workaround in order to move forward until a solution is found. Also thank you for the link to the React issue.

Actually, one of the first things I tried was
```diff
        name="${name}"
-      checked="${checked}"
+      ${checked ? 'checked="checked"' : ''}
       value="${value}"
```
... but it didn't work. Then I realized that this transformation from `checked="true/false"` to `checked="checked"` or no checked attribute displayed at all was already taken care of by nanohtml as it was working correctly at first render. The consecutive re-renders based on the subsequent changed state were not re-rendered correctly.

So, probably using withReact will be the solution as @AndyOGo points out in #439.

AndyOGo:
Released some bugfixes and published a working example at code-sandbox:
https://codesandbox.io/s/o4940jx1y

**Note:** importing polyfills is a must.

AndyOGo:
Luckily we could finally fix this :)

pavestru:
This is super awesome! However, I have to restore the branch as our build process depends on it. I will delete it as soon as we rewrite all our components using `withReact`, so bear with me a little while, please :-)

AndyOGo:
Sorry for being to quick...

PR #444, opened 2018-05-28
Bugfix/lib external helpers undefined

### LINKED ISSUE #430: with-react only avaiable as src. should be transpiled for dist

 - [ ] Bug
 - [x] Feature

## Expected Behavior

with-react only avaiable as src. should be transpiled for dist

AndyOGo:
fixes #430 

Fixes `babelHelpers` undefined

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,6 +88,7 @@
     "babel-plugin-transform-custom-element-classes": "^0.1.0",
     "babel-plugin-transform-es2015-destructuring": "^6.23.0",
     "babel-plugin-transform-object-rest-spread": "^6.26.0",
+    "babel-plugin-transform-runtime": "^6.23.0",
     "babel-preset-env": "~1.6.1",
     "babel-preset-react": "^6.24.1",
     "chalk": "^2.4.0",
@@ -135,6 +136,7 @@
   },
   "dependencies": {
     "@webcomponents/webcomponentsjs": "^1.0.18",
+    "babel-runtime": "^6.26.0",
     "classnames": "^2.2.5",
     "core-js": "~2.5.1",
     "innersvg-polyfill": "0.0.2",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -32,7 +32,7 @@
     "build-icons-prod": "node stack/tasks/icons.js PROD",
     "build-dev": "node stack/tasks/cleanup.js DEV && npm run build-css-dev && npm run build-html-dev && npm run build-js-dev && npm run build-icons-dev",
     "build": "node stack/tasks/cleanup.js PROD && node stack/tasks/build.html.js PROD && node stack/tasks/build-examples.js PROD && npm run build-css-prod && npm run build-js-prod && npm run build-lib && npm run build-icons-prod && node stack/tasks/copy-dist-to-docs.js",
-    "build-lib": "babel src --out-dir lib --ignore src/demos",
+    "build-lib": "babel src --out-dir lib --ignore src/demos --plugins transform-runtime",
```
AndyOGo: @LucaMele 
@pavestru 
For rollup we used the [`runtimeHelpers` option of rollup-plugin-babel](https://github.com/rollup/rollup-plugin-babel#helpers)
But we should really clean this up and have all within config files, not custom config overwrites in JS

https://github.com/axa-ch/patterns-library/search?q=runtimeHelpers&unscoped_q=runtimeHelpers

AndyOGo:
still getting:

```sh
TypeError: WebComponent is not a constructor
```

for:

```js
var node = new WebComponent();
```

package.json
LucaMele: generally, the whole build need to be cleanup up. it was a quick solution for a limited scope. :)

AndyOGo: Yeah, it really takes more time to craft a stable build and document all of it - not meant as accusation

LucaMele: totally agree :)

AndyOGo:
@LucaMele 
@pavestru 
Since we still get this **TypeError** of `WebComponent is not a constructor` we need to find another way of getting the `tagName` used by `withReact`.
I will create another issue and PR for this.

LucaMele:
@AndyOGo use the plugins: `"transform-custom-element-classes", "transform-class-properties",` and test again

LucaMele:
`"build-lib": "babel src --out-dir lib --ignore src/demos --plugins transform-runtime transform-custom-element-classes transform-class-properties",`

LucaMele:
ok yes. weird

AndyOGo:
Yeah, I thought this at first to, but we actually use them already.

It's again this fu**** issue of asynchronous vs. synchronous registering, instantiation and rendering...

PR #452, opened 2018-05-30
add optional path prefix removal for better integration in AEM and ot…

LucaMele:


src/js/abstract/property-exists-exception.js
```diff
@@ -1,6 +1,6 @@
 class PropertyExistsException extends Error {
   constructor(key, node, ...params) {
-    const message = `Property >>>${key}<<< exists at ${node.nodeName}#${node._id} and evaluates to -> ${node[key]}`;
+    const message = `Native Property >>>${key}<<< exists already at ${node.nodeName}#${node._id} and evaluates to -> ${node[key]}`;
```
AndyOGo: Let's make it even more explicit, by wording it like:
```js
`Native DOM Property key >>>${key}<<< exists already at ${node.nodeName}#${node._id} and evaluates to -> ${node[key]}.
Please consult the HTML spec for inherited DOM properties:
- Element Interface https://developer.mozilla.org/en-US/docs/Web/API/Element#Properties
- Node Interface https://developer.mozilla.org/en-US/docs/Web/API/Node#Properties`
```

LucaMele: ??

AndyOGo: And I would also update the docs and add inline comments describing the problem

LucaMele: So what should i write
Why do i have to consult
please can you specify what is the issue with the updated text

PR #485, opened 2018-06-18
Bugfix/dropdown native width fix

jan-capiak:
Fixes # (issue).

 - Very long text in m-dropdown native is cut off
 - The reason is the margin-left of the .m-dropdown__select-icon class

- Before:
before

- After:
after

## Type of change

 # How Has This Been Tested?
 
 
 
 # Checklist:
 

src/components/m-dropdown/mixins/dropdown.scss
```diff
@@ -129,7 +129,7 @@
   float: right;
   height: $height;
   padding: 0 $indent;
-  margin-left: -100%;
+  margin-left: $indent;
```
AndyOGo: @jan-capiak 
Thanks for your PR.

May I ask your to provide positive test for firefox, and mobile on iOS and Android.
The point is that some firefox for example does not support `text-indent` at styled window-objects.

AndyOGo: Furthermore the icon is not meant to consume space, now it takes up the size of `$indent`

AndyOGo: And it shall overlay any browser-intrinsic dropdown icon, which can't be always removed by CSS.

AndyOGo: The overflow of `<select/>` is made intentionally for cross browser support.

screen shot 2018-06-18 at 15 21 02

src/components/m-dropdown/mixins/dropdown.scss
```diff
@@ -145,6 +145,11 @@
     position: relative;
     right: auto;
   }
+  
+  @include target(moz) {
```
AndyOGo: @LucaMele 
@jan-capiak 
This should fix the doubled margin issue at firefox.

We still need a positive test, that layouts don't blow up on mobiles

AndyOGo:
@jan-capiak 
@LucaMele 
Can we quickly test this? I fixed the firefox doubled `margin`/`text-indent` bug.
The rest should be fine, would be keen to merge it this way.

Issue #515, opened 2018-07-09
Using it in axa.ch and without waiting ondomready crahes the flattened DOM


LucaMele:
this: 

```html
<axa-cookie-disclaimer fixed button-name="Akzeptieren" title="Datenschutzbestimmungen" classes="o-cookie-disclaimer">
  <p>
    Wir verwenden Cookies und Analyse Tools, um die Nutzerfreundlichkeit der Internet-Seite zu verbessern und die Werbung von AXA und Werbepartnern zu personalisieren. Weitere Infos:
    <axa-link color="white" icons-path-prefix="/etc/clientlibs/axa/frontend-lib/images/icons-v2.svg#" href="/content/axa/de/informationen/datenschutz.html" arrow>
      Datenschutz
    </axa-link>
  </p>
</axa-cookie-disclaimer>
```

LucaMele:
generates:

```
<axa-cookie-disclaimer fixed="" button-name="Akzeptieren" title="Datenschutzbestimmungen" classes="o-cookie-disclaimer" class=" o-cookie-disclaimer"><article class="o-cookie-disclaimer o-cookie-disclaimer--fixed"><axa-container><article class="o-container "><div class="o-cookie-disclaimer__container o-cookie-disclaimer__container--lx"><h1 class="o-cookie-disclaimer__title">Datenschutzbestimmungen</h1></div><div class="o-cookie-disclaimer__container o-cookie-disclaimer__container--rx"><axa-button ghost="ghost" classes="js-cookie-disclaimer__button" color="white" tag="button"><button type="button" class="m-button js-button js-cookie-disclaimer__button m-button--white m-button--ghost">Akzeptieren</button></axa-button></div></article></axa-container></article>
    <p>
        Wir verwenden Cookies und Analyse Tools, um die Nutzerfreundlichkeit der Internet-Seite zu verbessern und die Werbung von AXA und Werbepartnern zu personalisieren. Weitere Infos: 
[...]

PR #516, opened 2018-07-09
Temporary fix for broken disclaimer

LucaMele:
temp solution for #515


LucaMele:
Totally agree. I hope we can find a day where we can sit togehter and learn, test and understand it the right way

PR #533, opened 2018-07-16
Feature/datepicker molecule continuation

LucaMele:
Fixes # (issue).

new component molecules preparation for organism datepicker

 # How Has This Been Tested?

Browserstack all main devices


.stylelintrc
```diff
@@ -147,6 +147,7 @@
       "ignoreAtRules": ["if", "else"]
     } ],
     "at-rule-no-unknown": null,
+    "selector-type-no-unknown": null,
```
AndyOGo: Do we really want to disable it for all?
They support ignoring `custom-elements`:
https://stylelint.io/user-guide/rules/selector-type-no-unknown/#ignore-custom-elements-default-namespace

src/components/m-datepicker-body/_template.js
```diff
@@ -0,0 +1,14 @@
+import html from 'nanohtml';
+
+const renderCells = (store) => {
+  if (!store) {
+    return [];
+  }
+  return store.getCells();
```
AndyOGo: shouldn't that be done outside of template?

src/components/m-datepicker-body/index.scss
```diff
@@ -0,0 +1,64 @@
+@import "../../styles/settings/colors";
+
+$datepicker-width: 309px !default;
+
+$datepicker-padding-left-right: 30px !default;
```
AndyOGo: I suggest call it `$datepicker-padding-x` (for x-axis and y for y-axis in case of top/bottom)

src/components/m-datepicker/_preview.html
```diff
@@ -1,4 +1,6 @@
-<!--axa-datepicker button-ok="ok" button-cancel="abbrechen" classes="m-datepicker"></axa-datepicker-->
-<axa-container>
-  <h1>Coming soon...</h1>
-</axa-container>
+<axa-m-datepicker button-ok="ok" button-cancel="abbrechen" locale="de-ch" classes="m-datepicker"></axa-m-datepicker>
+<axa-m-datepicker output-iso button-ok="Ja" button-cancel="abbrechen" classes="m-datepicker"></axa-m-datepicker>
```
AndyOGo: how about validations like, min-date, max-date?
Or will this be part of the organism, still I guess somehow it needs to be incorporated here 🤔

src/components/m-datepicker/_template.js
```diff
@@ -2,40 +2,97 @@ import html from 'nanohtml';
...
+            name: year, url: '#', value: year,
+          })))}">
         </axa-dropdown>
       </div>
       <div class="m-datepicker__weekdays">
-      ${getLocalWeekdayArray(locale).map(day => html`<p class="m-datepicker__weekdays__day">${day}</p>`)}
+        ${getLocalWeekdayArray(locale).map(day => html`<p class="m-datepicker__weekdays__day">${day}</p>`)}
       </div>
-      <div class="m-datepicker__body" data-calendar-area="month"></div>
+      <axa-datepicker-body allowed-years="${JSON.stringify(specificYears)}" year="${startDate.year}" month="${startDate.month}" class="js-datepicker__datepicker-body" locale="${locale}"></axa-datepicker-body>
```
AndyOGo: BTW. we still have this problem of inline `JSON.stringifiy` usage ... other topic.

src/components/m-datepicker/index.js
```diff
@@ -5,17 +5,23 @@ import styles from './index.scss';
...
 
 class AXADatepicker extends BaseComponentGlobal {
-  static tagName = 'axa-datepicker'
+  static tagName = 'axa-m-datepicker'
 
   // Specify observed attributes so that attributeChangedCallback will work,
   // this is essential for external re-rendering trigger.
-  static get observedAttributes() { return ['classes', 'button-ok', 'button-cancel']; }
+  static get observedAttributes() {
+    return ['classes', 'button-ok', 'button-cancel', 'locale', 'value', 'selected-index', 'start-year', 'start-month', 'lower-end-year', 'higher-end-year', 'output-iso'];
```
AndyOGo: `selected-index` of what? year, month, both?

src/components/m-datepicker/index.scss
```diff
@@ -1,49 +1,96 @@
 @import "../../styles/settings/colors";
 
-$datepicke-height: 322px !default;
-$datepicke-width: 309px !default;
+$datepicker-height: 385px !default;
+$datepicker-width: 309px !default;
+
+$datepicker-margin-top-bottom: 20px !default;
```
AndyOGo: I suggest `y` for y-axis (top-bottom)

src/components/m-datepicker/index.scss
```diff
@@ -1,49 +1,96 @@
...
+$datepicker-margin-left-right: 30px !default;
```
AndyOGo: and `x` for x-axis

src/js/abstract/hocs/with-render.js
```diff
@@ -107,7 +107,7 @@ const withRender = Base =>
             this.childrenFragment.children = Array.from(this.childrenFragment.childNodes).filter(node => node.nodeType === 1);
           }
 
-          const items = template(this._props, this.childrenFragment);
+          const items = template(this._props, this.childrenFragment, this);
```
AndyOGo: cool

src/components/m-datepicker/_template.js
```diff
@@ -2,40 +2,97 @@ import html from 'nanohtml';
...
+  startMonth = TODAY,
+  lowerEndYear,
+  higherEndYear,
+}) => {
+  const startDate = getStartDate({ lowerEndYear, higherEndYear }, startYear, startMonth);
+  const specificYears = getSpecificYears({ lowerEndYear, higherEndYear });
+  return html`
+    <article class="${classes} m-datepicker__article">
+      <div class="">
+        <axa-dropdown data-month="true" class="m-datepicker__dropdown m-datepicker__dropdown__month js-datepicker__dropdown__month"
```
AndyOGo: I would use the native dropdown on mobiles

AndyOGo: note will need to change dynamically based on `resize`/`reorientation`, device detection...

AndyOGo:
hmh, I can't change the date 🤔 

datepicker-click-day

.stylelintrc
LucaMele: i think so, but open for discussions

src/components/m-datepicker-body/_template.js
LucaMele: where? cause the FB u gave last time was to do it inside the template for DOM diffing and the cells are filled from the "model" inside datepicker-body.js. 

Open for suggestions :)

src/components/m-datepicker-body/index.scss
LucaMele: agree

src/components/m-datepicker/_preview.html
LucaMele: Validation works here that we cannot choose more dates than allowed in the dropdown (so no validation cause u cannot choose what is not allowed).

The input text instead will have validation for user manual input

LucaMele: example: `<axa-m-datepicker lower-end-year="1980" higher-end-year="2018" start-month="11" locale="it-ch" button-ok="ok" button-cancel="abbrechen" classes="m-datepicker"></axa-m-datepicker>`

LucaMele: the calendar wont allow u to select 2019 for example

src/components/m-datepicker/_template.js
LucaMele: i don't think they are design-compliant. But is a very good point. What if we create an issue and verify it with Design and skip the implementation of it in this PR?

@NoahLiechti can u please verify it?

src/components/m-datepicker/index.js
LucaMele: it was needed for the datepicker-body. Here is wrong. @NoahLiechti please remove it

src/components/m-datepicker/index.scss
LucaMele: agree

src/components/m-datepicker/index.scss
LucaMele: agree

src/components/m-datepicker-body/_template.js
AndyOGo: hmh, I thought that the `store` will provide those cells by default.

src/js/abstract/hocs/with-render.js
LucaMele: 😇

LucaMele:
@AndyOGo u have to press on the "ok" or "ja" button to select the date. It wont select by just pressing on the date (design requirement)

src/components/m-datepicker/_preview.html
AndyOGo: Okay, I see.
It would be nice though, to describe a full date, including days for min/max or start/end

LucaMele:
@AndyOGo we tested all our required devices for all swiss locales needed and some more. Therefore its ok IMHO and no polyfil is needed. we can always add it in case we have issues. Also Chines did work btw :)

src/components/m-datepicker/_template.js
AndyOGo: Fine with me, I think it's in general good practice to use native UIs on mobiles.
Actually we should also question to utilize the build in `<input type="date" />` and check how we can be compliant 🤔

AndyOGo:
@LucaMele Though this design suggests the wrong selected date IMHO.
Personally I would question that, but if it is what they want, then it's okay for me.

LucaMele:
wrong selected date is not possible to select cause the datepicker wont allow to choose it. I will double check thought with design

src/components/m-datepicker/_template.js
LucaMele: we spoke about the <input type="date" /> option as well and i will investigate it when we do the organism

src/components/m-datepicker/_preview.html
LucaMele: we will double check the start-end date. that would mean that on the calendar some current month days will be grey as well and therefore need a design approval

src/components/m-datepicker-body/_template.js
LucaMele: they have to be calculated related from date object and those cells need to be used in the model and in the template

src/components/m-datepicker-body/index.scss
NoahLiechti: done

LucaMele:
@NoahLiechti will ask Ken about the native dropdown and also if a certain range of days can be set. We will treat it as enhancement and merge this so far as its a HUGE pr.

PR #537, opened 2018-08-09
Commercial hero banner

nakovskijosif:
Commercial Hero Banner Organism added.

## Type of change
- New feature

 # Checklist:

### DIFF
--- a/src/components/o-commercial-hero-cover/_preview.html
+++ b/src/components/o-commercial-hero-cover/_preview.html
@@ -1,0 +1,37 @@
+<axa-commercial-hero-cover
+  src="https://i.pinimg.com/originals/b5/4c/97/b54c972b2a1138d85af5692431c02a2c.jpg"
+  gradient="black"
+  content-align="left">
+
+    <section class="js-commercial-hero-cover__first">
+      <h2 class="o-commercial-hero-cover__subtitle">Der Subtitel steht über dem Titel</h2>
+      <h2 class="o-commercial-hero-cover__title">Willkommen auf der AXA Landingpage!</h2>
+      <p class="o-commercial-hero-cover__text">Guten Tag! Ich bin der Chatbot Glasschaden der AXA. Ich kann für Sie mit wenigen Angaben einen Glasbruchschaden an</p>
+      <axa-button tag="a" href="http://www.google.com" class="o-commercial-hero-cover__button" color="red" motion>Berechnen Sie den Preis sofort</axa-button>
+    </section>
+
+    <section class="js-commercial-hero-cover__second">
+      <p class="o-commercial-hero-cover__text">Guten Tag! Ich bin der Chatbot Glasschaden der AXA. Ich kann für Sie mit wenigen Angaben einen Glasbruchschaden an2222Guten Tag! Ich bin der Chatbot Glasschaden der AXA. Ich kann für Sie mit wenigen Angaben einen Glasbruchschaden an Guten Tag! Ich bin der Chatbot Glasschaden der AXA. Ich kann für Sie mit wenigen Angaben einen Glasbruchschaden an2222Guten Tag! Ich bin der Chatbot Glasschaden der AXA. Ich kann für Sie mit wenigen Angaben einen Glasbruchschaden an Guten Tag! Ich bin der Chatbot Glasschaden der AXA. Ich kann für Sie mit wenigen Angaben einen Glasbruchschaden an2222Guten Tag! Ich bin der Chatbot Glasschaden der AXA. Ich kann für Sie mit wenigen Angaben einen Glasbruchschaden an</p>
+    </section>
+
+</axa-commercial-hero-cover>
+<axa-commercial-hero-cover
+  src="https://i.pinimg.com/originals/b5/4c/97/b54c972b2a1138d85af5692431c02a2c.jpg"
+  gradient="white"
+  content-align="right">
+    <h2 class="o-commercial-hero-cover__subtitle">Der Subtitel steht über dem Titel</h2>
+    <h2 class="o-commercial-hero-cover__title">Willkommen auf der AXA Landingpage!</h2>
+    <p class="o-commercial-hero-cover__text">Guten Tag! Ich bin der Chatbot Glasschaden der AXA. Ich kann für Sie mit wenigen Angaben einen Glasbruchschaden an</p>
+    <axa-button tag="a" href="http://www.google.com" class="o-commercial-hero-cover__button" color="red" motion>Berechnen Sie den Preis sofort</axa-button>
+
+</axa-commercial-hero-cover>
+<axa-commercial-hero-cover
+  src="https://i.pinimg.com/originals/b5/4c/97/b54c972b2a1138d85af5692431c02a2c.jpg"
+  gradient="white"
+  content-align="left">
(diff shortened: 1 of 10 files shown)


LucaMele:
bildschirmfoto 2018-08-09 um 11 16 17

src/components/o-commercial-hero-cover/_preview.html
```diff
@@ -0,0 +1,43 @@
+<axa-commercial-hero-cover
+  src="https://i.pinimg.com/originals/b5/4c/97/b54c972b2a1138d85af5692431c02a2c.jpg"
+  gradient="black"
+  content-align="left">
+
+    <section class="js-commercial-hero-cover__first">
+      <h2 class="o-commercial-hero-cover__subtitle">Der Subtitel steht über dem Titel</h2>
```
LucaMele: Please use Demo Texts in english as also AXA global can use the component.

src/components/o-commercial-hero-cover/_template.js
```diff
@@ -0,0 +1,26 @@
+import html from 'nanohtml';
+
+export default ({
+                  src,
+                  alt = 'Commercial Hero cover',
+                  gradient = 'white',
+                  contentAlign = 'left'
+}, childrenFragment) => html`
+  <header class="o-commercial-hero-cover__header ${gradient === "white" ? "gradient-white" : "gradient-black"}">
+    <picture class="m-picture o-commercial-hero-cover__picture">
```
LucaMele: as we dont have a molecule in the plib and we need it for AEM, please use a property for setting classes from outiside.

src/components/o-commercial-hero-cover/_template.js
```diff
@@ -0,0 +1,26 @@
...
+      <img class="m-picture__image o-commercial-hero-cover__picture__image a-focal-point a-focal-point--center"
+      src="${src}" alt="${alt}" data-object-fit="cover"
```
LucaMele: if i remember correctly, this is the polyfill used for IE11 for object fitting.. and if the picture element does not have srcsets maybe the picture tag is not necessary.. but in aem we need it.. 

my suggestion, lets keep the picture tag and add the polyfill as import to this template as well so that the data- is correct. I will make an issue in the plib for creating the picture element

src/components/o-commercial-hero-cover/_template.js
LucaMele: a prop like "classes" for example

LucaMele: or picture-classes might be better

src/components/o-commercial-hero-cover/_template.js
```diff
@@ -0,0 +1,26 @@
...
+      data-object-position="center">
+    </picture>
+    <axa-container class="o-commercial-hero-cover__info">
+      <axa-row>
+        <axa-col size="sm-12 lg-6" order="${contentAlign === "left" ? "0" : "1"}">
+          ${childrenFragment.querySelector(".js-commercial-hero-cover__first") || childrenFragment}
```
LucaMele: i like this logic, that section is not mandatory

src/components/o-commercial-hero-cover/index.scss
```diff
@@ -0,0 +1,115 @@
...
+@import "../../styles/grid/grid-mixins";
+@import "../../styles/mixins/respond";
+@import "../../styles/typo/typo-mixins";
+@import "../../styles/mixins/word-break";
+@import "../m-button/mixins/button";
+
+.o-commercial-hero-cover {
+  display: block;
+
+  .o-commercial-hero-cover__header {
```
LucaMele: i like that the bem classes works only in the main block element, but this can lead to the following problem:

If a user wants to override some styling, he also need to have the same specificty + 1 which would lead to unneded code and more difficulty in maintaing the code.

Therefore, my suggestion is to not put the elements inside the block and relay on naming to avoid collision and pollution

src/components/o-commercial-hero-cover/index.scss
```diff
@@ -0,0 +1,115 @@
...
+    position: relative;
+    min-height: 256px;
```
LucaMele: ok am sorry i forgot to say it, but we use variables to be defined on top of the scss file instead of just values inside the scss for maintainability reasons. Could you fix that please and sorry.

src/components/o-commercial-hero-cover/index.scss
```diff
@@ -0,0 +1,115 @@
...
+    flex-direction: column;
+    justify-content: flex-start;
+    background: linear-gradient(180deg, transparent 200px, white 0);
+    @include respond-up(md) {
+      min-height: 341px;
+    }
+
+    @include respond-up(lg) {
+      height: 490px;
+      &.gradient-white {
```
LucaMele: this class is not BEM like named

src/components/o-commercial-hero-cover/index.scss
```diff
@@ -0,0 +1,115 @@
...
+        background: linear-gradient(270deg, fade_out($color-prim-white, 1) 30%, fade_out($color-prim-white, 0.2) 60%, $color-prim-white 100%);
+        .o-commercial-hero-cover__info {
+          color: $color-text;
+        }
+      }
+      &.gradient-black {
```
LucaMele: dtto

src/components/o-commercial-hero-cover/index.scss
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  .o-commercial-hero-cover__picture {
+    display: block;
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: 100%;
+    z-index: -1;
+    content: '';
```
LucaMele: is content:'' needed?

src/components/o-commercial-hero-cover/_template.js
Ben-Zahler: this is not only for IE11, it's how we fit all images in AEM.
For most browsers, this works with CSS, but there is a polyfill for IE11.

I think we should simply remove the a-focal-point classes here and add them in AEM.

Ben-Zahler: ah, no, this is not going to work... we need these classes.

src/components/o-commercial-hero-cover/index.scss
Ben-Zahler: done

src/components/o-commercial-hero-cover/index.scss
Ben-Zahler: done dtto

src/components/o-commercial-hero-cover/_template.js
Ben-Zahler: done

src/components/o-commercial-hero-cover/_template.js
Ben-Zahler: object-fit property passed as an argument and polyfill for object-fit added.
we should be able to reuse the object-fit property once we implement the picture element in the plib

src/components/o-commercial-hero-cover/_preview.html
nakovskijosif: Done

LucaMele:
@nakovskijosif and @Ben-Zahler , thank you very much! very good work! As soon as you merge, i will release

PR #550, opened 2018-08-22
trigger always a custom even on axa change fi the WC has a value prop…

### LINKED ISSUE #376: Trigger custom events of every component

 - [x] Feature

## Expected Behavior

I expect that every custom element, triggers any internal events necessary for the outer world as a custom event, which can be `defaultPrevented`, `propagationStopped`, etc.

And I expect all available events for consumers to be properly documented (with [jsdoc `@event`](http://usejsdoc.org/tags-event.html)).

All event names shall share a consistent naming, like `axa-click`. **Important** those names have to be simple, no fancy separators shall be used for integration purposes.

**Related:**
 - Integration POC #330 

**Affected Components**
- `<axa-heckbox>`
- `<axa-choice>`
- `<axa-forminput>`
- `<axa-radio>`
- `<axa-accordion>`
- `<axa-accordion-item>`
- `<axa-button>`
- `<axa-dropdown>`
- `<axa-footer-languages>`
- `<axa-footer-legals>`
- `<axa-footer-social>`
- `<axa-header-burger>`
- `<axa-header-languages>`
- `<axa-header-logo>`
- `<axa-header-mobile-languages>`
- `<axa-header-mobile-navigation>`
- `<axa-header-mobile-others>`
- `<axa-header-navigation>`
- `<axa-header-others>`
- `<axa-header-search>`
- `<axa-header-sub-navigation>`
- `<axa-link>`

## Actual Behavior

Not a single component implements any custom event for consumers.

LucaMele:
Fixes #376 and minor issues

### DIFF
--- a/src/demos/todomvc/todo-item.jsx
+++ b/src/demos/todomvc/todo-item.jsx
@@ -4,10 +4,12 @@ import withReact from '../../js/with-react';
 import AXAButton from '../../components/m-button';
 import AXAIcon from '../../components/a-icon';
 import AXACheckbox from '../../components/a-checkbox';
+import AXADropDown from '../../components/m-dropdown';
 
 const AXAButtonReact = withReact(AXAButton);
 const AXAIconReact = withReact(AXAIcon);
 const AXACheckboxReact = withReact(AXACheckbox);
+const AXADropDownReact = withReact(AXADropDown);
 
 const ESCAPE_KEY = 27;
 const ENTER_KEY = 13;
@@ -119,6 +121,15 @@ class TodoItem extends Component {
     return (
       <li className="m-todo__item">
         <div className="m-todo__wrap">
+          <AXADropDownReact
+            title="Switch status"
+            value={completed ? 'selected' : 'unselected'}
+            onAxaChange={(ev) => {
+              todo.completed = ev.detail === 'unselected';
+              onToggle(todo);
+            }}
+            items={[{ name: 'Selected', value: 'selected' }, { name: 'Unselected', value: 'unselected' }]}
+          />
           <AXACheckboxReact checked={completed} name="completed" onChange={() => onToggle(todo)} />
 
           <input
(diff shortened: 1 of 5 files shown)


src/js/abstract/hocs/with-update.js
```diff
@@ -144,6 +146,11 @@ const withUpdate = Base =>
       } else { // delete attribute
         this[key] = null;
       }
+
+      // if value is updated, we presume that an axa on change event have to be triggered
+      if (name === 'value' && newValue !== null && oldValue !== newValue) {
```
AndyOGo: Check for updating is done above by `shouldUpdateCallback`.

Hmh do we really want to fire this for all value attributes?

src/demos/todomvc/todo-item.jsx
```diff
@@ -119,6 +121,15 @@ class TodoItem extends Component {
     return (
       <li className="m-todo__item">
         <div className="m-todo__wrap">
+          <AXADropDownReact
+            title="Switch status"
+            value={completed ? 'selected' : 'unselected'}
```
AndyOGo: I appreciate the `value` prop, I think we should follow react's principle here for all input components

LucaMele: which is?

LucaMele: i emulated all input principles.. all user-inputs HTML tags have value as attribute to set value, and this is what i emulated. What do you mean with "we should follow react's principle"? The sentence as it is unfortunately says nothing

AndyOGo: Sorry I meant to reference all [React's Form concepts](https://reactjs.org/docs/forms.html):
- [Controlled Components](https://reactjs.org/docs/forms.html#controlled-components)
- `[value=""]` attribute over HTML spec for [`<textarea>`](https://reactjs.org/docs/forms.html#the-textarea-tag), [`<select>`](https://reactjs.org/docs/forms.html#the-select-tag)

LucaMele: Ahh ok.. well i think is quite similar how we started now isn't it?

LucaMele: but anyway as best practice for future design desicions, we should keep to HTML standard/way to do and not react

PR #551, opened 2018-08-23
Bugfix/aem 3792 landingpage small fixes

nakovskijosif:
Fixes issue #552 
additional minor UI improvements and fixes

## Type of change

 # How Has This Been Tested?
 Tested on local dev environment on google chrome and IE11(commercial hero banner fixes)
 # Checklist:
 
### DIFF
--- a/src/components/m-policy-feature-item/_template.js
+++ b/src/components/m-policy-feature-item/_template.js
@@ -1,8 +1,9 @@
 import html from 'nanohtml';
 
-export default ({ src, alt = 'Policy feature icon', title, description }) => html`
+export default ({ src, alt = 'Policy feature icon', title, description }, childrenFragment) => html`
   <article class="m-policy-feature-item__content">
-    <img class="m-policy-feature-item__icon" src="${src}" alt="${alt}">
+    ${src && html`<img class="m-policy-feature-item__icon" src="${src}" alt="${alt}">`}
+    ${childrenFragment.querySelector('.js-m-policy-features-icon') || ''}
     ${title && html`<h1 class="m-policy-feature-item__title">${title}</h1>`}
     ${description && html`<p class="m-policy-feature-item__description">${description}</p>`}
   </article>
--- a/src/components/o-policy-features/_preview.html
+++ b/src/components/o-policy-features/_preview.html
@@ -1,27 +1,27 @@
 <axa-policy-features title="A 5 star car insurance with affordable premium services">
 
   <axa-policy-feature-item title="Get Discount"
-                           src="https://dev.w3.org/SVG/tools/svgweb/samples/svg-files/beacon.svg"
                            alt="Discount Svg Icon"
                            description="We reward safe drivers : 75% no claims discount + an extra 10% off if you get a quote online">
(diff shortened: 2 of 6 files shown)


src/components/m-policy-feature-item/_template.js
```diff
@@ -1,8 +1,9 @@
 import html from 'nanohtml';
 
-export default ({ src, alt = 'Policy feature icon', title, description }) => html`
+export default ({ src, alt = 'Policy feature icon', title, description }, childrenFragment) => html`
   <article class="m-policy-feature-item__content">
-    <img class="m-policy-feature-item__icon" src="${src}" alt="${alt}">
+    ${src && html`<img class="m-policy-feature-item__icon" src="${src}" alt="${alt}">`}
```
AndyOGo: This could print falsy values

src/components/o-policy-features/_preview.html
```diff
@@ -1,27 +1,27 @@
...
...85-2.58-4.26zm-2.05-.19c-3.44-.46-5.17-3.07-6-5.63l2 .6c2.04.49 3.63 2.84 3.99 5.02zm-79.86-3.46h-3v-4.73a7.84 7.84 0 0 1 3 4.73zm-4 2h5.06a1 1 0 0 0 1-1 9.25 9.25 0 0 0-3.5-7c4.47-1.35 10-5.65 14.3-8.94 1.6-1.23 3-2.31 4-3l2.06-.34c8.24-1.37 18.49-3.07 23.49-.78 5.47 2.51 16.63 12.25 16.74 12.35a1 1 0 0 0 .37.21l12 3.62c1.35 5.39 4.69 8.35 9.41 8.35.42 0 1.53 0 1.53 2.38v5.75H82.31a9.43 9.43 0 0 0-16.81 0H43V45h24.78a1 1 0 0 0 .68-1.74A161.15 161.15 0 0 0 51.55 30l-.17-.09c-2.85-1.2-9.77-.88-20.55.94a1 1 0 0 0-.54.28c-.84.84-1.86 1.86-2.94 3l-.11.12c-2.75 2.8-6 6.18-8.31 9.13a1 1 0 0 0 .79 1.62H41v13.85h-9.81a9.43 9.43 0 0 0-16.81 0H5V48.72a1.53 1.53 0 0 1 1.53-1.53zM21.84 43c1.59-1.86 3.41-3.82 5.16-5.64V43zM29 35.28l2.48-2.5c3.89-.65 7-1.05 9.52-1.29V43H29zm14-4c4.56-.3 6.61.05 7.53.41A156.85 156.85 0 0 1 65.14 43H43zM10.47 64.22l-4.21-1.7A2 2 0 0 1 5 60.85h8.59a9.71 9.71 0 0 0-.33 2.49v.89zM22.79 71a7.64 7.64 0 1 1 7.5-7.64 7.58 7.58 0 0 1-7.5 7.64zm9.46-6.77v-.86a9.71 9.71 0 0 0-.25-2.52h32.74a9.71 9.71 0 0 0-.33 2.49v.86zM73.91 71a7.64 7.64 0 1 1 7.5-7.64 7.58 7.58 0 0 1-7.5 7.64zm16.44-6.75h-7v-.89a9.71 9.71 0 0 0-.33-2.49H93v.72a2.66 2.66 0 0 1-2.65 2.63z"></path></g></svg>
```
AndyOGo: Guess this is a requirement of how icons are implemented in AEM?

Ben-Zahler: this allows to pass svg from AEM as in similar components with the full SVG source, SVG is stored in DAM.
Having an svg rather than  tag allows to color the svg correctly

src/components/m-policy-feature-item/_template.js
Ben-Zahler: no, I added this because the src attribute is now optional in the same way as for the other attributes... is this a security risk? otherwise, I think anyone invoking this with "src=false" should not be surprised if the component is not working as expected :-)

AndyOGo: that's true.
I'm just curious how falsy expressions (`false`, `null`, `undefined`, etc.) are rendered to DOM. I'm not sure if they get a string representation or not 🤔 
I would be glad if we can test this out quickly to be sure.

Ben-Zahler: I checked: values are always passed as String (Luca actually mentioned that's the behaviour).
So they are evaluated as true and then an invalid url is printed, in my opinion the calling module should be responsible for validating, agreed?

AndyOGo: Hey Ben,
That is correct HTML attributes are always strings, we mention this fact in the contribution file [here](https://github.com/axa-ch/patterns-library/blob/develop/CONTRIBUTING.md#custom-elements)
> HTML attributes (always 'string') VS DOM properties ([first class props](https://github.com/axa-ch/patterns-library/blob/develop/CONTRIBUTING.md#first-class-props))

This and other issue are **always fixed inside** the component not outside - In generall every component validates itself.
Therefore I created a bunch of helper functions, one of is `getAttribute` and `getAttributes`, which turns aforementioned strings into [first class props](https://github.com/axa-ch/patterns-library/blob/develop/CONTRIBUTING.md#first-class-props) by utilising `JSON.parse`.

The issue with src is, if it's not used, it will be `undefined` which is a falsy value.
And I'm curious how `nanohtml` renders `undefined` (I expect that it skips it, if not we have to prevent it from being rendered - same for all falsy values).

AndyOGo: Here is code snippet testing just that case, unfortunately it renders `undefined` as a string...
https://codesandbox.io/s/2pr52l26my

AndyOGo: I updated the example with a working one:
[Edit 2pr52l26my](https://codesandbox.io/s/2pr52l26my)

AndyOGo: @LucaMele 
Since it seems to be a generic behavior of `nanohtml` to render falsy values as strings or to always use the `toString()` representation for any value, we might should consider adding a helper function for hiding falsy values by outputting empty strings instead, like:

```js
import nanohtml from 'nanohtml';

const hideFalsy = (value) => value || ''; // may we need to exclude `0` and other falsy values

export default () => nanohtml`<div>
  undefined should not render by ${undefined || ''} <br>
  with a helper function: ${hideFalsy(undefined)}, ${hideFalsy(null)}, ${hideFalsy(false)}, 
</div>`;
```

Ben-Zahler: @LucaMele @AndyOGo my 2 cents: why should we add complexity to fix something that is obviously an incorrect invocation and does not break any functionality?
If you see any way on how this can be exploitet for security issues, I change my mind immediately

AndyOGo: @Ben-Zahler 
I think we speak about different things here.
As I understood `src` is optional, so the user can omid it, right?
If `src` is omitted it's value will be `undefined`, thus the expression `${src && html``}` will lead to a falsy value.
And `nanohtml` prints falsy values.
This is what I mean, user is allowed to omit `src` and sees `false`, `undefiend` or what ever as rendered output. It's not a security concern, just presentation.

Ben-Zahler: @AndyOGo the case you're writing about is actually part of _preview.html, where the "src" is omitted:

  <axa-policy-feature-item title="Get Discount"
                           alt="Discount Svg Icon"
                           description="We reward safe drivers : 75% no claims discount + an extra 10% off if you get a quote online">

the  is not rendered for me

AndyOGo: @Ben-Zahler 
Well I just updated my extracted test case and it does render 😲 
https://codesandbox.io/s/6nxqyj00nz

Though may my test is wrong or I missed something 🤔

PR #558, opened 2018-08-31
WIP - Feature/datepicker organism

LucaMele:
DO NOT REVIEW YET


src/components/a-input/index.js
```diff
@@ -26,6 +29,9 @@ class AXAInput extends BaseComponentGlobal {
       'a-input--disabled': disabled,
     });
   }
+  didRenderCallback() {
```
LucaMele: dont forget to cleanup events like in dropdown

src/components/o-datepicker/index.scss
```diff
@@ -1,5 +1,25 @@
...
+
+  z-index: $zindex-popover;
+
+  display: none;
+  background-color: $color-prim-white;
+}
 
-  // IMPORTANT: make sure to deal with inherited CSS properties here, like text-align!
+.o-datepicker__calender--move-up {
+  // TODO $control-size !!!!! ausrechnen !!!!!!!!
```
LucaMele: TODO

src/components/o-datepicker/js/datepicker.js
```diff
@@ -0,0 +1,135 @@
...
+    this.wcNode = wcNode;
+    this._value = '';
+    this._locale = '';
+    this._localValue = '';
+  }
+
+  init() {
+    this.datepickerInput = this.wcNode.querySelector('.js-datepicker__input');
+    this.datepickerCalender = this.wcNode.querySelector('.js-datepicker__calender');
+    this.datepicker = this.wcNode.querySelector('.o-datepicker');
```
LucaMele: needs `js-datepciker class for this`

src/components/o-datepicker/js/datepicker.js
```diff
@@ -0,0 +1,135 @@
...
+    if (getAttribute(this.wcNode, 'open') === true) {
+      this.wcNode.removeAttribute('open');
+    } else {
+      this.wcNode.setAttribute('open', true);
+    }
+  }
+
+  isItemInLowerHalf(target) {
+    const { top, height } = target.getBoundingClientRect();
+    const toBottom = window.innerHeight - (top + (height / 2));
```
LucaMele: `(top + (height / 2))` operation is duplicated

src/components/o-datepicker/js/datepicker.js
```diff
@@ -0,0 +1,135 @@
...
+    const toTop = top + (height / 2);
+
+    if (toBottom >= toTop) {
```
LucaMele: can be simplified: `return toBottom < toTop;`

src/js/date.js
```diff
@@ -95,3 +96,58 @@ export const getNumericWeekday = (locale = 'en-uk', date = new Date()) => {
...
+  const objDate = date.toLocaleString(locale, { day: 'numeric', month: 'numeric', year: 'numeric' });
+  return objDate;
+};
+
+export const isDateValid = (locale = 'en-uk', inputValue = '') => {
+  let splittedInput = [];
+  let orderedInput = [];
+  let date = null;
+
+  if (locale === 'de-ch') {
```
LucaMele: please check with indexOf if in the `inputValue` is more than one `.`

src/components/o-datepicker/js/datepicker.js
NoahLiechti: done

src/components/o-datepicker/js/datepicker.js
NoahLiechti: done

src/components/o-datepicker/js/datepicker.js
NoahLiechti: done

src/js/date.js
AndyOGo: I'm not sure whats the goal here?
Though I'm curious why I can't see usage native `Intl` APIs like [`Intl.DateTimeFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat)

AndyOGo: Btw. don't think of using [`Date.parse`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse)

AndyOGo: Any valid Date will always be representable as a number, if it's invalid it's value will be `NaN`.
https://stackoverflow.com/questions/1353684/detecting-an-invalid-date-date-instance-in-javascript

src/js/date.js
```diff
@@ -95,3 +96,58 @@ export const getNumericWeekday = (locale = 'en-uk', date = new Date()) => {
...
+  filters.forEach((filter) => {
+    options[filter] = '';
+  });
+
+  const combinations = [];
+  selectors.forEach((el1) => {
+    selectors.forEach((el2) => {
+      selectors.forEach((el3) => {
+        options = ({
+          year: el1,
```
AndyOGo: This seems to be overkill and those variable names aren't semantic - what is `el1`, `el2`...

src/components/a-input/index.js
NoahLiechti: done

src/js/date.js
AndyOGo: @NoahLiechti 
Cause the datepicker needs to be able to deal with various **internationalized**/**localized** formats.
And the [`Date.parse` API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse) does not support this as far as I know.
Further more it's spec is far from being deterministic:
> It is not recommended to use Date.parse as until ES5, parsing of strings was entirely implementation dependent. There are still many differences in how different hosts parse date strings, therefore date strings should be manually parsed (a library can help if many different formats are to be accommodated).

`dataString` Parameter:
> A string representing an [RFC2822](https://tools.ietf.org/html/rfc2822#section-3.3) or (a variant of) ISO 8601 date (**other formats may be used, but results may be unexpected**).

For more reasons see above links

LucaMele: Thank you @AndyOGo, we wont use date parse and find a solution that always works.

src/js/date.js
```diff
@@ -104,51 +107,37 @@ export const getLocaleDayMonthYear = (locale = 'en-uk', date = new Date()) => {
...
-
-  if (locale === 'de-ch') {
-    splittedInput = inputValue.split('.');
-    orderedInput = [splittedInput[1], splittedInput[0], splittedInput[2]];
-    date = new Date(orderedInput.join('.'));
-  } else {
-    date = new Date(inputValue);
-  }
+  // year, monthIndex, day
+  const blueprint = new Date(2017, 2, 23);
```
NoahLiechti: why not `const blueprint = new Date();`?

and then on row 133 this:

`const monthIndex = splittedBlueprint.indexOf(blueprint.getMonth());
  const dayIndex = splittedBlueprint.indexOf(blueprint.getDate());
  const yearIndex = splittedBlueprint.indexOf(blueprint.getFullYear());`

src/js/date.js
```diff
@@ -104,51 +107,37 @@ export const getLocaleDayMonthYear = (locale = 'en-uk', date = new Date()) => {
...
 
-  let objDate = '';
-  const numeric = 'numeric';
-  const twoDigit = '2-digit';
-  const selectors = [numeric, twoDigit];
-  const filters = ['year', 'month', 'day'];
-  let options = {};
-  let valid = false;
+  // find out which out of 4 valid seperator the current locale has
+  const localisedBlueprintDate = new Intl.DateTimeFormat(locale).format(blueprint);
```
AndyOGo: We can query supported locales with:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat/supportedLocalesOf

src/js/date.js
```diff
@@ -1,6 +1,9 @@
 export const DEFAULT_NEW_YEARS = 10;
 export const TODAY = 'TODAY';
 
+// https://en.wikipedia.org/wiki/Date_format_by_country
+const ALL_DATE_SEPERATORS = [' ', '.', '/', '-'];
```
AndyOGo: Maybe better to use:
https://github.com/Brightspace/intl

> Why not just use the standard [ECMAScript Internationalization API (ECMA-402)](http://www.ecma-international.org/ecma-402/1.0/#sec-8) and [related polyfills](https://github.com/andyearnshaw/Intl.js)? Firstly, **the standard doesn't include any parsing functionality**. Additionally, Brightspace supports fine-grained locale customization at the organization and user levels -- a level of configuration that simply isn't present in the standard. This library does attempt to follow the standard API syntax and naming conventions when possible.

AndyOGo: IMHO there are dozens of various locale files and we should rely on proper parsing based on concrete rules.

LucaMele: not necceary to add another dependecy when those 4 seperators are internationally defined

LucaMele: but thanks

LucaMele: https://en.wikipedia.org/wiki/Date_format_by_country

src/js/date.js
LucaMele: added local check that throws an expetion. thanks

src/js/date.js
AndyOGo: Please be carefull and watch out.
Those separates do not tell you how date information is order - i.e big-endian, little-endian, middle-endian.
And for proper parsing we need this information of `YYYY-MM-DD` vs `MM-DD-YYYY` vs `DD-YYYY-MM` and all other combination used in different locales.

LucaMele: that part is covered below in the logic

LucaMele: that is what the blueprint is for

LucaMele: here we ONLY want to know on what to split (`-`, `.`, etc). Below we get the position for a certain locale using the blueprint

LucaMele: it also includes YY or YYYY and MM or M.

AndyOGo: Yeah I got that.
It's clever though this way we have to trust a browsers implementation which maybe buggy.
And I know to do it ourselves is way to much work and very detailed...

LucaMele: agree in not re-inventing the wheel, but am sorry if works in all our browsers we need to support with YYYY-DD-MM and YYYY-MM-DD and MM-DD-YY and so on, why do we have to use a library? I mean why investing time and add a dependecy to another npm module if there is not evidence yet that native browsers have issue in implementing a standard api? I get your concern, but i think here is maybe a bit of overthinking. or am i wrong?

LucaMele: every npm dependecy decreases maintability and compatibility to other frontend stacks

src/js/date.js
```diff
@@ -104,51 +107,37 @@ export const getLocaleDayMonthYear = (locale = 'en-uk', date = new Date()) => {
...
-        });
-        combinations.push(options);
-      });
-    });
-  });
-  combinations.forEach((option) => {
-    objDate = date.toLocaleString(locale, option);
-    if (objDate === inputValue && date.getFullYear().toString().length === 4) {
-      valid = true;
+  for (let i = 0; i < ALL_DATE_SEPERATORS.length; i++) {
```
AndyOGo: I would suggest to use a regex here for matching by `String.match`, like:

```js
const reDateSeperators = / |,|.|-/;
const usedSeperator  = localisedBlueprintDateString.match(reDateSeperators)
```

src/js/date.js
```diff
@@ -104,51 +107,37 @@ export const getLocaleDayMonthYear = (locale = 'en-uk', date = new Date()) => {
...
-    return date;
+  }
+
+  // find out how the locale date is structured (YYYY-MM-DD, YYYY-DD-MM, etc) using the blueprint
+  const splittedValue = inputValue.split(usedSeperator);
+  const splittedBlueprint = localisedBlueprintDateString.split(usedSeperator);
+
+  // we know month is 3 cause we set 2 in the date creation. In the creation it take 2 as monthIndex and
+  // in reading gives the actual month (index + 1)
+  const monthIndex = splittedBlueprint.indexOf(3);
```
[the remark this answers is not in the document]
AndyOGo: Does this work with numbers vs strings?
And how is `'03'` dealt with?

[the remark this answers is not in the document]
AndyOGo: this should compare to strings, here are my local browser console tests:

screen shot 2018-09-05 at 12 32 41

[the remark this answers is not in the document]
LucaMele: this commit is old. it have to be string. plase see new commits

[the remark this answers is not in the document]
LucaMele: nice catch thought :)

src/js/date.js
```diff
@@ -95,3 +100,41 @@ export const getNumericWeekday = (locale = 'en-uk', date = new Date()) => {
...
+
+  if (!Intl.DateTimeFormat.supportedLocalesOf(locale).length) {
+    throw new Error(`locale not supported: ${locale}`);
+  }
+
+  // find out which out of 4 valid seperator the current locale has
+  const localisedBlueprintDate = new Intl.DateTimeFormat(locale).format(blueprint);
+  const localisedBlueprintDateString = localisedBlueprintDate.toString();
+
+  const usedSeperator = localisedBlueprintDateString.match(ALL_DATE_SEPERATORS);
```
AndyOGo: @LucaMele 
Be carefull `.match` returns an array of matches or `null` if nothing matched.
So you should pick `usedSeperator [0]` and check for `null`

AndyOGo: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match

src/js/date.js
```diff
@@ -95,3 +100,41 @@ export const getNumericWeekday = (locale = 'en-uk', date = new Date()) => {
...
+
+  // find out how the locale date is structured (YYYY-MM-DD, YYYY-DD-MM, etc) using the blueprint
+  const splittedValue = inputValue.split(usedSeperator);
+  const splittedBlueprint = localisedBlueprintDateString.split(usedSeperator);
+
+  // we know month is 3 cause we set 2 in the date creation. In the creation it take 2 as monthIndex and
+  // in reading gives the actual month (index + 1)
+  const monthIndex = splittedBlueprint.indexOf('3');
```
AndyOGo: Looks good to me.
One last question, since we are actively parsing the date here, I would suggest to put it into it's own `parseLocalisedDate` function (however you name it)

src/js/date.js
LucaMele: @AndyOGo nice catch! thanks!!

src/js/device-state.js
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  init() {
+    if (!this.node) {
+      this.node = document.body;
+      add(this.node, deviceStateClass);
+    }
+
+    const _handleResize = throttle(this.handleResize, 100);
+    this.offResize = on(ownerWindow(this.node), 'resize', _handleResize);
```
AndyOGo: result of call to `ownerWindow` could be cashed

src/components/a-device-state/_preview.html
```diff
@@ -1 +1 @@
-<div class="a-device-state"></div>
+<div class="a-device-state js-device-state"></div>
```
AndyOGo: Remember we talked about moving it to `utility`, what about doing it right now?

src/js/date.js
LucaMele: the method returns a parse date already. maybe i should change the name

src/components/a-device-state/_preview.html
LucaMele: is not a utility yet as css is loaded as a component. We need to engeener it better later one. im in such a stress at the moment that i would prefer to keep this for later on

src/js/device-state.js
LucaMele: thanks. i leave it for the moment (time issue)

src/js/device-state.js
```diff
@@ -0,0 +1,94 @@
...
+    this.offOrientationchange = on(ownerWindow(this.node), 'orientationchange', _handleResize);
+  }
+
+  listen(callbackOnChange) {
+    this.node = document.querySelector(deviceStateClass);
+    this.callbackOnChange = callbackOnChange;
```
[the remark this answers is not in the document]
LucaMele: correct.. more callbacks needs more instances. I wanted to be sure context is separate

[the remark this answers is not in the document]
AndyOGo: how does it handle first notification?
As I can see it only notifies on change, but what if the consumer needs the data at instanciation?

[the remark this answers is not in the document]
LucaMele: 1 observer for 1 listener for 1 callback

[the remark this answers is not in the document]
LucaMele: first notifcation: u call triggerOnce method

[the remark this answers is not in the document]
LucaMele: @AndyOGo line 39. Usage exmaple in src/components/m-datepicker/js/datepicker.js

src/components/m-footer-links/js/footer-links.js
```diff
@@ -31,18 +33,22 @@ export default class FooterLinks {
...
       if (hasDropdown && !this.dropDown) {
-        this.dropDown = new DropDown(this.wcNode);
+        this.dropDown = new DropDown(wcNode);
       } else if (!hasDropdown && this.dropDown) {
         this.dropDown.destroy();
         delete this.dropDown;
       }
     });
 
+    this.deviceStateObserver.triggerOnce();
```
AndyOGo: shouldn't that come directly form the `deviceStateObserver `?

LucaMele: what if you dont want the initial callback and only onchange?

src/js/device-state.js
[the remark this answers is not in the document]
LucaMele: why

src/js/ui-events.js
```diff
@@ -3,11 +3,12 @@ import on from './on';
 import getAttribute from './get-attribute';
 import outer from './outer';
 
-export const EVENTS = Enum('click', 'keyup', 'enter', 'move', 'leave', 'Escape', 'Esc', 'touchstart', 'touchmove', 'touchend');
+export const EVENTS = Enum('click', 'change', 'keyup', 'enter', 'move', 'leave', 'Escape', 'Esc', 'touchstart', 'touchmove', 'touchend', 'input');
```
AndyOGo: Actually I never intended `ui-events` to be an accumulation of all possible events.
Since it's useful though I may would create a separate file 🤔

LucaMele: :) make sense. I cant do it now thought for time reasons. Lets make an issue with all the things left

src/js/device-state.js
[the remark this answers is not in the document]
AndyOGo: May I ask which context you mean?
> correct.. more callbacks needs more instances. I wanted to be sure context is separate
context of `this`, or a contextual custom element, or some thing else?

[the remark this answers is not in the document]
LucaMele: not `this` but the wcNode. So that there is no shared memory.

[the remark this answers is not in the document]
LucaMele: relative to wcNode is always a new context. Ergo 4 items = 4 instances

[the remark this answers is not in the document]
AndyOGo: > why
Because I understand that, device state should:
- provide it's state upon consumption automatically (without the need to call `triggerOnce` from outside) - Since the whole concept of an observer is to get notified about subscription und not insist on being notified right now
- by attaching multiple `resize`/`orientationchange` events, those will be computed for every instance. By this it creates performance issues and memory issues. Overall this state should just be compute once and notify x subscribers
- Since `pub/sub` is generic pattern, I would favour one implementation so that devs hav to care only for one. If we have multiple implementation each dev has to first ask himself which oberserver do I need to import to observe x, y, z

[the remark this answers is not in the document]
AndyOGo: Alright, may I ask to see these?

> not `this` but the wcNode. So that there is no shared memory.
> relative to wcNode is always a new context. Ergo 4 custom elements = 4 instances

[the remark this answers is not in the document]
LucaMele: @AndyOGo we either merge this PR or we forget about the datepicker. I dont have time to run behind nice to have or minimal milliseconds improvemts. If in future is needed, we can change it. But building the golden solution form beginning does not match CleanCode best practices. triggerOnce is there so that someone is not forced to render if he only wants to do it onchange. Is not a pub/sub pattern atm. Due to not common sharing of memory is impossible to implement what you suggest without hacking around window object.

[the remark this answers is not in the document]
LucaMele: multiple listeners versus one that always call all subscriber and also cleanup logic of subscriber is irrelevant performance increase but relevant complexity increase

src/js/ui-events.js
AndyOGo: fine with me

src/js/device-state.js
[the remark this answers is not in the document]
LucaMele: No if we revert it wont work for datepicker and it trigger double on change due to the global shared listsner.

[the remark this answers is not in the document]
LucaMele: agree DOM OPS are heavy.. so better render once!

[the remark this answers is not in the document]
LucaMele: and how is this colerated to render once or not?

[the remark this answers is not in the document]
LucaMele: it does not trigger more renderings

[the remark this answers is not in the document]
LucaMele: it does render less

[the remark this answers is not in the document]
AndyOGo: hmh, maybe I miss some information here 🤔 
So the current implementation of `device-state` has bugs, I would be very interested to know them, because `<axa-footer-links>`, the switch between `<axa-header-navigation>` and `<axa-header-navigation-mobile>` and others rely on it.
And I thought it works properly and is well tested.

I would really appreciate to see with life code, what is broken or does not work as expected.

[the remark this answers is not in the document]
LucaMele: i dont have time andy am so sorry. But the pub/sub is ONLY used in axa-footer-links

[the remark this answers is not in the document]
AndyOGo: True, `device-state` is only used for footer-links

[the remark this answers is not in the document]
AndyOGo: Though I would really like to see the bug...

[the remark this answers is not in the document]
LucaMele: So, ISSUE 1

bildschirmfoto 2018-09-05 um 17 24 31

ISSUE 2
bildschirmfoto 2018-09-05 um 17 25 30

bildschirmfoto 2018-09-05 um 17 25 55

ISSUE 3
When open and closing the links, the counters increases.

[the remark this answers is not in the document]
AndyOGo: Allright, so lets finish this.
Refactoring it is fine for me, though I would ask to refactor it into a singleton (maybe as follow-up). And I still would like to see what went wrong with the previous one...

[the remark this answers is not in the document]
AndyOGo: > So, ISSUE 1
True, this is the problem with our build not supporting code-splitting right now.
If I remember correctly we wanted to have one common chunks build for all components (so its a build issue not a `pub/sub` issue)

> ISSUE 2
Agreed that `4` count is weird 👍 

> ISSUE 3
I'm not sure which counters you mean, if those of `console.log('HERE')`. Agreed again, that sucks.

PR #564, opened 2018-09-07
target attribute is implemented

NoahLiechti:
Fixes # (issue).

 - User can set the `target` attribute to `_blank`. The link will open in a new tab.

## Type of change


jan-capiak:
I was working on this PR as well, but you were faster, no problem :)
Maybe i would change:
- put it in m-link as well, because it is a tag too
- dont make default value target='', because you get in a tag target='undefined', but instead put there target = '_self', which is default value when you dont write target attribute at all

NoahLiechti:
@jan-capiak thanks for your feedback. I have implemented it.

PR #572, opened 2018-09-21
Bugfix/button disabled style

### LINKED ISSUE #543: Button has no disabled style

 - [x] Bug
 - [ ] Feature

## Expected Behavior

Buttons should have a `[disabled]` style and correct disabled behavior.

## Actual Behavior

Neither designed nor implemented.

NoahLiechti:
Fixes #543 .

 - Every button, doesn't matter what a constellation, can now be disabled 

## Type of change

 # How Has This Been Tested?
 
 Tested manually in the patterns library with different arguments like  `ghost motion icon arrow`
 # Checklist:
 

src/components/m-button/mixins/_button.scss
```diff
@@ -290,8 +299,24 @@
       background: $color-button-bg--motion--hover;
     }
   }
+
+  &:disabled {
```
LucaMele: Ok i said lets increase specificty instead of using important.. but i guess as we already used a lot of importants we can also here.

src/components/m-button/_template.js
```diff
@@ -33,14 +37,19 @@ export default function ({
     genericIcon = raw(`<axa-icon icon="${icon}" classes="m-button__icon"></axa-icon>`);
   }
 
-  if (tag.toLowerCase() === 'a') {
+  if (tag.toLowerCase() === 'a' && disabled) {
+    return html`<a href="${href}" target="${target}" class="${buttonClasses}" ${ARIA_DISABLED} tabindex="-1">
```
nakovskijosif: Can you use inline check for the disabled param like you did it for the button below?
We could avoid duplicate condition like this.
${disabled ? `${ARIA_DISABLED} tabindex="-1"` : ''}

NoahLiechti: I tried this before, but it didn't work so I switched to this method.

It gives me the error in the console:
`index.js:2780 DOMException: Failed to execute 'setAttribute' on 'Element': 'aria-disabled tabindex="-1"' is not a valid attribute name.
    at nanohtmlSetAttribute`

PR #578, opened 2018-10-01
add all components in one native exports

LucaMele:
add index js export for all components


src/index.js
```diff
@@ -0,0 +1,152 @@
+/* eslint-disable */
+import _AXACheckbox from './components/a-checkbox';
```
AndyOGo: You can export it directly by:

```js
export { default as AXACheckbox } from './components/a-checkbox';
```

https://stackoverflow.com/questions/35665759/es6-how-can-you-export-an-imported-module-in-a-single-line
https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export#Module_Redirects

LucaMele: nice thanks

LucaMele: will do asap

src/js/react-exports.js
```diff
@@ -0,0 +1,157 @@
+// First import the withReact integration helper function
```
AndyOGo: don't forget to rebase onto develop

LucaMele:
@AndyOGo PR FB implemented :)

PR #606, opened 2018-10-12
fix event on axa change

### LINKED ISSUE #244: Date-Picker component
For choosing Dates we need a AXA date picker component

LucaMele:
Fixes #244 .


LucaMele:
Merge without PR approval

Issue #634, opened 2018-10-24
Datepicker selected value seems to be a timestamp


AndyOGo:
## Are you reporting a bug or a feature?

 - [x] Bug
 - [ ] Feature

## Expected Behavior

It should display a localised Date format

## Actual Behavior

Display a time stamp

## Steps to Reproduce the Problem

![date-picker-timestamp-values](https://user-images.githubusercontent.com/914443/47420953-df2eba00-d77f-11e8-963f-a0c0e108cdd9.gif)

MarekLacoAXA:
Fixed by #635

PR #628, opened 2018-10-22
Bugfix/some dropdown issues

LucaMele:
Fixes #623 #624 #625 #626.

PR #635, opened 2018-10-24
fix(Datepicker): fix "output-iso" being always on problem

MarekLacoAXA:
**See my comment from 13:49 first.**

After the merge of #628 from axa-ch/bugfix/some-dropdown-issues in
version 2.0.1-beta.202, the Datepicker was always outputing number
of seconds (as with the output-iso option) instead of normal date.

This commit removes "output-iso" attribute from `<axa-m-datepicker>`
element when it has falsy value. Bacause otherwise this did lead into
"false" string value which was evaluated as true, causing this problem.

Also these other attributes are now left out when falsy:
higher-end-year, lower-end-year, selected-day

image

## Type of change

 # How Has This Been Tested?
 
 
 
 # Checklist:
 
### DIFF
--- a/src/components/m-datepicker/index.js
+++ b/src/components/m-datepicker/index.js
@@ -31,7 +31,7 @@ class AXADatepicker extends BaseComponentGlobal {
     selectedDay: PropTypes.number,
     lowerEndYear: PropTypes.number,
     higherEndYear: PropTypes.number,
-    outputIso: PropTypes.string,
+    outputIso: PropTypes.bool,
   }
 
   // Specify observed attributes so that attributeChangedCallback will work,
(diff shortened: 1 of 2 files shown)


MarekLacoAXA:
Let me know whether to merge all commits (the including the 5e61ccf), or the 9fde7e3 change is enough.

(I am not sure whether using raw() instead of html`` is okay in 5e61ccf)

LucaMele:
@MarekLacoAXA very well seen!! Thank you soo much for this!

MarekLacoAXA:
Fixes #634

MarekLacoAXA:
Close #634 
(just trying...)

PR #690, opened 2018-11-26
Native dropdown numbers as values selection fixed.

### LINKED ISSUE #688: Dropdown: native selection not working with numbers as values

 - [x] Bug

## Expected Behavior

When I select item from a native dropdown that has number as value it is not selected correctly.

 - Add the items attribute in which the values are set as numbers.
`items='[{"name": "Item 1 empty value", "url": "#", "value":3}, {"name": "Item 2", "url": "#", "value":4}, {"name": "Item 3", "url": "#", "value":5}]'`

 - Select item from the list.
The value will be set but the dropdown will not show the selection.

nov-26-2018 15-51-09

nakovskijosif:
Fixes #688 .

fixed the comparison between different prop types.

## Type of change

 # Checklist:
 
### DIFF
--- a/src/components/m-dropdown/_template.js
+++ b/src/components/m-dropdown/_template.js
@@ -17,7 +17,7 @@ const nativeSelect = ({ title, items, size, value }) => html`<div class="${class
       items.map(({ name, value: itemValue, url }, index) => {
         const _itemValue = getItemValue(itemValue, index);
         return html`<option value="${_itemValue}" data-url="${url}" ${
-          _itemValue === value ? 'selected' : ''
+          `${_itemValue}` === `${value}` ? 'selected' : ''
         }>${name}</option>`;
       })}
     </select>
--- a/src/components/m-dropdown/_template.js
+++ b/src/components/m-dropdown/_template.js
@@ -17,7 +17,7 @@ const nativeSelect = ({ title, items, size, value }) => html`<div class="${class
       items.map(({ name, value: itemValue, url }, index) => {
         const _itemValue = getItemValue(itemValue, index);
         return html`<option value="${_itemValue}" data-url="${url}" ${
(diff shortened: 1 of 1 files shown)


src/components/m-dropdown/_template.js
```diff
@@ -17,7 +17,7 @@ const nativeSelect = ({ title, items, size, value }) => html`<div class="${class
       items.map(({ name, value: itemValue, url }, index) => {
         const _itemValue = getItemValue(itemValue, index);
         return html`<option value="${_itemValue}" data-url="${url}" ${
-          _itemValue === value ? 'selected' : ''
+          `${_itemValue}` === `${value}` ? 'selected' : ''
```
AndyOGo: Guess this check should not be of strict type?
If so I would just write `_itemValue  == value`, the javascript engine will handle type conversion for you, so that stuff like `'4' == 4` will be truthy

nakovskijosif: I can do that, will need to add  // eslint-disable-line though.

LucaMele: NO PLEASE nogo for `_itemValue == value` from my side. Better type conversion to string so that we are type save. @nakovskijosif

nakovskijosif: I will revert my last commit in a sec.

AndyOGo: I agree with you Luca, not many people read specs and definitely saver to stick with `===` than `==`. My point here is that this is perfect valid example of a `==` use case.

But to make this point more prevalent, we could also use `toString` conversion helpers, instead of meaning-lesl template lits

LucaMele: ok with the toString. For me is the same as both clearly show what is happening.

Issue #720, opened 2018-12-03
build huge script waterfall


AndyOGo:
## Are you reporting a bug or a feature?

 - [ ] Bug
 - [x] Feature

## Expected Behavior

All `<scipts>` should be loaded asynchronously **as fast as possible** in **all browsers** - ideally by a sophisticated bundler/loader.

- [ ] Load just one bundle
- [ ] Or ideally lazy-load in-view components by `import().then()`, like react-loadable

## Actual Behavior

All `<scipts>` are loaded synchronously and therefore block the next script.

## Steps to Reproduce the Problem

1. inspect network traffic.

![build-script-huge-waterfall](https://user-images.githubusercontent.com/914443/49392353-764b4200-f72e-11e8-8e8c-8ea8aac69271.gif)

## Specifications

 - Browser:
 - Browser Version:
 - Node Version:
 - NPM Version:
 - Platform:

LucaMele:
Closing as V2 is released. If issue is still relevant in v2, please re-open

PR #724, opened 2018-12-04
fix(table): added used custom elements as dependencies

### LINKED ISSUE #723: Dependencies between components not expressed

 - [x] Bug
 - [ ] Feature

## Expected Behavior

All components should clearly state on which other components they depend upon.

This is fundamental for **Single Component Views**.

## Actual Behavior

We have no dependency tree between components.

AndyOGo:
**Blocked** by not using a module bundler, chunk optimiser.

Fixes #723 and paves the way for #720 

 - Each component `import`s all custom elements used within it's template
   **Note:** all components passed as children, still have to be loaded manually, e.g. contents of `preview.html`
 - This paves the way for our new **Single Component View** to reduce script load water fall #720 and just load one script instead of all of them

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
 
 # How Has This Been Tested?
 
 
 
 # Checklist:
 
### DIFF
--- a/src/components/m-accordion-item/_template.js
+++ b/src/components/m-accordion-item/_template.js
@@ -1,6 +1,8 @@
 import html from 'nanohtml';
 import classnames from 'classnames';
 
+import '../a-icon';
+
 export default ({ header, headerSecondary, headerColor, icon = '' }, childrenFragment) => {
   const headerPrimaryClasses = classnames('m-accordion-item__heading', 'm-accordion-item__heading--primary', {
     [`m-accordion-item__heading--${headerColor}`]: headerColor,
--- a/src/components/m-button/_template.js
+++ b/src/components/m-button/_template.js
@@ -3,6 +3,8 @@ import raw from 'nanohtml/raw';
 import classnames from 'classnames';
 import isEmptyFragment from '../../js/is-empty-fragment';
 
+import '../a-icon';
+
 const DISABLED = 'disabled';
 const ARIA_DISABLED = 'aria-disabled';
 
--- a/package.json
+++ b/package.json
@@ -85,6 +85,7 @@
     "autoprefixer": "^7.2.3",
     "babel-cli": "^6.26.0",
     "babel-eslint": "^8.2.1",
+    "babel-plugin-discard-module-references": "^1.1.2",
     "babel-plugin-external-helpers": "^6.22.0",
     "babel-plugin-inline-replace-variables": "^1.3.1",
     "babel-plugin-transform-class-properties": "^6.24.1",
--- a/src/components/m-datepicker/_template.js
+++ b/src/components/m-datepicker/_template.js
@@ -7,6 +7,9 @@ import {
   TODAY,
 } from '../../js/date';
 
+import '../m-dropdown';
+import '../m-datepicker-body';
+
 /**
  * @param yearsRange object with keys: lowerEndYear, higherEndYear
  * @param startYear number|string year number, or 'TODAY' string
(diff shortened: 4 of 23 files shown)


src/components/m-accordion-item/_template.js
```diff
@@ -1,6 +1,8 @@
 import html from 'nanohtml';
 import classnames from 'classnames';
 
+import '../a-icon';
```
LucaMele: Why `a-icon` here?

src/components/m-button/_template.js
```diff
@@ -3,6 +3,8 @@ import raw from 'nanohtml/raw';
 import classnames from 'classnames';
 import isEmptyFragment from '../../js/is-empty-fragment';
 
+import '../a-icon';
```
LucaMele: dtto

LucaMele:
@AndyOGo i would discuss this here. I would instead of hard dependecy add simply a console error in case a dependecy has not loaded on the webside. Like this can brake productions as we would have tons of icons.

src/components/m-accordion-item/_template.js
AndyOGo: Because the template below outputs `<axa-icon>`

LucaMele:
So to create a generic system for dependency and if that system cant find the registred element it needs it will throw an error but i would not make hard dependecy (especially for the dist build is a nogo)

src/components/m-button/_template.js
AndyOGo: same as above, use in template below

AndyOGo:
@LucaMele 
Correct, I'm aware of this. That's why I stared my PR message by:

> **Blocked** by not using a module bundler, chunk optimiser.

Anyway, we need proper dependency resolution

LucaMele:
@AndyOGo yes but i wont agree on hard dependency, just soft dependcy with error messages in my opinion is better design. If you think otherwise we need to discuss this please and weight out pro and cons

LucaMele:
and bundler wont solve anything is we have per component one dist file (which should be). Bundler only solve the issue on the LIB side or if we have 1 dist for everything (which is a nogo and against the design of the patterns lib)

AndyOGo:
@LucaMele 
Indeed I think that webpack/rollup/systemJS should be able resolve dependencies and create a proper dependency tree, to make nice chunk optimisations

LucaMele:
@AndyOGo the chunk & dependecy only works if you can only include certain dependecy. also AJAX requests can not be a solution cause not all of our implementations maybe can download JS dynamically

LucaMele:
So i insist, we need an error system not a hard dependcy otherwise we lock ourself in a certain stack

LucaMele:
and we provide single disted components AND ES Modules

LucaMele:
This PR only benefits the ES Module export damaging the DISTED files. So please lets go wirh a "soft" dependecy mgt or we need to discuss it on monday

AndyOGo:
@LucaMele 
The one dist file per component would still work - just that any dependencies would be automatically included instead of manually importet 👍 
As I understand it, using the UMD dist files directly as `<scipts>` is not the recommended way, also because of a huge performance bottleneck #720.
**ESM** should be the recommended approach and we should do all to make it easy with ESM.

LucaMele:
UMD is not reccomended, IFEE is also not reccomended but sometimes necessary and therefore we should provide it. Immagine if all components have all theier depemdcy inside. We will end up with Disted files that are 100KB PER COMPONENT! NO go from my side and i wont comment futher. Either Soft Dependecy or we discuss it on Monday

AndyOGo:
Yes we can discuss it

AndyOGo:
Alright Luca.

We really have created an Anti-Patterns with these dist files, which results in a huge performance bottleneck, which i documented here #720.
Please let us discuss it, and let's stick to modern standards like **ESM**.
Again you right, `o-table` bundle will be bigger, but they don't have to import all other relating components, so there will be no increase in overall download. And loading single scripts is much faster, than multiple scripts.

PR #740, opened 2018-12-05
breakes built-in ie but fixes with-react

LucaMele:
Fixes https://github.com/axa-ch/patterns-library-examples/issues/12.

@AndyOGo this PR fixes `with-react` in IE for AEM, Patterns-Library and Patterns-Library-Examples.
@AndyOGo it brakes built-in ONLY on IE (Safari, EDGE, FF and Chrome worked with built-ins).

### DIFF
--- a/package.json
+++ b/package.json
@@ -154,12 +154,12 @@
   },
   "dependencies": {
     "@ungap/custom-elements-builtin": "^0.1.1",
+    "@webcomponents/webcomponentsjs": "^2.2.1",
     "babel-plugin-transform-builtin-classes": "^0.6.1",
     "babel-runtime": "^6.26.0",
     "classnames": "^2.2.5",
     "connect-slow": "^0.4.0",
     "core-js": "~2.5.1",
-    "document-register-element": "^1.12.0",
     "innersvg-polyfill": "0.0.2",
     "nanohtml": "^1.3.0",
     "objectFitPolyfill": "^2.0.5",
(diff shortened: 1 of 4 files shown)


LucaMele:
@AndyOGo , we need this so that we can release AEM and the form refactored so i guess not working built in in IE is the minor issue and we have more time to find out why

AndyOGo:
@LucaMele 
Well that's so unfortunate, WebReflection's polyfill is the only battle tested and feature complete polyfill.
And it is really weird that our preview works across browsers, but the consumed patterns-library doesn't 😕 

Reverting `built-in` support is a regression honestly.
And I think that something is going mad during build or transpilation time.

Please do what you prefer.
I can totally see, that this should be released at AEM right now.

LucaMele:
@AndyOGo is weird combo between with-react and the "battle" proofed polyfill. And actually our preview does not work either: https://axa-ch.github.io/patterns-library/react.html

AndyOGo:
Jep but that is since the build broke...

LucaMele:
I will sleep it over, but in worst case i have to merge. But u can test, built in works everywhere apart of IE so is fine

LucaMele:
@AndyOGo i think it has other issues as well. I will close it and try again tomorrow

AndyOGo:
@LucaMele 
Do tested `built-in` in safari? and it worked?
I think they should be broken in all non-custom-element browsers and Safari

LucaMele:
@AndyOGo yes they did with webcomponents and built in polyfill but in ie some console error came so i gave up

PR #750, opened 2018-12-06
Fix some datepicker issues

LucaMele:
Fixes some datepicker issues

### DIFF
--- a/src/components/o-datepicker/js/datepicker.js
+++ b/src/components/o-datepicker/js/datepicker.js
@@ -126,6 +126,9 @@ export default class Datepicker {
   }
 
   isItemInLowerHalf(target) {
+    if (!target) {
+      return true;
+    }
     const { top, height } = target.getBoundingClientRect();
     const calculation = top + (height / 2);
     const toBottom = window.innerHeight - calculation;
--- a/src/js/prop-types/locale-prop-type.js
+++ b/src/js/prop-types/locale-prop-type.js
@@ -1,7 +1,6 @@
 import isRequiredHelper from './is-required-helper';
 
-// link https://stackoverflow.com/questions/8758340/regex-to-detect-locales
-const reLocale = /^[a-z]{2,4}(-([A-Z][a-z]{3}|[0-9]{3}))?(-([A-Z]{2}|[0-9]{3}))?$/;
+const reLocale = /^([\w]{2})-([\w]{2})/;
 
 // eslint-disable-next-line consistent-return
 function localePropType(props, propName, componentName) {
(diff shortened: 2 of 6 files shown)


src/js/prop-types/locale-prop-type.js
```diff
@@ -1,7 +1,6 @@
 import isRequiredHelper from './is-required-helper';
 
-// link https://stackoverflow.com/questions/8758340/regex-to-detect-locales
-const reLocale = /^[a-z]{2,4}(-([A-Z][a-z]{3}|[0-9]{3}))?(-([A-Z]{2}|[0-9]{3}))?$/;
+const reLocale = /^([\w]{2})-([\w]{2})/;
```
NoahLiechti: does this validate if the last two letters are uppercase?

LucaMele: good check! I just tested it and seems like it works:

bildschirmfoto 2018-12-06 um 12 03 46

AndyOGo: hmh, it does also accept lower case country codes `de-ch`.
https://regex101.com/r/UfQQOu/1

screen shot 2018-12-06 at 12 10 30

LucaMele: @AndyOGo yepp. Otherwise it crashes. The intl api accepts both so i guess is ok

LucaMele: @AndyOGo mmm im actually worried now. On chrome it accepts `de-de` and `de-DE`. I was overviewving the spec https://www.ecma-international.org/ecma-402/1.0/#sec-6.2 and couldnt get clever about what is allowed and what not

LucaMele: Yepp ok so the regex is good as is as the specs contradict iteslf and at the end if intl api can work with it, lets set it as intl api want it

AndyOGo: hehe, yeah. Many people/organisations have tried to solve i18n problems and spec them, but these are very complex and changing over time

LucaMele: And a very well distributed and accepted engeeiring principle says:

Flexible im entegennehmen und präxis im weitergeben

src/components/o-datepicker/js/datepicker.js
```diff
@@ -126,6 +126,9 @@ export default class Datepicker {
   }
 
   isItemInLowerHalf(target) {
+    if (!target) {
```
NoahLiechti: in the constructor there's a little mistake. change `this._localValue` to `this.localeValue`.

PR #756, opened 2018-12-13
Add TypeScript types

MarekLacoAXA:

 - add several *.d.ts files containing typescript type definitions
 - add "build-types" run script which copies these files from src/ to lib/
 - change "build-lib" run script so that it also runs the above "build-types"

## Type of change

 # How Has This Been Tested?
 
 
 
 # Checklist:
 

LucaMele:
@MarekLacoAXA thank you for this. It looks good so far. I will test it today and then we can merge it

MarekLacoAXA:
Ok, the AMO script is updated to produce index.d.ts and update react-exports.d.ts

LucaMele:
@MarekLacoAXA we have now React dependecy to the Types. What if we want to use the Webcomponent without react and with typescript. Im genererally not fan of this, but as is only typings, i think is ok.

@jackblackCH please share your opinion as well

jackblackCH:
👍 for typescript.

@MarekLacoAXA, is it necessary to export it as `React.ComponentClass<AXAIconProps>;`?
The purpose of this pattern library is indeed to be framework agnostic, what if I use typescript with Vue.JS or Angular or technology x?

MarekLacoAXA:
> @MarekLacoAXA we have now React dependecy to the Types. What if we want to use the Webcomponent without react and with typescript. Im genererally not fan of this, but as is only typings, i think is ok.

Hi @LucaMele & @jackblackCH !

With this pull request we would keep the TS declaration files alongside with each component (in same directory). The disadvantage is more crowded repository (more technologies together side by side). An advantage is that it is easier to keep the types **up to date**. Whenever you create/update new properties, the type declaration file is close (same directory), so each commit/release can contain an up-to-date version of types.

Another approach would be to maintain a separate repo/package for TS types. Name of it would be like **@types/patterns-library**

As for the react in PL: The PL already includes the with-react.js (which imports from 'react') and react-exports.js. So I think it fine to rely on @types/react also (if we keep the TS types in PL repo).

Another approach would be to split the react stuff to a separate repo/package and then maintain it. Name of it would be like **@axa-ch/patterns-library-react**.
And again, we could keep the TS types within the @axa-ch/patterns-library-react, or even split them further into a **@types/patterns-library-react**

To wrap this up:

- I think its a good to keep the TS types within PL repo, as long as these TS types would be really kept up to date
- But IF the included TS types should become often outdated (not maintained), then its easier to have them separate in **@types/patterns-library**, for more flexibility (no need to wait for official PL build when a quick change in types is needed)

MarekLacoAXA:
@LucaMele , ok, I see. So I propose this change (as in the last commit):

the individual index.d.ts would not use react any more, but export just:
```
export interface AXAIcon extends HTMLElement, AXAIcon.Props {}
```

and the react-exports.d.ts would then import react and export:
```
export const AXAIcon: React.ComponentClass<AXAIcon.Props>;
```

Is it okay this way? I would then rewrite all component types accordingly.

LucaMele:
@MarekLacoAXA yes thats perfect! Please dont forget also the AMO Script. Thank you very much

MarekLacoAXA:
Ok, done.

MarekLacoAXA:
Types declarations are now rewritten to allow better versatility. Ready to merge from my point of view.

PR #759, opened 2018-12-14
dropdown

### LINKED ISSUE #758: Dropdown native - selectedIndex is not updated when items are changed.

 - [x] Bug
 - [ ] Feature

## Expected Behavior

When I change "items" attribute via javascript the selectedIndex is not updated although the component is rendered well.

## Actual Behavior

The selectedIndex in not updated when the "items" attribute is changed after the components is rendered.

remove the "items" attribute on already rendered component via javascript and add it again. You will see that last item is shown as selected even though the value is not for the last item.

nakovskijosif:
Fixes #758.

on reset the selected index will be updated if the index changed or set to 0 if there is no selected option

## How to reproduce
Prepare a custom items array (where you change the order) and update it via setAttribute. (Pasting a serialized string into inspector does not work)

Make the element accessible via id first, i.e, add an id='dropdown' and then set the attribute items
```
document.getElementById('dropdown').setAttribute('items', '[{"name": "Item 1 empty value", "url": "#", "value":"a"}, {"name": "Item 2", "url": "#", "value": "b"}, {"name": "Item 3", "url": "#", "value":"c"}]');
```
The native element (when you click) should be updated accordingly to the items array

## Type of change

 # Checklist:
 
### DIFF
--- a/src/components/m-dropdown/js/drop-down.js
+++ b/src/components/m-dropdown/js/drop-down.js
@@ -11,7 +11,6 @@ class DropDown extends UiEvents {
     isOpenClass: 'is-dropdown-open',
     isAnimatingClass: 'is-dropdown-animating',
     selectClass: 'js-dropdown__content',
-    selectedItemSelector: 'option[selected=selected]',
   }
 
   constructor(wcNode, options) {
--- a/src/components/m-dropdown/js/drop-down.js
+++ b/src/components/m-dropdown/js/drop-down.js
@@ -11,6 +11,7 @@ class DropDown extends UiEvents {
     isOpenClass: 'is-dropdown-open',
     isAnimatingClass: 'is-dropdown-animating',
     selectClass: 'js-dropdown__content',
+    selectedItemSelector: 'option[selected=selected]',
   }
 
   constructor(wcNode, options) {
@@ -147,6 +148,17 @@ class DropDown extends UiEvents {
 
       lastElementChild.style.height = '';
       remove(node, this.options.isOpenClass);
+    }
+    const select = this.wcNode.querySelector(`.${this.options.nativeSelectClass}`);
+    if (select) {
+      const selectedOption = this.wcNode.querySelector(this.options.selectedItemSelector);
+      if (selectedOption) {
+        if (selectedOption.index !== select.selectedIndex) {
+          select.selectedIndex = selectedOption.index;
+        }
+      } else {
+        select.selectedIndex = 0;
+      }
     }
   }
 
--- a/src/components/m-dropdown/js/drop-down.js
+++ b/src/components/m-dropdown/js/drop-down.js
@@ -149,17 +149,6 @@ class DropDown extends UiEvents {
       lastElementChild.style.height = '';
       remove(node, this.options.isOpenClass);
     }
-    const select = this.wcNode.querySelector(`.${this.options.nativeSelectClass}`);
-    if (select) {
-      const selectedOption = this.wcNode.querySelector(this.options.selectedItemSelector);
-      if (selectedOption) {
-        if (selectedOption.index !== select.selectedIndex) {
-          select.selectedIndex = selectedOption.index;
-        }
-      } else {
-        select.selectedIndex = 0;
-      }
-    }
   }
 
   destroy() {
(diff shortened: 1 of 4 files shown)


nakovskijosif:
@AndyOGo I can confirm that the issue is fixed for the dropdown too, I tested it out with every possible scenario and it works fine. If your only concern is that this is not working properly I can assure you it is working just fine.

AndyOGo:
@nakovskijosif 
Thanks for your reply.

You are right, I know that this is working, I told you the cause of `selectedIndex` yesterday.
But what is important, this PR only deals with the datepicker.
The ideal place to fix this is for all and every components which may ever use `<select>`.
So please fix it in the DOM morphing/diffing part for all components, not just the datepicker.

src/components/m-dropdown/js/drop-down.js
```diff
@@ -148,6 +149,17 @@ class DropDown extends UiEvents {
...
     }
+    const select = this.wcNode.querySelector(`.${this.options.nativeSelectClass}`);
+    if (select) {
+      const selectedOption = this.wcNode.querySelector(this.options.selectedItemSelector);
+      if (selectedOption) {
+        if (selectedOption.index !== select.selectedIndex) {
+          select.selectedIndex = selectedOption.index;
+        }
+      } else {
+        select.selectedIndex = 0;
```
AndyOGo: `0` to unset a `<select>` is incorrect.

Please see the docs here:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/selectedIndex

nakovskijosif: thank you Andy, I'll update the PR and change the implementation, I see your point and agree.

nakovskijosif:
@AndyOGo 
Please review my update. I applied special handlers for "weird dom nodes" after the node and its children are morphed/copied, otherwise the special handlers were overwritten when the children were morphed/copied. 
the update option function was not being called when a child node is copied in updateChildren, this is why i did this.
Let me know what you think.

src/js/abstract/utils/special-el-handlers.js
```diff
@@ -0,0 +1,115 @@
...
+    updateSelect(newNode, oldNode);
+  }
+}
+
+function updateOption(newNode, oldNode) {
+  updateAttribute(newNode, oldNode, 'selected');
+}
+
+function updateSelect(newNode, oldNode) {
+  var selectedIndex = -1;
```
LucaMele: @nakovskijosif is there any reason to use `var` here?

LucaMele:
@nakovskijosif to me looks good apart of that `var`. And please note that jenkins is failing so please fix that as well:

LucaMele:
bildschirmfoto 2018-12-17 um 09 52 46

nakovskijosif:
> @nakovskijosif to me looks good apart of that `var`. And please note that jenkins is failing so please fix that as well:

@LucaMele I will fix the es-lint issues, I was just wondering what you think of the concept because it changes the flow of the morphing a bit. So far from what I tested seems to be working fine and I think this is the way we should do it.

src/components/m-dropdown/js/drop-down.js
```diff
@@ -11,6 +11,7 @@ class DropDown extends UiEvents {
     isOpenClass: 'is-dropdown-open',
     isAnimatingClass: 'is-dropdown-animating',
     selectClass: 'js-dropdown__content',
+    selectedItemSelector: 'option[selected=selected]',
```
LucaMele: will this work in all the browser? maybe better `option[selected="selected"]`

LucaMele:
@nakovskijosif am ok with it, but will take more time to test and understand it this afternoon

src/components/m-dropdown/js/drop-down.js
nakovskijosif: This will  be removed, it is part of the old implementation that I forgot to clean up and it is not used atm.

LucaMele:
@AndyOGo it looks like @nakovskijosif did implement your Change Requests and therefore i will take your flag away as you are on holiday. Thank you all for this!

jackblackCH:
As discussed with @LucaMele, I also tested with a custom prepared items array (see updated description). The dropdown and native dropdown changed accordingly to the new array. lgtm.

nakovskijosif:
Thank you all for your feedback and testing!

Issue #757, opened 2018-12-14
Datepicker is not setting value in IE.


nakovskijosif:
## Are you reporting a bug or a feature?

 - [x] Bug

## Expected Behavior

When I select a date in IE the "value" attribute on <axa-datepicker ../> should be updated accordingly.

## Actual Behavior

the value is always set to false.

jackblackCH:
Please always add the exact version of IE version and OS

LucaMele:
Closing as V2 is released. If issue is still relevant in v2, please re-open

PR #767, opened 2018-12-19
Feature new /datepicker/dropdown

### LINKED ISSUE #764: Datepicker sets a "wrong" iso date compared to the "clicked" one
When you chose a date and click ok, the iso-date that is set, is not matching with the chosen one, it's actually set to the day before.

## How to reproduce:
`const date = new Date(2020, 11, 31)` `// Thu Dec 31 2020 00:00:00 GMT+0100 (Central European Standard Time)`

`date.toISOString()` `// 2020-12-30T23:00:00.000Z`

It looks as if the output date of `date.toISOString()` is one day behind.
This is technically correct but probably not the expected behavior.

Update:
If we really work correctly with iso-dates, we should accept them and convert them back on the backend accordingly.

jackblackCH:
Fixes #764, #757

## Changes and features proposed in this pull request:
 - legacy friendly ( still needs to be tested a lot if you update)
 - Refactoring of the datepicker. cleanup of functions and dependencies
 - Fixes  "iso-date" and human readable date issue #764 
 - Fixes Duplicate of classnames
 - Allow closing the datepicker on body click
 - Fixes/Improvements of styling and spacings
    - Approved with designer
- Dropdown refactorings, makes dropdowns more flexible (needed for the datepicker, too)
- Allow a range for the year definition (i.e 1990-2018)

## TODOs for later.
- Accessibility. Keyboard usage.
   - close on ESC..
- Should be possible to use datepicker with keyboard only

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)


ranostaj:
Hey, have anyone any idea when this is going to be reviewed and resolved? It seems it has fixed hot issue with IE which I was referring from #757 
Thank you!

jackblackCH:
Please have review @LucaMele @AndyOGo

LucaMele:
@jackblackCH can you please fix the conflicts before?

src/app/_preview.scss
```diff
@@ -27,6 +27,7 @@ $sg-min-height: 50px;
 $sg-gutter-size: 15px;
 
 body {
+  height: 100vh;
```
AndyOGo: @LucaMele 
@jackblackCH 
Do we really want to use viewport units, if there is nor real benefit in comparision to `100%` in this case?
https://caniuse.com/#feat=viewport-units

src/components/a-icon/index.scss
```diff
@@ -1,11 +1,16 @@
 a-icon__root {}
 
-.a-icon {
+.a-icon .a-icon__svg {
```
AndyOGo: This CSS specificity seems to be to high

src/components/a-input/_template.js
```diff
@@ -20,9 +20,9 @@ export default function ({
           value="${value}"
           disabled="${disabled}">
           <span class="a-input__valid-icon"></span>
-          ${icon ? html`<button class="a-input__icon__button js-input__icon__button" type="button">
+          ${icon ? html`<button class="a-input__icon-button js-input__icon-button" type="button">
```
AndyOGo: @LucaMele 
@jackblackCH 
hmh, may off-topic.
But why do we have a `<button>` next to `<input />` ?

src/components/m-datepicker-body/index.js
```diff
@@ -17,58 +17,153 @@ class AXADatepickerBody extends BaseComponentGlobal {
...
+    this.year = date.getFullYear();
+    this.month = date.getMonth();
+    this.day = date.getDate();
+    this.index = index;
+
+    // Set the day to the chosen day
+    this.date = date.toISOString();
+  }
+
+  get index() {
```
AndyOGo: @LucaMele 
@jackblackCH 
The boilerplate for getters/setters repeat most of the time.
May we find a more generic approach to add them

src/components/m-datepicker-body/index.scss
```diff
@@ -21,25 +15,33 @@ $datepicker-calender-body-padding-top: 5px !default;
 }
 
 .m-datepicker-body__cell {
-  width: $with-one-seventh;
-  height: $with-one-seventh;
-
+  width: 33px;
```
AndyOGo: Please use `$variables`

src/components/m-datepicker-body/index.scss
```diff
@@ -21,25 +15,33 @@ $datepicker-calender-body-padding-top: 5px !default;
...
   background-color: transparent;
-
   border-radius: $datepicker-cell-border-radius;
   cursor: pointer;
   border: none;
 }
 
+.m-datepicker-body__cell:active {
+  outline: 0;
+  border: solid 2px $color-axa-blue;
```
AndyOGo: Cell border should also have `$variables`

src/components/m-datepicker/index.js
```diff
@@ -2,106 +2,177 @@ import PropTypes from '../../js/prop-types'; // eslint-disable-next-line import/
...
+
+    range(start, end) {
+      return Array.from({ length: (end - start) }, (v, k) => k + start);
+    }
+
+    connectedCallback() {
+      super.connectedCallback();
+
+      // This method is triggered up to a millions times.... but it's needed here else, no items are available
+      this.props.monthItems = getAllLocaleMonthsArray(this.props.locale).map((item, index) => ({
```
AndyOGo: @LucaMele 
@jackblackCH 
These values should not change often, right?
So how about memoization or inline caching?

src/components/m-datepicker/index.scss
```diff
@@ -49,29 +50,36 @@ $datepicker-padding-left-native-dropdown: 8px !default;
...
-    display: inline-block;
-  }
-
   .m-dropdown__select {
     @include dropdown-select($indent: 10px, $height: 40px);
   }
 
   .m-dropdown__select-icon {
     @include dropdown-select-icon($indent: 10px, $height: 40px);
+    width: 35px;
```
AndyOGo: @jackblackCH 
@MarekLacoAXA 
May it fits better to add `width` arg to [`dropdown-select-icon`](https://github.com/axa-ch/patterns-library/blob/dbaf0823e7d5de376d2943c135d0c8d80af1b96f/src/components/m-dropdown/mixins/dropdown.scss#L127) mixin

src/components/m-datepicker/js/datepicker.js
```diff
@@ -1,179 +1,81 @@
 import { EVENTS, AXA_EVENTS } from '../../../js/ui-events';
 import on from '../../../js/on';
 import fire from '../../../js/fire';
-import DeviceStateObserver from '../../../js/device-state';
```
AndyOGo: @jackblackCH 
@LucaMele 
Is responsive awareness somewhere else?
Or is it removed?

src/components/m-datepicker/js/datepicker.js
```diff
@@ -1,179 +1,81 @@
...
+      parsedDate.setMonth(month);
 
-    // or if a dropdown changes, update the datepicker body
-    if (month || month === 0) {
-      this.datepickerBody.setAttribute('month', month);
+      this.datepickerBody.setAttribute('date', parsedDate.toISOString());
     }
   }
 
-  handleChangeDropdownYear = (e) => {
```
AndyOGo: I would revert back to [class fields](https://github.com/tc39/proposal-class-fields), so that context binding is in one place

src/components/m-datepicker/js/datepicker.js
```diff
@@ -1,179 +1,81 @@
...
+    const year = e.detail;
+    if (year) {
+      const newDate = this.datepickerBody.getAttribute('date');
+      const parsedDate = new Date(Date.parse(newDate));
+      parsedDate.setFullYear(year);
+      this.datepickerBody.setAttribute('date', parsedDate.toISOString());
     }
   }
 
-  handleChangeDatepickerBody = (e) => {
```
AndyOGo: same here, please revert to class fields

src/components/m-dropdown/_template.js
```diff
@@ -2,67 +2,32 @@ import html from 'nanohtml';
 import raw from 'nanohtml/raw';
 import classnames from 'classnames';
 
-const arrowIcon = '<axa-icon icon="angle-bracket-down" classes="m-dropdown__icon"></axa-icon>';
+const arrowIcon = '<axa-icon size="sm" icon="angle-bracket-down" classes="m-dropdown__icon"></axa-icon>';
 
-const getItemValue = (itemValue, index) => itemValue === null || itemValue === undefined ? index : itemValue;
-
-const nativeSelect = ({ title, items, size, value }) => html`<div class="${classnames('m-dropdown__select-wrap', {
```
AndyOGo: @jackblackCH 
@LucaMele 
Please rethink this change.
With this update both are rendered, the native `<select>` and an enhanced js based dropdown.
The **original** idea was to show the native `<select>` at mobile only, though it was never finished...

src/components/m-dropdown/index.js
```diff
@@ -1,60 +1,172 @@
...
-      'm-dropdown--in-flow': inFlow,
-      'm-dropdown--gpu': gpu,
-      [`m-dropdown--${size}`]: size,
-    });
+    this.onDropdownClick =
+      on(this, EVENTS.CLICK, DEFAULTS.toggleClass, this.handleDropdownClick, { capture: true, passive: false });
+
+    // the enhanced dropdown is a list with links -> clicks event. the native is a select change event
+    this.onDropdownValueClick =
+      on(this, EVENTS.CLICK, DEFAULTS.selectClass, e => this.handleDropdownValueClick(e), { capture: true, passive: false });
```
AndyOGo: please use class fields here too

src/components/m-dropdown/index.scss
```diff
@@ -80,6 +84,13 @@
 
 .m-dropdown__icon {
   @include dropdown-icon();
+  pointer-events: none;
```
AndyOGo: @LucaMele 
@jackblackCH 
If any style changes are required, they should be either added to an existing mixin or provided by a nother mixin

src/components/m-dropdown/index.scss
```diff
@@ -90,6 +101,21 @@
   @include dropdown--sm();
 }
 
+
+.m-dropdown__list--native {
```
AndyOGo: It was not the goal to have unnecessary DOM.
Though that will work too

src/components/m-dropdown/index.scss
```diff
@@ -120,9 +146,13 @@
   }
 }
 
+.m-dropdown__toggle,
 .m-dropdown__link {
-  @include typo-text-longer();
+  @include typo-decls('typo-18');
```
AndyOGo: @jackblackCH 
We have other more semantice mixins for typography.
Please use them instead

src/components/m-dropdown/mixins/dropdown.scss
```diff
@@ -74,7 +74,6 @@
   &:focus {
     border: none;
     box-shadow: none;
-    outline: none;
```
AndyOGo: @jackblackCH 
@LucaMele 
`outline` is removed by intention here and replaced below with `color`.
Please undo.
Originally this was reviewed and approved with Anja Wagner, since she left axa. @jojackpot  or @kenaxa could be asked.

src/components/m-dropdown/settings/dropdown.scss
```diff
@@ -6,7 +6,7 @@ $dropdown__frame-color: $color-prim-gray-mine-shaft !default;
 $dropdown__frame-color--hover: $color-axa-blue !default;
 $dropdown__frame-y-padding: 10.5px !default;
 $dropdown__frame-y-padding-sm: 5.5px !default;
-$dropdown__frame-x-padding: 20px !default;
+$dropdown__frame-x-padding: 12px !default;
```
AndyOGo: @jackblackCH 
@LucaMele 
Is this reviewed and approved in general by Design/UX.
Because we have a 10-based grid

src/components/m-datepicker/js/datepicker.js
jackblackCH: Yes, made with css media queries only.

src/components/m-datepicker/js/datepicker.js
jackblackCH: true. done.

src/components/m-dropdown/_template.js
jackblackCH: I saw that. IMO it's less complex, more readable and less expensive to render both and hide and show the enhanced version. 

Mobile First -> Take the native as a mobile first approach and enhance it with the list on bigger viewports. no need to rerender.

src/app/_preview.scss
LucaMele: @AndyOGo im generally very worried about `vh vw` but as its supported from all our browsers (as u can see in the provided can i use link) AND is "only" the preview, i say its ok to use

src/components/a-icon/index.scss
LucaMele: seems or is? might i ask you for more details. Can we live with it or in your opinion is better to be less specific in this contest?

I generally worry about specificity as for a consumer for the patterns library is harder to override classes if they need to.

src/components/a-input/_template.js
LucaMele: @AndyOGo cause a input has a "icon" button for an action on the website

src/components/m-dropdown/index.scss
jackblackCH: Agree, this is not a style change though. `pointer-events` is added on purpose in this situation (to allow clicks through that icon)

src/components/m-button/mixins/_button.scss
```diff
@@ -38,10 +38,7 @@ $color-button-border--disabled: get-color($button-color, default, border--disabl
   &:active,
   &:focus {
     cursor: pointer;
-
     text-decoration: none;
-
-    outline: none;
```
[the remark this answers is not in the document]
LucaMele: @AndyOGo as i understand, outline makes it "tabable" which is important for accessibility. Where did we decide to remove it and why? I cant rememeber

src/components/m-datepicker-body/index.js
LucaMele: @AndyOGo agree. But is also part of best practice from google. https://developers.google.com/web/fundamentals/web-components/best-practices 

So we should maybe generate them in future automatically or something like this. But i think for this PR is ok to merge or not?

src/components/m-dropdown/_template.js
LucaMele: mmm ok i need to see the code here in order to have an opinion.. i will get back on this

src/components/m-datepicker-body/index.js
AndyOGo: @LucaMele 
Yep I know that best-practise guide.
Anyway, we do not follow much of it right now (like we would need to use shadow-DOM, no complex data types as attributes, etc.)

src/components/m-dropdown/settings/dropdown.scss
jackblackCH: I suggest we discuss topics like that with designers. I highly agree on pixel perfect solutions. Nevertheless, pixels in CSS are never accurate as they are always a combo of different properties and side effects.

jackblackCH: Design changes are discussed. @jojackpot agreed.

src/components/m-dropdown/index.scss
AndyOGo: I understand.
May I ask you if you have studied the [typography guidelines](https://design.axa.com/web-guidelines/typography)?
If not please do so.
We don't want this meaningless and occasionally wrong numbers, but rather prefer semantic names.

src/components/m-datepicker/js/datepicker.js
jackblackCH: done.

src/components/m-dropdown/settings/dropdown.scss
AndyOGo: As long as it is approved by Design/UX I'm okay with it

src/components/m-dropdown/index.scss
AndyOGo: Okay, did you recognize that the `18` size changes for various breakpoints?
If not please read it mindfully.

And as said the reasoning is semantic code and readability:
> We don't want this meaningless and occasionally wrong numbers, but rather prefer semantic names.

All available mixins are here:
https://github.com/axa-ch/patterns-library/blob/develop/src/styles/typo/_typo-mixins.scss#L73-L253

src/components/a-input/_template.js
AndyOGo: **Off-Topic**:
Interesting.
But why is it combined with basic `<axa-input>` 🤔 
Shouldn't it be separated.

src/components/m-dropdown/index.js
jackblackCH: Oh. thanks. true, this fixes the IE 11 issues, too. Would be nice if you could explain this "magic"  behind there in an upcoming axa code review.

src/components/m-dropdown/index.scss
AndyOGo: Cool.
I would argue that it makes sense to have `pointer-events:none` in general for drop down icons

src/components/m-dropdown/index.js
jackblackCH: done.

src/app/_preview.scss
AndyOGo: Well, if you are worried.
And we obviously have no benefit here compared to `100%`.
Let's just use safe code.

src/components/m-dropdown/index.js
AndyOGo: Indeed, because it's rather obscure, I have even written it down:
https://github.com/axa-ch/patterns-library/blob/develop/CONTRIBUTING.md#constructor

src/components/a-icon/index.scss
LucaMele: @MarekLacoAXA if `.a-icon__svg {}` can live without the need to increase specificity, please do so as @AndyOGo suggests

src/app/_preview.scss
LucaMele: @AndyOGo and @jackblackCH , safe code is anyway. As is not an issue i let @jackblackCH choose what is best.

src/components/a-input/_template.js
LucaMele: no i think is ok, as is a input having a button make sense that the button is inside the input ;)

src/components/m-button/mixins/_button.scss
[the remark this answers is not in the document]
LucaMele: @jackblackCH as it is design change and current buttons are tabable atm [LINK](https://axa-ch.github.io/patterns-library/components/m-button/index.html), please revert outline and add a focus behaviour as @AndyOGo said **OR** ask @jojackpot **OR** @kenaxa for design change permission

src/components/m-datepicker-body/index.js
LucaMele: yepp lets leave it as @jackblackCH did and issue this in a separate task

src/components/m-datepicker-body/index.scss
LucaMele: agree

src/components/m-datepicker-body/index.scss
LucaMele: disagree / neutral to it

src/components/m-datepicker/index.js
LucaMele: @AndyOGo please explain more in detail what you mean or add a code sample. If not possible, @jackblackCH u can ignore this comment

src/components/m-datepicker/index.scss
LucaMele: anyway not plain `px` as it is in our StyleLint and StyleLint is now in our [CoC](https://github.com/axa-ch/patterns-library/blob/develop/CODE_OF_CONDUCT.md)

LucaMele: REF: _We make sure that our code respects ESlint, StyleLint and EditorConfig settings. We can challenge the setting but wont add our own code style if it goes against linting files._

src/components/m-button/mixins/_button.scss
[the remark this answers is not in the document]
AndyOGo: @LucaMele 
@jackblackCH 
We have existing `:focus` behaviour, it only comes from other mixins, like `button--color`:
https://github.com/axa-ch/patterns-library/blob/554e86c369a59ecd21de571d1e1b018bc808d55f/src/components/m-button/mixins/_button.scss#L69-L85

or `button--ghost` mixin:
https://github.com/axa-ch/patterns-library/blob/554e86c369a59ecd21de571d1e1b018bc808d55f/src/components/m-button/mixins/_button.scss#L98-L122

src/components/m-dropdown/index.scss
LucaMele: neutral here on my side

src/components/m-dropdown/index.scss
LucaMele: @AndyOGo this is scss so i cant see what you mean with "unnecessary DOM"

src/components/m-dropdown/index.scss
LucaMele: agree with @AndyOGo , please use one of the defined mixings

src/components/m-dropdown/mixins/dropdown.scss
LucaMele: as said in my privious comment above either focusable or ask ken or jonas

src/components/m-dropdown/settings/dropdown.scss
LucaMele: perfect, so ok for me for this change

src/components/a-input/_template.js
AndyOGo: hmh, okay.
I would rather solve it with composition, i.e. have a third component which composes `<axa-input>` and `<button>` to keep concerns separated.

src/components/m-button/mixins/_button.scss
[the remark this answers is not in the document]
LucaMele: Thank you @AndyOGo for this additional help and code snipped, this is really appreciated

src/components/m-datepicker/index.js
AndyOGo: @LucaMele 
Well, I asked a question. And gave a suggestion to his inline comment:
> This method is triggered up to a millions times.... but it's needed here else, no items are available

So to clarify my question is:
> These values should not change often, right?

Am I correct with above assumption?

Only if my assumption is correct, this fact of "method is triggered up to a millions times" can be mitigated by caching the functions result or a memoization library.

src/components/m-datepicker/js/datepicker.js
AndyOGo: I see.
This approach works too, but was not the original intention.
Whatever the teams decides is okay with me.

src/components/m-dropdown/_template.js
AndyOGo: Okay, as I commented at the related code above.
Whatever the teams sees fitting best will be done

src/components/a-input/_template.js
LucaMele: could also be ok

src/components/m-dropdown/index.scss
AndyOGo: @LucaMele 
This whole code block relates to some Javascript too `DeviceStateObserver` and is also discussed above.

src/components/m-datepicker/index.js
LucaMele: ahhh ok now i understood i didnt get the connection between your comment and the comment on the code :) ok very good suggestions thank you

src/components/m-datepicker/js/datepicker.js
LucaMele: i guess we have to challange both settings and stick with one. For me is not a criteria for PR Stop.

src/components/m-dropdown/index.scss
LucaMele: ahhhhhhh ok got it! the relation to the comment above was missing. Ok so we have unncessery dom cause of media query use instead of the JS solution already in place. ok i would if possible use then the `DeviceStateObserver` rather then adding a new solution and extra DOM. But is not a PR stopper for me

src/components/m-dropdown/mixins/dropdown.scss
AndyOGo: Okay.

Please let me mention again, that all `outline` are replaced by `:focus`.
Concretely about this file check a few lines below please:

https://github.com/axa-ch/patterns-library/blob/554e86c369a59ecd21de571d1e1b018bc808d55f/src/components/m-dropdown/mixins/dropdown.scss#L80-L86

src/components/m-dropdown/_template.js
LucaMele: i checked the code and now i cleaner to look at but we render additional HTML that is not necessary, which could have a performance issue if code would grow a lot. I would prefer to revert as it was as is cleaner form the consumer of the dropdown (if a user instepcts the DOM he will debug over dead code, which might be confusing). Rather that we have more complexity then our consumers therefore agree with @AndyOGo

LucaMele:
@jackblackCH dropdown is broken  on mobile

bildschirmfoto 2019-01-22 um 14 16 05

src/components/m-dropdown/_template.js
LucaMele: @jackblackCH ok i got both arguments and it make sense to have native on mobile and enhanced on desktop. Thought that is not what it was intended and we need to leave the possibility to have both. Some desktop use cases might require just native as well. So i Suggest to have 2: Enhanced which on mobile is native automatically (as you already did @jackblackCH ) AND with a setting JUST native. Please also provide more examples including title (title is what is displayed at the beggining and IS NOT in the options. Therefore once you choose one option, title wont be visibile anymore. Also please add example on "Small" and "Big" sized dropdown)

src/components/m-dropdown/mixins/dropdown.scss
LucaMele: @AndyOGo agree and needs to be fixed

src/components/a-icon/index.scss
jackblackCH: I changed it.  So we have to use a specific icon class "modifier" to pass down that sits on the svg element.

src/components/m-button/mixins/_button.scss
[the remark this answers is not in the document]
jackblackCH: I re-added the `outline: 0`. Meaning at the moment using the `button()` mixin without anything else, includes as a non-accessible button, which I find could be solved better.  Let's keep this topic OT. It's a super important one though.

src/components/m-datepicker-body/index.js
jackblackCH: Sure, let's discuss it in a code review.

src/components/m-datepicker-body/index.scss
jackblackCH: Done.

src/components/m-datepicker/index.js
jackblackCH: True, it's based on props though. Can we discuss this issue/pattern/problem like in a code review and define some "best practices" or, common pitfalls or whatever? for the moment I could not find a proper solution that's not breaking the whole component.

src/components/m-dropdown/_template.js
jackblackCH: I'll re-add the `native` "only" prop discussed with @LucaMele

src/components/m-dropdown/index.scss
jackblackCH: for dropdown icons yes, I don't see a benefit at all to put things into a mixin if you don't reuse it. In this case, if you need the dropdown icon clickable, you are stuck.

src/components/m-dropdown/index.scss
jackblackCH: It's an ok discussion, DeviceStateObserver vs additional DOM. I argue not to use js if we don't need. Additional DOM is not expensive and the DeviceStateObserver is just overkill.

src/components/m-dropdown/index.scss
jackblackCH: I did not need the responsive font in this case, that was the reason I used `typo-18` over `typo-text-longer`

src/components/m-dropdown/mixins/dropdown.scss
jackblackCH: I re-added... Strongly disagree though. Just changing color is not enough, it decreases accessibility a lot for a little visual "issue". Disable people and/or keyboard users are used to an outline and a similar behavior as global as possible, at least page wide. I open this discussion with @jojackpot and @kenaxa .

src/components/m-datepicker/js/datepicker.js
jackblackCH: As long as the observer is used only to add a css class, you can achieve the same with a media query. Please correct, if I missed another thing.

src/components/m-dropdown/_template.js
jackblackCH: It's changed. the `native` prop enforces the visibility of the native on all vp's

src/components/m-datepicker/index.js
jackblackCH: The comment was copied when the method was still in `init`. It is not relevant anymore since the `connectedCallback()` is only called once per component.

package.json
```diff
@@ -15,7 +15,7 @@
     "start": "cross-env NODE_ENV=production npm run serve-build-prod",
     "new": "node stack/tasks/create-a-m-o.js",
     "server": "node stack/server.js",
-    "postinstall": "npm rebuild node-sass",
+    "postinstall": "npm rebuild node-sass && npm run build",
```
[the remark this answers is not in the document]
jackblackCH: thx. nice catch. done. was a temporary change.

package.json
```diff
@@ -80,8 +80,25 @@
   "author": "Luca Mele",
   "license": "Copyright 2017 AXA Versicherungen AG",
   "devDependencies": {
-    "@axa-ch/generic-release": "^1.0.9",
+  },
+  "dependencies": {
```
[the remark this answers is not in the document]
jackblackCH: done.

jackblackCH:
> @jackblackCH there is still an issue with the icon:
> 
> bildschirmfoto 2019-01-31 um 08 28 42
> 
> I think it should not look like this

Fixed.

LucaMele:
Tested and all PR requests are implemented. Mergin now! yaaay

PR #775, opened 2019-01-15
integrate jest

MKaHead:
Jest Integration and necessary config changes

 - regarding future Babel 7 and deprecated env key - add babelrc.js file, if babel 7 is integrated babelrc is obsolete
 - add jest, config, and jest runner script
 - add eslint jest recommented rules
 - remove babel external-helpers out of babel to fix issue, add it in rollup bundle proccess, but kicked out runtime helpers

- add Documentation to Contribution file

## Type of change

 # How Has This Been Tested?
 
 
 
 # Checklist:
 
 - [x] [update]I have made corresponding changes to the documentation


package.json
```diff
@@ -173,5 +177,25 @@
     "@ungap/custom-elements-builtin": "^0.1.1",
     "document-register-element": "^1.12.0",
     "react": ">=0.14.0"
+  },
+  "jest": {
```
AndyOGo: Personally I prefer to keep configuration scoped in separate config files.
May we moved it `jest.config.js` as described here: https://jestjs.io/docs/en/configuration.html

stack/tasks/bundle-demos.js
```diff
@@ -43,7 +43,8 @@ async function buildComponents() {
           .then(result => result.css),
       }),
       babel({
-        runtimeHelpers: true,
```
AndyOGo: By removing this completely we loose reuse of babel-helpers

AndyOGo: Here is an interesting read about how babel's cli and helper optimisation is working:
https://brunoscopelliti.com/a-simple-babel-optimization-i-recently-learned/

AndyOGo: unfortunately official docs are rather obscure:
https://www.npmjs.com/package/babel-plugin-external-helpers

stack/tasks/bundles/app.js
```diff
@@ -1,10 +1,9 @@
 const rollup = require('rollup');
 const uglify = require('rollup-plugin-uglify');
-const babel = require('rollup-plugin-babel');
 const resolve = require('rollup-plugin-node-resolve');
-
 const constants = require('../../constants');
 const common = require('./_common');
+const babel = require('rollup-plugin-babel');
```
AndyOGo: We should import dependencies before relative files

stack/tasks/bundles/app.js
```diff
@@ -19,10 +18,11 @@ async function buildApp() {
         browser: true,
         preferBuiltins: false,
       }),
+      ENV === constants.ENV.PROD ? uglify() : () => {},
```
AndyOGo: uglify should happen after transpilation

stack/tasks/bundle-demos.js
```diff
@@ -43,7 +43,8 @@ async function buildComponents() {
...
+        plugins: ['external-helpers'],
+        externalHelpers: true,
```
AndyOGo: further info at [`rollup-babel-plugin`](https://github.com/rollup/rollup-plugin-babel#helpers)

> If you do not wish the babel helpers to be included in your bundle at all (but instead reference the global `babelHelpers` object), you may set the `externalHelpers` option to `true`:

> By default `externalHelpers` option is set to `false` so babel helpers will be included in your bundle.

LucaMele:
@AndyOGo what framework would u suggest instead? we were also thinking to try mocha

AndyOGo:
@LucaMele 
Karma, mocha, sinon, snapshots with enzym for mocha and real DOM not VDOM

LucaMele:
@MKaHead ok lets stick to our original plan: Lets try Jest and Mocha/Karma etc and compare them. 

The idea of using Karma is actually really good. We can so test various browser locally. I would also see if u can use Karma togheter with Jest.

So my suggestion is to evaluate JEST + KARMA + SNAPSHOT Capability VS Mocha + Sinon + Karma + Snapshot capability

AndyOGo:
Agree.
Though as I understand Jest, it comes with a build-in test-runner within a Node environment only (no browser).
More info here:
https://jestjs.io/docs/en/configuration.html#testenvironment-string
https://github.com/facebook/jest/issues/139

LucaMele:
ok thanks @AndyOGo , i guess that will come up then in the evaluation

AndyOGo:
So we are not the only team with that Jest problem of no browser environment.
There are solutions like:
https://www.npmjs.com/package/jest-browser

But only for headless chrome https://github.com/GoogleChrome/puppeteer

AndyOGo:
Agreed that's the other downside, CI/CD Pipeline support.

Ideally we would have a tooling which allows us to:
- run tests in any browsers, concurrently
- run tests within headless browsers
- run tests directly in node, with fake API's like jsdom (for CI/CD speed)

All of that are different test environments.
And as said karma supports it all, like:
https://www.npmjs.com/package/karma-jsdom-launcher

Issue #791, opened 2019-01-22
Nicer appearing of "error" in form-group (AXAFormGroup)


MarekLacoAXA:
When the error in form-group suddenly appears, then the form gets expanded, which doesn't look very nice.

Ideally the error would be absolutely positioned, or appear smoothly (animated).

Mentioned by @Hettlingen 

![image](https://user-images.githubusercontent.com/43809369/51540041-c05ec300-1e55-11e9-8685-d9bc5a055590.png)

![image](https://user-images.githubusercontent.com/43809369/51540062-c94f9480-1e55-11e9-9d3e-4feaf665dbbd.png)

raphaellueckl:
There is a PR open that fixes the issue in a controversial way. As soon as we have a UX concept, we can start to look at this issue again.

Link: https://github.com/axa-ch/patterns-library/pull/819

LucaMele:
Closing as V2 is released. If issue is still relevant in v2, please re-open

PR #819, opened 2019-02-12
#791 Fix validation error message from axa-form-group

raphaellueckl:
Height is now fixed, even if validation message is not displayed

## Type of change
- [ x] Bug fix (non-breaking change which fixes an issue)

Link: https://github.com/axa-ch/patterns-library/issues/791

### DIFF
--- a/src/components/m-form-group/_template.js
+++ b/src/components/m-form-group/_template.js
@@ -23,7 +23,7 @@ export default function ({
         </div>
 
         ${childrenFragment}
-        <span class="m-form-group__error">${error}</span>
+        <span class="m-form-group__error">${error ? error : '&nbsp;'}</span>
       </fieldset>`,
     html`
       <div class="m-form-group__info js-form-group__info">
--- a/src/components/m-form-group/_template.js
+++ b/src/components/m-form-group/_template.js
@@ -23,7 +23,7 @@ export default function ({
         </div>
 
         ${childrenFragment}
-        <span class="m-form-group__error">${error ? error : '&nbsp;'}</span>
+        <span class="m-form-group__error">${error || '&nbsp;'}</span>
       </fieldset>`,
     html`
       <div class="m-form-group__info js-form-group__info">
--- a/src/components/m-form-group/_preview.html
+++ b/src/components/m-form-group/_preview.html
@@ -12,33 +12,50 @@
     <axa-input value="Pre-filled content" name="test-name"></axa-input>
   </axa-form-group>
   <axa-form-group legend="Test Legend"
-                  info="Caramels liquorice ice cream gingerbread gummi bears. Lemon drops cookie wafer cake. Sweet roll sesame snaps chocolate bar bear claw lemon drops sugar plum cupcake candy canes icing.">
+    info="Caramels liquorice ice cream gingerbread gummi bears. Lemon drops cookie wafer cake. Sweet roll sesame snaps chocolate bar bear claw lemon drops sugar plum cupcake candy canes icing.">
     <axa-input value="Pre-filled content" name="test-name"></axa-input>
   </axa-form-group>
 
   <axa-form-group legend="Test Legend"
-                  info="Caramels liquorice ice cream gingerbread gummi bears. Lemon drops cookie wafer cake. Sweet roll sesame snaps chocolate bar bear claw lemon drops sugar plum cupcake candy canes icing.">
+    info="Caramels liquorice ice cream gingerbread gummi bears. Lemon drops cookie wafer cake. Sweet roll sesame snaps chocolate bar bear claw lemon drops sugar plum cupcake candy canes icing.">
     <axa-input value="Pre-filled content" name="test-name"></axa-input>
   </axa-form-group>
   <axa-form-group legend="Test Legend"
-                  info="Caramels liquorice ice cream gingerbread gummi bears. Lemon drops cookie wafer cake. Sweet roll sesame snaps chocolate bar bear claw lemon drops sugar plum cupcake candy canes icing.">
+    info="Caramels liquorice ice cream gingerbread gummi bears. Lemon drops cookie wafer cake. Sweet roll sesame snaps chocolate bar bear claw lemon drops sugar plum cupcake candy canes icing.">
     <axa-input value="Pre-filled content" name="test-name"></axa-input>
   </axa-form-group>
   <axa-form-group legend="Test Legend"
-                  info="Caramels liquorice ice cream gingerbread gummi bears. Lemon drops cookie wafer cake. Sweet roll sesame snaps chocolate bar bear claw lemon drops sugar plum cupcake candy canes icing.">
+    info="Caramels liquorice ice cream gingerbread gummi bears. Lemon drops cookie wafer cake. Sweet roll sesame snaps chocolate bar bear claw lemon drops sugar plum cupcake candy canes icing.">
     <axa-input value="Pre-filled content" name="test-name"></axa-input>
   </axa-form-group>
 
-  <axa-form-group legend="Test Legend"
-                  error="Caramels liquorice ice cream gingerbread gummi bears."
-                  info="Caramels liquorice ice cream gingerbread gummi bears. Lemon drops cookie wafer cake. Sweet roll sesame snaps chocolate bar bear claw lemon drops sugar plum cupcake candy canes icing.">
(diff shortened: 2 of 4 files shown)


raphaellueckl:
> Better to just remove the DOM node if there is no error.
> 
> Adding whitespace will negatively affect layout

Generally I agree, what is not needed in the DOM, should completely go away, but in this case, this was the bug that was opened. The layout was shifting everytime a validation message appeared. I understand that for this case, where those messages can appear and go away often and and overflow functionality is imho not an option.

Link: https://github.com/axa-ch/patterns-library/issues/791

AndyOGo:
@raphaellueckl 
Thanks. Please check with the design guidelines.
Also the Forms are currently not fully implemented and broken at various parts.

Proper way would be to animate the height smoothly.

Please no whitespace hacks, which where valid for old table layouts but not for modern web development.

raphaellueckl:
> @raphaellueckl
> Thanks. Please check with the design guidelines.
> Also the Forms are currently not fully implemented and broken at various parts.
> 
> Proper way would be to animate the height smoothly.
> 
> Please no whitespace hacks, which where valid for old table layouts but not for modern web development.

Thx for the feedback.
I agree again, but animation is not part of this bugfix imho. It didn't animate before, it doesn't animate now. The issue is about the layout shifting around. Animation should be done in another PR and not combined with this one.

About the whitespace hack: I don't have a better idea, but I would be fine with a better proposal. Setting the height to a fixed value is even more wrong imho, since as soon as anyone is going to change the font height, it will break the design. A whitespace won't do that, the layout will still render properly, even if someone is setting the font size to 40 px.

AndyOGo:
Hey @raphaellueckl .
We haven't introduced each other yet.

May I ask if you are fully familiar with the global design web-guidelines?
https://design.axa.com/web-guidelines/

Especially regarding forms:
https://design.axa.com/web-guidelines/forms-update#design-specifications-for-forms

Regarding the link issue, have you checked if it is an actual issue pre design specs?

AndyOGo:
Also your linked issues #791 states:
> or appear smoothly (animated).

raphaellueckl:
> Hey @raphaellueckl .
> We haven't introduced each other yet.
> 
> May I ask if you are fully familiar with the global design web-guidelines?
> https://design.axa.com/web-guidelines/
> 
> Especially regarding forms:
> https://design.axa.com/web-guidelines/forms-update#design-specifications-for-forms
> 
> Regarding the linked issue, have you checked if it is an actual issue per design specs?

Hi Andreas
Can you tell me, where I went against the guidelines? I searched the document for anything about animations and transitions, but there is nothing mentioned. Why exactly did you post it?

Just telling me "you did wrong", does not help me at this point. What exactly is bad in your opinion? :)
I'm not saying that you're wrong, but can you proof that you are not?

Thanks for the link though, I will definitely have a closer look at it.

> or appear smoothly (animated).

Please kindly consider the "OR" that you quoted. There was no animation before, that means this was never a feature. Why should it become one now? We are still talking about a bugfix here. :) Not saying we should not consider animating it, all I'm saying is, that this is a bugfix and not a feature that I am implementing and messing this structure up is against gitflow. Here is a link to the topic: https://es.atlassian.com/git/tutorials/comparing-workflows/feature-branch-workflow

Thanks for the feedback!

AndyOGo:
@raphaellueckl thanks for your answer.

Please correct me if I'm wrong.

Your solution adds a non-breaking white space `&nbws;`, right?
So as a consequence, the `<span class="m-form-group__error">` will always have space (width and height).
The issue I see here is, that the distance to other form elements without errors get corrupted by this approach.
This is not desired.
Or do I see this wrong?

Regarding Your question:
We don't make sole design decision by ourselves, all has to be approved by either at least global design guidelines or locally by @kenaxa or @jojackpot .
Have you revised this change with them?

raphaellueckl:
> @raphaellueckl thanks for your answer.
> 
> Please correct me if I'm wrong.
> 
> Your solution adds a non-breaking white space `&nbws;`, right?
> So as a consequence, the `<span class="m-form-group__error">` will always have space (width and height).
> The issue I see here is, that the distance to other form elements without errors get corrupted by this approach.
> This is not desired.
> Or do I see this wrong?

The requested functionalityfrom the issue is, that the layout does not shift around upon the appearance of a validation message. As a consequence, we need to have a fixed height for validation messages like that, since validation messages can "come out of nowhere".

> Please also consider all breakpoints and potential multi-line error messages.

I evaluated that and my view on this is, that multiline validation messages are a UX-anti-pattern. And IF there would be a case, where those would still be necessary, then the layout would still shift.

> We don't make sole design decisions by ourselves, all has to be approved by either at least global design guidelines or locally by @kenaxa or @jojackpot .

I sit together with them, but I did not ask yet. Good point!

AndyOGo:
Thanks for your reply @raphaellueckl 

It would be awesome if it would be that easy.
Unfortunately your fixed height approach has this undesired effects:

- no "fix height" guarantee for small screens or too long error messages
- it corrupts the vertical rhythm, i.e. all form group should have a `10px` vertical distance to each other, now with your approach they have `10px` + the font-size/line-height of your `&nbws;`
- regarding i18m - sometimes statements just take much more word in another language to be expressed
- as a matter of fact: using `<br>`, `&nbws;` etc. for layout or spacing is bad-practice, violates separation of concerns, and is not acceptable any more - https://webmasters.stackexchange.com/questions/79199/why-isnt-the-html-break-tag-acceptable-anymore-for-top-and-bottom-margins

As I said at the beginning, the proper way is to utilize CSS and/or JavaScript to build a holistic solution, which can properly handle all cases.

src/components/m-form-group/_template.js
```diff
@@ -23,7 +23,7 @@ export default function ({
         </div>
 
         ${childrenFragment}
-        <span class="m-form-group__error">${error}</span>
+        <span class="m-form-group__error">${error || '&nbsp;'}</span>
```
LucaMele: @raphaellueckl can we wrap <span class="m-form-group__error"> in a IF so that if no error is there it wont render the span tag?

raphaellueckl: I guess we have to. Then we have, what we had before. I guess we need to discuss, how we handle github issues.

LucaMele: @raphaellueckl i have not red all the above arguments, but i will have a look so i have more context. Just at looking at the code seems like a better way to just remove it if is not there. If the height is required, might be easier to have a min-height or something like that so that.

raphaellueckl:
> Thanks for your reply @raphaellueckl
> 
> It would be awesome if it would be that easy.
> Unfortunately your fixed height approach has this undesired effects:
> 
> * no "fix height" guarantee for small screens or too long error messages
> * it corrupts the vertical rhythm, i.e. all form group elements should have a `10px` vertical distance to each other, now with your approach they have `10px` + the font-size/line-height of your `&nbws;`
>   
>   * especially at smaller viewports it reduces the amount of available space to present UI to the user
> * regarding i18n - sometimes statements just take much more words in another language to be expressed

Yes, I'm aware of all that.

> * as a matter of fact: using `<br>`, `&nbws;` etc. for layout or spacing is bad-practice, violates separation of concerns, and is not acceptable any more - https://webmasters.stackexchange.com/questions/79199/why-isnt-the-html-break-tag-acceptable-anymore-for-top-and-bottom-margins

I disagree here. I agree that this is probably not the best solution, but the best I could find. Reason: This is not a mix of content and style. Your quoted stackoverflow-guy (not sure if the guy is even worth a reference, but he has a point) guy states, that content should not mix with style. I don't mix any content with any style. What I do is, I use an empty content, to always have a content. This obviously affects styling, but every div and paragraph (and therefore every content) affects styling, the more content you put in a <p>, the more your website will grow in height. All I am doing is keeping a value in place of the absende of an error. I am not spamming <br>s or anything like that to increase the padding somewhere. <br> is NOT content, and this is what the point of the guy you quote, is.

> 
> As I said at the beginning, the proper way is to utilize CSS and/or JavaScript to build a holistic solution, which can properly handle all cases.

I talked to @jojackpot and he said they are still on the draft for that, so I will freeze this PR now. Maybe we will use animation in the future, but my whole team seems to have different opinions about this topic. :)

Thanks for your inputs, you made some good points.

raphaellueckl:
Since the handling of this issue is unclear, this PR gets freezed until we have a clear way of implementing that. Thanks for all the feedback!

PR #822, opened 2019-02-14
Feature/introduce prettier and travis ci

raphaellueckl:
Fixes # (issue).

 - As discussed in the team, introducing prettier into pipeline.

## Type of change

- [x ] New feature (non-breaking change which adds functionality)
 
 # How Has This Been Tested?
 
Manually and on CI level.
 
 # Checklist:
 

raphaellueckl:
Please have a look. I basically copied the old package.json over here and I guess there is a lot of waste in it. A lot of scripts we will probably able to remove. I also don't know if some of the rules bite each other, but @LucaMele didn't want to have too much changes at once, if I got that correctly.

jackblackCH:
prettier settings looks lgtm. A PR should only contain PR related code, so i.e the `package.json` should come cleaner with prettier related things. The rest of the scripts are not relevant to prettier and can be added independantly.

jackblackCH:
I just see we should have a new line at end of lines (github is mentioning it, too) for sevreal reasons, did we drop that somewhere? and can we add this check to the linter and prettier config?

raphaellueckl:
> I just see we should have a new line at end of lines (github is mentioning it, too) for several reasons. Did we drop that somewhere? and we may ensure that this check is within our linter and/or prettier config? (prettier i thought it's the default)
> https://eslint.org/docs/rules/eol-last
> https://prettier.io/docs/en/options.html#end-of-line

Hm, the ```eol``` ist just the "characters used to mark a newline". We use "\n", which is what I would prefer. But about enforcing a new line at the end of the document, I have to check

jackblackCH:
lgtm

package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "@axa-ch/patterns-library",
+  "version": "3.0.0",
```
AndyOGo: Watch out, this declares a stable `3.0.0` release according to semantic versioning.
Guess what you need is `next`, `alpha`, `beta` etc. tag.

More details at: https://semver.org/

PR #830, opened 2019-02-20
Feature/axa modal

raphaellueckl:
Fixes # (issue).

 - Introducing modal functionality

## Type of change

- [x ] New feature (non-breaking change which adds functionality)
 
 # How Has This Been Tested?
 
Manually
 
 # Checklist:
 

src/assets/icons/broken_glass_blue.svg
```diff
@@ -0,0 +1 @@
...0,0-2-.23l-1.14,10-4.2,1-3.75-1.64-1.81-7.3a1,1,0,0,0-1.94.48l3.08,12.46-3.41,3.44L21.68,20l-9.44-8.52A1,1,0,1,0,10.9,13l29,26.18-9.33,5.66H26.77L16,40.37h0L3.38,35.16A1,1,0,0,0,2.62,37l12.21,5,10,24.5L5.54,84.09a1,1,0,1,0,1.35,1.48L31.53,63.15l8.41,3.24L30,86.56a1,1,0,0,0,1.79.89L43.48,63.73l7.22-8.21,2.76,11.14a1,1,0,0,0,.41.59L79.58,84.58a1,1,0,0,0,1.12-1.66L66.12,73.09l-6.28-11,.59-1.19,9.22-.93L72,61.79a1,1,0,1,0,1.24-1.57L70.6,58.12h0l-6.73-5.29L62,47.13l4.5-5.46L87.19,52.14a1,1,0,1,0,.9-1.79L69,40.71,93.19,36a1,1,0,0,0-.38-2Zm-35.65-.15-5.63,6.78L50.09,37,51.26,26.8ZM17.42,43.13l8.77,3.63a1,1,0,0,0,.38.08h4.29a1,1,0,0,0,.52-.15l1.72-1,4.07,9.67-10.77,9.8Zm42.66,3.15a1,1,0,0,0-.18.95l2.14,6.5a1,1,0,0,0,.33.47l5,3.94-7.71.78a1,1,0,0,0-.8.55l-1.07,2.17a1,1,0,0,0,0,.94l4.55,8L55.31,65.8,52.19,53.18a1,1,0,0,0-1.72-.42l-8.57,9.75a1,1,0,0,0-.15.22l-.92,1.86-7.64-2.94L39,56.32a1,1,0,0,0,.25-1.13L34.82,44.6l7.27-4.41a1,1,0,0,0,.15-1.6L26.43,24.32,38,29.41a1,1,0,0,0,1.12-.21l4.29-4.33a1,1,0,0,0,.26-.94l-.8-3.25,2.65,1.16a1,1,0,0,0,.4.08l.23,0,3.82-.9L48.08,37.05a1,1,0,0,0,.06.48L50.28,43a1,1,0,0,0,.77.62h.16a1,1,0,0,0,.77-.36l7.55-9.08-.74,7.26a1,1,0,0,0,1,1.1l.19,0,3.86-.76Z"/></svg>
```
LucaMele: Please use `<path fill="currentColor"`. instead of inline style so that a user can override the color

src/assets/icons/check_circle_outline.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24"><path fill="none" d="M0 0h24v24H0V0zm0 0h24v24H0V0z"/><path d="M16.59 7.58L10 14.17l-3.59-3.58L5 12l5 5 8-8zM12 2C6.48 2 2 6.48 2 12s4.48 10 10 10 10-4.48 10-10S17.52 2 12 2zm0 18c-4.42 0-8-3.58-8-8s3.58-8 8-8 8 3.58 8 8-3.58 8-8 8z"/></svg>
```
LucaMele: Please use `<path fill="currentColor"`. instead of inline style so that a user can override the color

src/assets/icons/rocking_chair_blue.svg
```diff
@@ -0,0 +1 @@
...1,0,0,0,0-10.22h-3.1V36h3.69a1,1,0,0,0,0-2l-34.9,0C33.33,23.87,29.68,14.8,24.71,7.8a4.25,4.25,0,0,0-2.83-1.75,4.38,4.38,0,0,0-4.3,6.81c6.68,9.44,10.66,24.52,10.66,40.34a71.26,71.26,0,0,1-4.36,25.28,40.78,40.78,0,0,1-5.44-2.3A3.74,3.74,0,0,0,15,82.82C23.64,87.45,35.36,90,48,90s24.37-2.56,33-7.2a3.75,3.75,0,0,0,.92-5.92ZM72.79,56.69a3.11,3.11,0,0,1-3.1,3.11H30c.15-2,.23-4.07.24-6.22h14a.57.57,0,0,0,.45,0H54.81a.57.57,0,0,0,.45,0H65.36a.57.57,0,0,0,.45,0h3.88A3.11,3.11,0,0,1,72.79,56.69ZM30.22,51.58A93.92,93.92,0,0,0,28.69,36h5a95.23,95.23,0,0,1,1.53,15.61Zm15.27,0V36H54V51.58Zm19.1,0H56V36h8.55ZM43.48,36V51.58H37.25A97.29,97.29,0,0,0,35.76,36ZM19.75,8.48A2.43,2.43,0,0,1,21.57,8,2.26,2.26,0,0,1,23.08,9c4.75,6.69,8.25,15.35,10.26,25H28.29c-1.84-8.7-4.93-16.39-9.07-22.25A2.37,2.37,0,0,1,19.75,8.48Zm10,53.32H65a64.1,64.1,0,0,0,4.17,17.61A79.06,79.06,0,0,1,48,82.13a77.87,77.87,0,0,1-22.19-3A73.74,73.74,0,0,0,29.8,61.8ZM81,79.77a1.74,1.74,0,0,1-.9,1.27c-8.26,4.42-20,7-32.09,7s-23.8-2.53-32.06-6.94a1.75,1.75,0,0,1-.41-2.79h0a1.72,1.72,0,0,1,1.22-.5,1.71,1.71,0,0,1,.79.19,43.83,43.83,0,0,0,6.64,2.73h0A78.73,78.73,0,0,0,48,84.13c11.84,0,22.94-2.25,30.47-6.17A1.74,1.74,0,0,1,81,79.77Z"/></svg>
```
LucaMele: Please use `<path fill="currentColor"`. instead of inline style so that a user can override the color

src/components/o-modal/_preview.html
```diff
@@ -0,0 +1,76 @@
...
+<script>
+  function openModal() {
+    const modal = document.getElementsByTagName('axa-modal')[0];
+    modal.setAttribute('open', '');
+    modal.open = true;
+  }
+</script>
+
+<style>
+  h2 {
```
LucaMele: even is just for demo purposes, this will destroy all other components so please add a scope

src/components/o-modal/_preview.html
```diff
@@ -0,0 +1,76 @@
...
+    font-family: PublicoHeadline;
+    font-weight: bold;
+    color: #333333;
+    font-size: 36px;
+    margin-bottom: 50px;
+  }
+
+  p {
```
LucaMele: dtto

src/components/o-modal/_preview.html
```diff
@@ -0,0 +1,76 @@
...
+    font-size: 20px;
+    line-height: 1.5;
+    color: #333333;
+  }
+
+  section {
```
LucaMele: dtto

src/components/o-modal/_preview.html
```diff
@@ -0,0 +1,76 @@
...
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    margin-top: 60px;
+  }
+
+  article {
```
LucaMele: dtto

src/components/o-modal/_preview.html
```diff
@@ -0,0 +1,76 @@
...
+      Schützt Sie zusätzlich gegen Forderungen anderer.
+    </div>
+  </section>
+</axa-modal>
+
+<script>
+  function openModal() {
+    const modal = document.getElementsByTagName('axa-modal')[0];
+    modal.setAttribute('open', '');
+    modal.open = true;
```
LucaMele: should this not be set inside the WC elemnt?

src/components/o-modal/_preview.html
LucaMele: Reproduce: open http://localhost:3000/organisms.html

src/components/o-modal/_preview.html
```diff
@@ -0,0 +1,76 @@
+<!--Please create here a HTML example by using just default HTML tags-->
+
+<button onclick="openModal()">Open modal</button>
+
+<axa-modal open>
```
LucaMele: i would default wise not have it open as it will overlap over all other components: http://localhost:3000/organisms.html

src/components/o-modal/_template.js
```diff
@@ -0,0 +1,23 @@
+import html from 'nanohtml';
+
+export default ({ open }, childrenFragment) => {
+  const closeModal = () => {
+    const modal = document.getElementById('modal-wrapper');
```
LucaMele: Please do not use id.
Reason: There can be 2 modals on one page. Having both with the same ID will result in invalid HTML.

src/components/o-modal/_template.js
```diff
@@ -0,0 +1,23 @@
...
+    modal.style.display = 'none';
```
LucaMele: please use Modifier class instead of overriding style.

Reason: Specificity increases in this way too much which can lead to difficulty in overriding styles

src/components/o-modal/_template.js
```diff
@@ -0,0 +1,23 @@
...
+    modal.classList.add('m-modal--hidden');
```
LucaMele: Here you have the modifier, so why display none in the row above?

src/components/o-modal/_template.js
```diff
@@ -0,0 +1,23 @@
+import html from 'nanohtml';
+
+export default ({ open }, childrenFragment) => {
+  const closeModal = () => {
```
LucaMele: I would put this logic into the `index.js` as this is just templating. But its just an opinion

src/components/o-modal/_template.js
```diff
@@ -0,0 +1,23 @@
...
+    const modal = document.getElementById('modal-wrapper');
+    modal.style.display = 'none';
+    modal.classList.add('m-modal--hidden');
+  };
+
+  const closeModalByClickingOutside = (event) => {
+    if (event.target.id === 'modal-wrapper') closeModal();
```
LucaMele: please never use id as explained before

src/components/o-modal/_template.js
```diff
@@ -0,0 +1,23 @@
...
+  };
+
+  return html`
+    <div id="modal-wrapper" class="m-modal ${open ? '' : 'm-modal--hidden'}" onclick="${closeModalByClickingOutside}">
```
LucaMele: article would be better here instead of div. 

Reason: 
`The article element represents a complete, or self-contained, composition in a document, page, application, or site and that is, in principle, independently distributable or reusable, e.g. in syndication. This could be a forum post, a magazine or newspaper article, a blog entry, a user-submitted comment, an interactive widget or gadget, or any other independent item of content.` https://html.spec.whatwg.org/multipage/sections.html#the-article-element

And as discussed, never relay on global window objects. please use add event listener instead of using onclick

src/components/o-modal/_template.js
```diff
@@ -0,0 +1,23 @@
...
+      <div class="m-modal__content">
+        <span class="m-modal__close-button"><axa-icon icon="cross-gap" onclick="${closeModal}"></axa-icon></span>
```
LucaMele: no onclick

LucaMele: span is used only for text, not for containing other elements

`Contexts in which this element can be used:
Where phrasing content is expected.`
https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-span-element

src/components/o-modal/index.js
```diff
@@ -0,0 +1,63 @@
...
+  static get observedAttributes() {
+    return ['open'];
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+
+    const axaButtons = document.querySelectorAll('axa-button[data-modal-close]');
+    if (axaButtons && axaButtons.length) {
+      axaButtons.forEach((button) => {
```
LucaMele: some browser wont work with forEach as it does NOT return a native array. It returns a "A non-live NodeList" https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll#Return_value

You can transform it like so: `export const querySelectorAll = (selector, parent = document) => [].slice.call(parent.querySelectorAll(selector))`

src/components/o-modal/index.js
```diff
@@ -0,0 +1,63 @@
...
+        button.onclick = this.closeModal;
+      });
+    }
+
+    document.addEventListener('keyup', (event) => {
```
LucaMele: please be scoped with event listeners as otherwise when listens to keyup, all the other one will. Use `this.` instead

src/components/o-modal/index.js
```diff
@@ -0,0 +1,63 @@
...
+      this.closeModalOnEsc(event);
+    });
+  }
+
+  disconnectedCallback() {
+    document.removeEventListener('keyup');
```
LucaMele: dtto. If is on document, you will destroy all modals not just this one

src/components/o-modal/index.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  closeModal() {
+    const modal = document.getElementById('modal-wrapper');
```
LucaMele: dtto ID.

src/components/o-modal/_template.js
LucaMele: my i suggest usage of `button` tag instead?

LucaMele:
Broken layout when screen with much larger then height:

bildschirmfoto 2019-02-25 um 09 46 55

LucaMele:
bildschirmfoto 2019-02-25 um 09 47 59

raphaellueckl:
All liked comments until here were corrected.

src/assets/icons/broken_glass_blue.svg
```diff
@@ -0,0 +1 @@
...
```
LucaMele: please remove Styles from SVGs

src/assets/icons/rocking_chair_blue.svg
```diff
@@ -0,0 +1 @@
...
```
LucaMele: PLease remove Styles `<style>.cls-1{fill:#00008f;}</style>`

src/components/o-modal/_preview.html
```diff
@@ -0,0 +1,80 @@
+<!--Please create here a HTML example by using just default HTML tags-->
+
+<button onclick="openModal()">Open modal</button>
+
+<axa-modal class="axa-modal-preview">
+  <h2>Prämie berechnen</h2>
+  <p><strong>Mehr als 80%</strong> unserer Kunden kombinieren die Hausratversicherung mit einer Privathaftpflichtversicherung.</p>
+  <p>Was möchten Sie?</p>
+
+  <section class="grid">
```
LucaMele: würde für das preview nicht selber classe schreiben sondern `axa-container` oder axa griod verwedenen

src/components/o-modal/_preview.html
```diff
@@ -0,0 +1,80 @@
...
+    <div>
+      <img class="huge" src="/assets/icons/rocking_chair_blue.svg" />
```
LucaMele: please use `axa-icon` instead

src/components/o-modal/index.js
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  closeModalOnEsc(event) {
+    const key = event.key || event.keyCode;
+    if (key === 'Escape' || key === 'Esc' || key === 27) {
+      this.closeModal();
+    }
+  }
+
+  closeModalFunction = () => this.closeModal();
```
LucaMele: did u test this on IE11?

LucaMele: due to the proptype chain issue the new polyfill have?

src/components/o-modal/index.js
```diff
@@ -0,0 +1,83 @@
...
+    const closeButton = this.querySelector('.js-modal-close-button');
+    closeButton.addEventListener('click', this.closeModalFunction);
+  }
+
+  disconnectedCallback() {
+    this.removeEventListener('keyup');
+
+    const axaButtons = this.querySelectorAll('axa-button[data-modal-close]');
+    if (axaButtons && axaButtons.length) {
+      axaButtons.forEach((button) => {
```
LucaMele: careful here again. use array from or the other method with slice

src/components/o-modal/index.scss
```diff
@@ -0,0 +1,68 @@
+@import '../../styles/mixins/respond';
+
+$modal-background-color: #fff;
```
LucaMele: this should be taken from settings

src/components/o-modal/index.scss
```diff
@@ -0,0 +1,68 @@
...
+$close-button-color: #00008f;
```
LucaMele: dtto

src/components/o-modal/index.scss
```diff
@@ -0,0 +1,68 @@
...
+$close-button-color-hovered: #000;
```
LucaMele: dtto

src/components/o-modal/index.scss
```diff
@@ -0,0 +1,68 @@
...
+.m-modal__close-button:focus {
+  color: $close-button-color-hovered;
+}
+
+/* Override browser button styles */
+button.m-modal__close-button {
+  background-color: transparent;
+  border: none;
+  font-family: inherit;
+  -moz-appearance: none;
```
LucaMele: please remove vendor specific css as we have autoprefixer

src/components/o-modal/index.scss
```diff
@@ -0,0 +1,68 @@
...
+  -webkit-appearance: none;
```
LucaMele: dtto

src/components/o-modal/index.scss
```diff
@@ -0,0 +1,68 @@
...
+  appearance: none;
+  cursor: pointer;
```
LucaMele: do we need cursor pointer for a button?

src/assets/icons/broken_glass_blue.svg
raphaellueckl: I downloaded the icons like this. It contained the color in the name. I removed it in the SVGs, but also removed it from the file name that mentioned it.

src/components/o-modal/_preview.html
raphaellueckl: I cannot "blackbox-style" it, at least I did not find a way. I don't want to go and change the styling by accessing the internal classes of that axa-icon, because as soon as those internal css classes change, my design would break with it. Since this is only a preview, I think it's better to not have an additional dependency there from here with this situation. But this is probably worth a discussion.

src/components/o-modal/index.js
raphaellueckl: I requested OpenPc to test exactly that. But that will probably take some time.

raphaellueckl:
todo: esc without modal focus

src/components/o-modal/index.js
raphaellueckl: I wanted to remove that part anyways and leave it up to the GC to clean it up, so that's what I did instead of adapting it like proposed.

src/components/o-modal/index.scss
raphaellueckl: this is the default. the question would imho rather be, if we do not need it? the styleguide does not mention it, the native browser element has it.

raphaellueckl: I just found out that there is a sass file I can import that is doing exactly what I did here. I will just import that. In there, the cursor is also set to pointer.

src/components/o-modal/index.js
```diff
@@ -51,9 +51,7 @@ class AXAModal extends BaseComponentGlobal {
     const axaButtonsRaw = this.querySelectorAll('axa-button[data-modal-close]');
     const axaButtons = Array.from(axaButtonsRaw);
     if (axaButtons && axaButtons.length) {
-      axaButtons.forEach((button) => {
-        button.addEventListener('click', this.closeModalFunction);
-      });
+      axaButtons.forEach(button => button.addEventListener('click', this.closeModalFunction));
```
AndyOGo: Why no event delegation?

src/components/o-modal/index.scss
```diff
@@ -48,19 +46,14 @@ $close-button-color-hovered: #000;
 
 .m-modal__close-button {
   float: right;
-  color: $close-button-color;
+  color: $color-axa-blue;
 }
 
 .m-modal__close-button:hover,
```
AndyOGo: In case you need to rebuild a button, you have plenty of generic mixins available from:
https://github.com/axa-ch/patterns-library/blob/develop/src/components/m-button/mixins/_button.scss

raphaellueckl: Hi Andy, this button is more or less only an "X" to close the modal. But good point anyways, I would say as soon as we need that X somewhere else too, I could move it to this button sass file.

AndyOGo: Ah I see, then you can ignore my comment

src/components/o-modal/index.js
raphaellueckl: Testing on IE should no longer be necessary, since the implementation of this method changed (naturally, not as workaround).

src/components/o-modal/_preview.html
LucaMele: @raphaellueckl please use `axa-container` instead of creating a custom grid class

src/components/o-modal/_preview.html
```diff
@@ -0,0 +1,87 @@
...
+      <svg class="huge">
+        <use xlink:href="#src--assets--icons--rocking_chair" href="#src--assets--icons--rocking_chair" />
+      </svg>
+      <svg class="huge">
+        <use xlink:href="#src--assets--icons--broken_glass" href="#src--assets--icons--broken_glass" />
+      </svg>
+    </div>
+    <div>
+      <svg class="huge">
+        <use xlink:href="#src--assets--icons--rocking_chair" href="#src--assets--icons--rocking_chair" />
```
LucaMele: please use `<axa-icon>`

src/components/o-modal/_preview.html
```diff
@@ -0,0 +1,87 @@
...
+  <p><strong>Mehr als 80%</strong> unserer Kunden kombinieren die Hausratversicherung mit einer Privathaftpflichtversicherung.</p>
+  <p>Was möchten Sie?</p>
+
+  <section class="grid">
+    <div>
+      <svg class="huge">
+        <use xlink:href="#src--assets--icons--rocking_chair" href="#src--assets--icons--rocking_chair" />
+      </svg>
+      <svg class="huge">
+        <use xlink:href="#src--assets--icons--broken_glass" href="#src--assets--icons--broken_glass" />
```
LucaMele: Please use `<axa-icon>`

src/components/o-modal/_preview.html
```diff
@@ -0,0 +1,87 @@
...
+
+<axa-modal class="axa-modal-preview">
+  <h2>Prämie berechnen</h2>
+  <p><strong>Mehr als 80%</strong> unserer Kunden kombinieren die Hausratversicherung mit einer Privathaftpflichtversicherung.</p>
+  <p>Was möchten Sie?</p>
+
+  <section class="grid">
+    <div>
+      <svg class="huge">
+        <use xlink:href="#src--assets--icons--rocking_chair" href="#src--assets--icons--rocking_chair" />
```
LucaMele: Please use `<axa-icon>`

src/components/o-modal/_template.js
```diff
@@ -0,0 +1,10 @@
+import html from 'nanohtml';
+
+export default ({ open }, childrenFragment) => html`
+  <div class="js-modal m-modal ${open ? '' : 'm-modal--hidden'}" tabindex="0">
```
LucaMele: As said in previous comment, this should be `<article>`

src/components/o-modal/index.js
```diff
@@ -0,0 +1,69 @@
...
+  get open() {
+    return this.getAttribute('open');
+  }
+
+  closeModal() {
+    this.removeAttribute('open');
+  }
+
+  closeModalByClickingOutside(event) {
+    if (event.target === this.querySelector('.js-modal')) this.closeModal();
```
LucaMele: @raphaellueckl i think this is not in the ESLint rules, therefore you dont have to change it, but i highly suggest to always use brackets. This can lead to issues in future.

src/components/o-modal/index.js
```diff
@@ -0,0 +1,69 @@
...
+    if (key === 'Escape' || key === 'Esc' || key === 27) {
+      this.closeModal();
+    }
+  }
+
+  init() {
+    super.init({ styles, template });
+  }
+
+  static get observedAttributes() {
```
LucaMele: @raphaellueckl this method is always on top as every static property so that when u look at the file is immediatly visible what properties are accessible from outside. Please move it to the top

src/components/o-modal/index.js
raphaellueckl: Good point. IMHO, if I do not expect a ton of buttons in there, readability goes over performance. Usually, I guess a modal will have two buttons mostly, maybe in some cases ten, who knows. Therefore I would go for readability. IMHO this is more readable than the complexity layer that event delegation would add.

src/components/o-modal/index.js
```diff
@@ -0,0 +1,69 @@
...
+
+    document.addEventListener('keyup', event => this.closeModalOnEsc(event));
+    this.addEventListener('click', event => this.closeModalByClickingOutside(event));
+
+    const closeButton = this.querySelector('.js-modal-close-button');
+    closeButton.addEventListener('click', closeModalFunction);
+  }
+
+  disconnectedCallback() {
+    document.removeEventListener('keyup', this.closeModalOnEsc);
```
LucaMele: @raphaellueckl im not sure this can work. `this.closeModalOnEsc` is another function compared to ` event => this.closeModalOnEsc(event)`. I think they must be the same in order to be deleted.

check out here: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener

LucaMele: @AndyOGo do u know if this works?

AndyOGo: That creates a possible memory leak.
Agree, definitely learn how events work in the DOM.
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener

I have written utilities to make working with native events in a cross-browser manner easier, see the `on` helper:
https://github.com/axa-ch/patterns-library/blob/develop/src/js/on.js

src/components/o-modal/_preview.html
raphaellueckl: I tried but I cannot resize it without blackbox-styling it.

src/components/o-modal/index.js
raphaellueckl: Right, interesting that eslint does not complain about that.

src/components/o-modal/index.js
raphaellueckl: Cool, thanks for the feedback guys!

src/components/o-modal/index.js
LucaMele: yepp

src/components/o-modal/_preview.html
LucaMele: should be able to pass a class down using `classes` attribute

src/components/o-modal/index.js
raphaellueckl: Please have a look at the implementation now.

src/components/o-modal/_preview.html
raphaellueckl: What the fuck... you are right.

src/components/o-modal/index.js
```diff
@@ -0,0 +1,75 @@
...
+    this.closeModalOnEscFunction = event => this.closeModalOnEsc(event);
+    const closeModalFunction = () => this.closeModal();
+
+    const axaButtonsRaw = this.querySelectorAll('axa-button[data-modal-close]');
+    const axaButtons = Array.from(axaButtonsRaw);
+    if (axaButtons && axaButtons.length) {
+      axaButtons.forEach(button => button.addEventListener('click', closeModalFunction));
+    }
+
+    this.onEscReleased = on(document, EVENTS.KEYUP, this.closeModalOnEscFunction);
```
AndyOGo: Better.
Because custom elements can be [adopted](https://html.spec.whatwg.org/multipage/custom-elements.html#custom-element-reactions:concept-node-adopt), you should always use the [owner-document](https://dom.spec.whatwg.org/#dom-node-ownerdocument).

src/components/o-modal/index.js
```diff
@@ -0,0 +1,75 @@
...
+
+    this.addEventListener('click', event => this.closeModalByClickingOutside(event));
+
+    const closeButton = this.querySelector('.js-modal-close-button');
+    closeButton.addEventListener('click', closeModalFunction);
```
AndyOGo: Please also use `on` for above event handlers and clean the up within `disconnectedCallback`

raphaellueckl: @AndyOGo As far as I can tell, this is an overkill, no? Because if the element is removed, the garbage collector will clean up remaining listeners that point to that element and its children.

AndyOGo: Well, let's find it out. Study the specs of [`disconnectedCallback`](https://html.spec.whatwg.org/multipage/custom-elements.html#custom-element-reactions:becomes-disconnected) and all related [`[CEReactions]`](https://html.spec.whatwg.org/multipage/custom-elements.html#cereactions) IDL extended attributes.

LucaMele: I also think its a bit of an overkill.. never than less, after reading this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Memory_Management#Release_when_the_memory_is_not_needed_anymore i might change my mind and might be better to clean up. Sorry @raphaellueckl

LucaMele: @AndyOGo ok seems that the algorithm is not clear but it should be able to handle references. As here is a recference to a dom node, i would be surprised if the garbage collector cannot handle this cleaning. Therefore for me is ok to not add additional code and relay on the GC. Unless there is a code proof that shows this, i am neutral to this so for me is ok to merge.

AndyOGo: I'm sure you will find it out.
I can only suggest to study the specs and observe the implementation of GC's in various JS engines like V8, SpiderMonkey, JScript, and so on

LucaMele: https://v8.dev/blog/tracing-js-dom This is a good read

LucaMele:
Please add a higher z-index to the modal

bildschirmfoto 2019-02-28 um 14 08 17

src/assets/icons/broken_glass.svg
```diff
@@ -0,0 +1 @@
...0,0-2-.23l-1.14,10-4.2,1-3.75-1.64-1.81-7.3a1,1,0,0,0-1.94.48l3.08,12.46-3.41,3.44L21.68,20l-9.44-8.52A1,1,0,1,0,10.9,13l29,26.18-9.33,5.66H26.77L16,40.37h0L3.38,35.16A1,1,0,0,0,2.62,37l12.21,5,10,24.5L5.54,84.09a1,1,0,1,0,1.35,1.48L31.53,63.15l8.41,3.24L30,86.56a1,1,0,0,0,1.79.89L43.48,63.73l7.22-8.21,2.76,11.14a1,1,0,0,0,.41.59L79.58,84.58a1,1,0,0,0,1.12-1.66L66.12,73.09l-6.28-11,.59-1.19,9.22-.93L72,61.79a1,1,0,1,0,1.24-1.57L70.6,58.12h0l-6.73-5.29L62,47.13l4.5-5.46L87.19,52.14a1,1,0,1,0,.9-1.79L69,40.71,93.19,36a1,1,0,0,0-.38-2Zm-35.65-.15-5.63,6.78L50.09,37,51.26,26.8ZM17.42,43.13l8.77,3.63a1,1,0,0,0,.38.08h4.29a1,1,0,0,0,.52-.15l1.72-1,4.07,9.67-10.77,9.8Zm42.66,3.15a1,1,0,0,0-.18.95l2.14,6.5a1,1,0,0,0,.33.47l5,3.94-7.71.78a1,1,0,0,0-.8.55l-1.07,2.17a1,1,0,0,0,0,.94l4.55,8L55.31,65.8,52.19,53.18a1,1,0,0,0-1.72-.42l-8.57,9.75a1,1,0,0,0-.15.22l-.92,1.86-7.64-2.94L39,56.32a1,1,0,0,0,.25-1.13L34.82,44.6l7.27-4.41a1,1,0,0,0,.15-1.6L26.43,24.32,38,29.41a1,1,0,0,0,1.12-.21l4.29-4.33a1,1,0,0,0,.26-.94l-.8-3.25,2.65,1.16a1,1,0,0,0,.4.08l.23,0,3.82-.9L48.08,37.05a1,1,0,0,0,.06.48L50.28,43a1,1,0,0,0,.77.62h.16a1,1,0,0,0,.77-.36l7.55-9.08-.74,7.26a1,1,0,0,0,1,1.1l.19,0,3.86-.76Z"/></svg>
```
LucaMele: Please remove id cause can lead to incorrect HTML

src/assets/icons/rocking_chair.svg
```diff
@@ -0,0 +1 @@
...1,0,0,0,0-10.22h-3.1V36h3.69a1,1,0,0,0,0-2l-34.9,0C33.33,23.87,29.68,14.8,24.71,7.8a4.25,4.25,0,0,0-2.83-1.75,4.38,4.38,0,0,0-4.3,6.81c6.68,9.44,10.66,24.52,10.66,40.34a71.26,71.26,0,0,1-4.36,25.28,40.78,40.78,0,0,1-5.44-2.3A3.74,3.74,0,0,0,15,82.82C23.64,87.45,35.36,90,48,90s24.37-2.56,33-7.2a3.75,3.75,0,0,0,.92-5.92ZM72.79,56.69a3.11,3.11,0,0,1-3.1,3.11H30c.15-2,.23-4.07.24-6.22h14a.57.57,0,0,0,.45,0H54.81a.57.57,0,0,0,.45,0H65.36a.57.57,0,0,0,.45,0h3.88A3.11,3.11,0,0,1,72.79,56.69ZM30.22,51.58A93.92,93.92,0,0,0,28.69,36h5a95.23,95.23,0,0,1,1.53,15.61Zm15.27,0V36H54V51.58Zm19.1,0H56V36h8.55ZM43.48,36V51.58H37.25A97.29,97.29,0,0,0,35.76,36ZM19.75,8.48A2.43,2.43,0,0,1,21.57,8,2.26,2.26,0,0,1,23.08,9c4.75,6.69,8.25,15.35,10.26,25H28.29c-1.84-8.7-4.93-16.39-9.07-22.25A2.37,2.37,0,0,1,19.75,8.48Zm10,53.32H65a64.1,64.1,0,0,0,4.17,17.61A79.06,79.06,0,0,1,48,82.13a77.87,77.87,0,0,1-22.19-3A73.74,73.74,0,0,0,29.8,61.8ZM81,79.77a1.74,1.74,0,0,1-.9,1.27c-8.26,4.42-20,7-32.09,7s-23.8-2.53-32.06-6.94a1.75,1.75,0,0,1-.41-2.79h0a1.72,1.72,0,0,1,1.22-.5,1.71,1.71,0,0,1,.79.19,43.83,43.83,0,0,0,6.64,2.73h0A78.73,78.73,0,0,0,48,84.13c11.84,0,22.94-2.25,30.47-6.17A1.74,1.74,0,0,1,81,79.77Z"/></svg>
```
LucaMele: dtto

src/components/o-modal/_preview.html
```diff
@@ -0,0 +1,79 @@
+<button onclick="openModal()">Open modal</button>
```
LucaMele: please use `type="button"` as the default is submit which is not a good idea outside `<form>`

src/components/o-modal/index.js
LucaMele: @raphaellueckl please add a comment with a todo when we migrate to v2 where we have shadow dom or change it to owner document please. Thanks andy

raphaellueckl:
@LucaMele Regarding your last comment (I cannot reply to it for some reason): Some events indeed bubble out of a shadow dom, including the keyup. So no matter if we later use a closed shadow root, that should still work.

https://developers.google.com/web/fundamentals/web-components/shadowdom#events

PR #831, opened 2019-02-21
Add AXA logo and AXA styles to the theme

jackblackCH:
Improves the look and feel of storybook a bit into AXA styles
- Adds the Logo
- Adds the title
- AXA Blue

Can be easily improved and adapted in the future by enhancing the new config.

axa-storybook1

https://github.com/axa-ch/patterns-library/projects/2#card-17938521


.storybook/config.js
```diff
@@ -8,6 +11,29 @@ import 'core-js/fn/object/create';
...
+    theme: {
+      ...themes.normal,
+      mainTextFace: '"Source Sans Pro", Arial, sans-serif;',
+      menuLink: {
+        color: 'black',
+      },
+      brand: {
+        height: '55px',
+        paddingLeft: '65px',
+        background: 'url("https://axa.ch/etc/clientlibs/axa/frontend-lib/images/logo-AXA.svg") no-repeat',
```
LucaMele: I wouldn’t do this. Rather local assets folder

package.json
```diff
@@ -9,11 +9,12 @@
     "storybook": "start-storybook -p 9001 -c .storybook"
   },
   "dependencies": {
+    "@skatejs/val": "^0.4.0",
+    "@storybook/addon-options": "^4.1.13",
```
LucaMele: This is not a dependency. Is a devDependecy cause our consumer do not have to install it

jackblackCH: Fine. Done.

.storybook/config.js
jackblackCH: Fine done.

package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "start": "npm run storybook",
     "test": "eslint src/",
-    "storybook": "start-storybook -p 9001 -c .storybook"
+    "storybook": "start-storybook -p 9001 -c .storybook -s ./src/static"
```
raphaellueckl: Should we really keep this stuff in the source folder? IMHO src is for code only.

LucaMele: its actually a good point. Agree with you

jackblackCH: True, but we need to access files from components, too?

raphaellueckl: I don't understand? Do you mean the storybook files or the components itself? Why don't we just reference the components? AFAIK it is not a requirement to keep storybook and sourcecode both in the src folder to work, but maybe I am misinterpreting.

PR #834, opened 2019-02-25
Dropdown selection and title issues

### LINKED ISSUE #833: Dropdown

 - [x] Bug
 - [ ] Feature

## Expected Behavior
When I add dropdown as described below I expect title to be displayed and when I change value the new value to be seen.
Also there should be "value" attribute which I can use to pre-select the value from the items if the value exists in the list.

## Actual Behavior

Title is not displayed and whenever I change item only "1" is displayed.

`<axa-dropdown  title="Please select" native items='[{"value":1,"name":1},{"value":2,"name":2},{"value":3,"name":3},{"value":4,"name":4},{"value":5,"name":5},{"value":6,"name":6},{"value":7,"name":7},{"value":8,"name":8},{"value":9,"name":9},{"value":10,"name":10},{"value":11,"name":11},{"value":12,"name":12},{"value":13,"name":13},{"value":14,"name":14},{"value":15,"name":15},{"value":16,"name":16},{"value":17,"name":17},{"value":18,"name":18},{"value":19,"name":19},{"value":20,"name":20},{"value":21,"name":21},{"value":22,"name":22},{"value":23,"name":23},{"value":24,"name":24},{"value":25,"name":25},{"value":26,"name":26},{"value":27,"name":27},{"value":28,"name":28},{"value":29,"name":29},{"value":30,"name":30},{"value":31,"name":31}]'>
</axa-dropdown>`

nakovskijosif:
Fixes #833.

 - Fixed issue where numbers in list were not compared to string value.
- Added title to native dropdown which will be displayed only when there is no selected item.
- If there is a value set to the dropdown which has no corresponding value in the item list, title will be displayed if set.

## Type of change

 # Checklist:
 
### DIFF
--- a/src/components/m-dropdown/_preview.html
+++ b/src/components/m-dropdown/_preview.html
@@ -76,3 +76,6 @@ items='[
     {"name": "The longer Item than the box", "value": "The longer Item than the box", "url": "#", "isSelected": false }
   ]'
 ></axa-dropdown>
+<h2>Native Dropdown With Title</h2>
+<axa-dropdown native title="Please select" items='[{"value":1,"name":1},{"value":"","name":2},{"value":3,"name":3},{"value":4,"name":4},{"value":5,"name":5},{"value":6,"name":6},{"value":7,"name":7},{"value":8,"name":8},{"value":9,"name":9},{"value":10,"name":10},{"value":11,"name":11},{"value":12,"name":12},{"value":13,"name":13},{"value":14,"name":14},{"value":15,"name":15},{"value":16,"name":16},{"value":17,"name":17},{"value":18,"name":18},{"value":19,"name":19},{"value":20,"name":20},{"value":21,"name":21},{"value":22,"name":22},{"value":23,"name":23},{"value":24,"name":24},{"value":25,"name":25},{"value":26,"name":26},{"value":27,"name":27},{"value":28,"name":28},{"value":29,"name":29},{"value":30,"name":30},{"value":31,"name":31}]'>
+</axa-dropdown>
--- a/src/components/m-dropdown/_template.js
+++ b/src/components/m-dropdown/_template.js
@@ -3,23 +3,18 @@ import raw from 'nanohtml/raw';
 import classnames from 'classnames';
 
 const arrowIcon = '<axa-icon size="sm" icon="angle-bracket-down" classes="m-dropdown__icon"></axa-icon>';
-const getSelectedItem = (items) => {
-  if (!items || !Array.isArray(items) || !items.length) {
-    return null;
-  }
-  return items.filter(item => item.isSelected)[0];
-};
 
 export default ({
   native,
   title,
   items = [],
   size,
+  selectedItem,
 }) => html`
 <article class="m-dropdown__wrap">
       <div class="${classnames('m-dropdown__select-wrap', { [`m-dropdown__select-wrap--${size}`]: size }, 'm-dropdown__list--native')} ${native ? ' m-dropdown__list--native--only' : ''}">
         <select class="${classnames('m-dropdown__select', 'js-dropdown__native-select', { [`m-dropdown__select--${size}`]: size })}">
-        ${title && html`<option ${getSelectedItem(items) == null ? 'selected' : ''}  value="" hidden disabled>${title}</option>`}
+        ${title && html`<option ${selectedItem === null ? 'selected' : ''}  value="" hidden disabled>${title}</option>`}
         ${Array.isArray(items) && items.map(({ name, value, isSelected }) =>
           html`<option value="${value}" ${isSelected ? 'selected' : ''}>${name}</option>`)
           }
--- a/src/components/m-dropdown/_template.js
+++ b/src/components/m-dropdown/_template.js
@@ -3,6 +3,12 @@ import raw from 'nanohtml/raw';
 import classnames from 'classnames';
 
 const arrowIcon = '<axa-icon size="sm" icon="angle-bracket-down" classes="m-dropdown__icon"></axa-icon>';
+const getSelectedItem = (items) => {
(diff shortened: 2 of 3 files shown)


src/components/m-dropdown/_template.js
```diff
@@ -3,6 +3,12 @@ import raw from 'nanohtml/raw';
 import classnames from 'classnames';
 
 const arrowIcon = '<axa-icon size="sm" icon="angle-bracket-down" classes="m-dropdown__icon"></axa-icon>';
+const getSelectedItem = (items) => {
+  if (!items || !Array.isArray(items) || !items.length) {
+    return null;
+  }
+  return items.filter(item => item.isSelected)[0];
+};
```
jackblackCH: Generally, we should keep code "together" in the controller (index.js) instead of moving it into the view (like the `getSelectedItem` function). Now parts of code are in the template and others are in the index.js. A view should only act as view and should not prepare nor manipulate data ( and If we had tests, it would even be harder to test in the view).

In this specific example, The `selectedItem` is already existing on line 160 in index.js, it may be enough to pass this prop down to the template and access it from there, no need to do figure out the selected item with another helper `(getSElectedItem)` again.

src/components/m-dropdown/_template.js
```diff
@@ -3,6 +3,12 @@ import raw from 'nanohtml/raw';
 import classnames from 'classnames';
 
 const arrowIcon = '<axa-icon size="sm" icon="angle-bracket-down" classes="m-dropdown__icon"></axa-icon>';
+const getSelectedItem = (items) => {
+  if (!items || !Array.isArray(items) || !items.length) {
+    return null;
+  }
+  return items.filter(item => item.isSelected)[0];
```
MKaHead: if `return items.filter(item => item.isSelected)[0]` is undefined - I would recommend to return null
like that `return items.filter(item => item.isSelected)[0] || null` or a other implementation.
because in the render template you should do a  save type check like `getSelectedItem(items) === null`

src/components/m-dropdown/_template.js
nakovskijosif: I agree. I'll update as soon as I can.

src/components/m-dropdown/_preview.html
```diff
@@ -75,4 +75,7 @@ <h2>Native Dropdown Long list</h2>
     {"name": "The longer Item than the box", "value": "The longer Item than the box", "url": "#", "isSelected": false },
     {"name": "The longer Item than the box", "value": "The longer Item than the box", "url": "#", "isSelected": false }
   ]'
-></axa-dropdown>
\ No newline at end of file
+></axa-dropdown>
+<h2>Native Dropdown With Title</h2>
+<axa-dropdown native title="Please select" items='[{"value":1,"name":1},{"value":"","name":2},{"value":3,"name":3},{"value":4,"name":4},{"value":5,"name":5},{"value":6,"name":6},{"value":7,"name":7},{"value":8,"name":8},{"value":9,"name":9},{"value":10,"name":10},{"value":11,"name":11},{"value":12,"name":12},{"value":13,"name":13},{"value":14,"name":14},{"value":15,"name":15},{"value":16,"name":16},{"value":17,"name":17},{"value":18,"name":18},{"value":19,"name":19},{"value":20,"name":20},{"value":21,"name":21},{"value":22,"name":22},{"value":23,"name":23},{"value":24,"name":24},{"value":25,"name":25},{"value":26,"name":26},{"value":27,"name":27},{"value":28,"name":28},{"value":29,"name":29},{"value":30,"name":30},{"value":31,"name":31}]'>
```
MKaHead: If `value` and `name` of a `item` can be number and string, then the PropTypes should be updated like that:
`items: PropTypes.arrayOf(PropTypes.shape({`
`     name: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),`
`     value: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),`
`     ...`

nakovskijosif: Done

src/components/m-dropdown/_template.js
nakovskijosif: I removed this helper and added selectedItem as attribute.

src/components/m-dropdown/_template.js
nakovskijosif: I removed this helper, but I return null in the duplicate helper and do type check.

src/components/m-dropdown/index.js
```diff
@@ -214,6 +215,14 @@ class AXADropdown extends BaseComponentGlobal {
   get native() {
     return this.getAttribute('native');
   }
+
+  set selectedItem(value) {
+    this.setAttribute('selected-item', JSON.stringify(value));
+  }
+
+  get selectedItem() {
+    return JSON.parse(this.getAttribute('selected-item'));
```
LucaMele: why setting this as prop? can it not just e a function or helper inside `index.js`?

nakovskijosif: Well, I need info in `template.js` whether there is selected item or not. I had a function for it in `template.js` but it was suggested that I remove it and get that info from the `index.js`. 
I'm not sure which is the best way to do it, and I would appreciate your suggestion.

LucaMele: i think is meant to just pass it down instead of calculating inside the template. Now we have a attribute that always changes and triggers a new-rendering. Can the funciton not just live in `index.js` and pass the filtered value with `get item()`? then in template u access it easly via wcNode

LucaMele: if its not clear, we can have a quick call

nakovskijosif: Thanks @LucaMele for the help. Requested changes implemented.

Issue #818, opened 2019-02-12
m-link arrow lost dimensions


AndyOGo:
## Are you reporting a bug or a feature?

 - [x] Bug
 - [ ] Feature

## Expected Behavior

Link component icon classes should properly style dimension, without adjoining classes:
https://github.com/axa-ch/patterns-library/blame/abf3d7529120e6c299b4e671e89eaccc35afd231/src/components/m-link/index.scss#L39-L43

## Actual Behavior

Needs to be adjoined with `.a-icon__svg` to make the selector work.

## Steps to Reproduce the Problem

1. import the CSS from `@import "@axa-ch/patterns-library/src/components/m-link/mixins/link";`
2. and build your own  link component using `.m-link__arrow` for the `<svg>` tag

<img width="734" alt="screenshot 2019-02-12 at 12 27 22" src="https://user-images.githubusercontent.com/914443/52637863-d4379b00-2ed0-11e9-9bcd-1b18e9b58cdd.png">

## Quick Workaround

Add `.a-icon__svg` class to your `<svg>` tag, and selectors match again 🎉

AndyOGo:
introduced with this  change:
https://github.com/axa-ch/patterns-library/blob/e07e99e12c1810592aff803d3fbc5e4bd3884222/src/components/m-link/index.scss#L40-L45

in relating commit https://github.com/axa-ch/patterns-library/commit/c39c298515fdef9e0eccb71e5e6617738c33bb6b

jackblackCH:
I try to reproduce, please send the exact snippet to reproduce it (like in the screenshot)

AndyOGo:
@jackblackCH
I dont' now if you have access rights for aletheia.

Meanwhile I quickfixed it myself by adding `.a-icon__svg` class at this commit
https://github.com/axa-ch/aletheia/pull/377/commits/879433e8b688c35cd8e336eac1b7f0c0cf0b9037

Within this PR:
https://github.com/axa-ch/aletheia/pull/377

LucaMele:
Closing as V2 is released. If issue is still relevant in v2, please re-open

AndyOGo:
This is still an issue for all plib v1 users

LucaMele:
Migrate to v2, we cannot fix this in v1

Issue #814, opened 2019-02-11
M-Link icon broken


LucaMele:
Since datepicker v2, icons broken: #809  
<img width="730" alt="bildschirmfoto 2019-02-11 um 16 16 00" src="https://user-images.githubusercontent.com/275235/52572613-7d6e8a80-2e18-11e9-8436-a89c3efce884.png">

PR #840, opened 2019-02-26
Fix broken icons in button, fixes #836

### LINKED ISSUE #836: Broken default m-button divider line and icon
- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed

 - [x] Bug
 - [ ] Feature

## Actual Behavior

Currently the default button with icon is broken. 
- The divider line stop 4/5 down
- The icon is smaller than it should be and placed towards the top.

Can be seen even in the showcase of components [m-button](https://axa-ch.github.io/patterns-library/components/m-button/index.html)

screenshot 2019-02-26 at 10 23 38

## Specifications

 - Browser: Chrome, IE

jackblackCH:
Button sizes are overridden by the main svg class without this #fix
Fixes #836

### DIFF
--- a/src/components/m-button/index.scss
+++ b/src/components/m-button/index.scss
@@ -10,7 +10,11 @@
   @include button--color();
 }
 
-.m-button__icon {
+// a-icon__svg is needed as css ordering is not loaded in the right order
+// I.e sizes in m-button__icon are overridden without this specifier.
+// Relates and fixes gh #836
+.m-button__icon,
+.m-button__icon.a-icon__svg {
   @include button__icon();
 }
 


src/components/m-button/index.scss
```diff
@@ -10,7 +10,11 @@
   @include button--color();
 }
 
-.m-button__icon {
+// a-icon__svg is needed as css ordering is not loaded in the right order
+// I.e sizes in m-button__icon are overridden without this specifier.
+// Relates and fixes gh #836
+.m-button__icon,
+.m-button__icon.a-icon__svg {
```
AndyOGo: @LucaMele 
@jackblackCH 
Indeed, order is an issue, actually dependencies between components aren't expressed yet - tracked here #723 and a PR here #724
Unfortunately this workaround, does not fix the underlying issue, but rather creates additional bugs - reported so far: #836, #818 and #814 
Also all CSS-only consumers have to deal with it

AndyOGo: Are there any other components affected?

jackblackCH: So far I think, no.

LucaMele: Agree with @AndyOGo, this way is not optimal. We will fix it in V2. For v1, i would say lets go with the current change and merge it.

AndyOGo: Unfortunately I found 2 more related issues 😞 
#841 #842

Issue #848, opened 2019-03-04
Modal Jittering after load


AndyOGo:
## Are you reporting a bug or a feature?

 - [x] Bug
 - [ ] Feature

## Expected Behavior

Modal should never be visible until it is opened.

## Actual Behavior

Modal is visible at the beginning and jitters, then invisible.

## Steps to Reproduce the Problem

1. https://axa-ch.github.io/patterns-library/components/o-modal/index.html

![modal-jittering](https://user-images.githubusercontent.com/914443/53737334-8806c780-3e8c-11e9-9c71-ba02185fe8f8.gif)

## Specifications

 - Browser: Latest Chrome

AndyOGo:
Guess is also related to FOUC #365 🤔

raphaellueckl:
Yes, this is related and we were aware of it. This won't be a problem for V2 anymore, but I will provide a fix for this, as soon as I am done with the current work. Thanks for the feedback! This should definitely be fixed, no matter of versions.

raphaellueckl:
#854

AndyOGo:
@raphaellueckl 
Thanks for the update.

FYI: You can close issues via pull requests, by `fixes #848` (you can close as many as you wish).
More info: https://github.blog/2013-05-14-closing-issues-via-pull-requests/

Issue #853, opened 2019-03-05
Modal content not optimised for mobile viewports


AndyOGo:
## Are you reporting a bug or a feature?

 - [x] Bug
 - [ ] Feature

## Expected Behavior

Content of the modal should be displayed nicely, like centered, buttons not glued together, etc. please see the pictures.

## Actual Behavior

Buttons glued together, some content may looks nicer if centered, etc.

## Steps to Reproduce the Problem

1. Take your iPhone
2. https://axa-ch.github.io/patterns-library/components/o-modal/index.html
3. click open button

![image from ios 1](https://user-images.githubusercontent.com/914443/53790667-cb624400-3f27-11e9-9c2c-87a03b67ccce.png)

## Specifications

 - Browser: iOS Webkit at iPhone

raphaellueckl:
Thanks for the feedback, but the content is up to the guy who implements it. That was a team internal decision. We only make the modal responsive, but the content is not our responsibility.

AndyOGo:
@raphaellueckl 
I understand.

But shouldn't the example show how to do it nicely?
@LucaMele Please reopen, IMHO the style-guide should not give examples that show unwanted practices.

LucaMele:
@raphaellueckl its up to you if you want to fix the example or not. As its true a nice example is better. Its up to you

raphaellueckl:
#854

LucaMele:
Closing as V2 is released. If issue is still relevant in v2, please re-open

PR #854, opened 2019-03-06
Bugfix/axa modal mobile display and fouc

raphaellueckl:
Fixes # (issue).

 - No more flash of unstyled content
 - Fixed the preview on mobile phones

## Type of change

- [x ] Bug fix (non-breaking change which fixes an issue)
 
Fixes: #848 #853

### DIFF
--- a/src/components/o-modal/_preview.html
+++ b/src/components/o-modal/_preview.html
@@ -66,7 +66,7 @@
     color: white;
   }
 
-  axa-col {
+  .axa-modal-preview axa-col {
     padding: 10px;
   }
 </style>
--- a/src/components/o-modal/_preview.html
+++ b/src/components/o-modal/_preview.html
@@ -9,23 +9,22 @@
   <axa-container>
     <axa-row>
       <axa-col>
-        <axa-icon icon="rocking_chair" icon-class="huge"> </axa-icon>
-        <axa-icon icon="broken_glass" icon-class="huge"> </axa-icon>
+        <div>
+          <axa-icon icon="rocking_chair" icon-class="huge"> </axa-icon>
+        </div>
+        <axa-button data-modal-close>Nur Hausrat</axa-button>
       </axa-col>
       <axa-col>
-        <axa-icon icon="rocking_chair" icon-class="huge"> </axa-icon>
+        <div>
+          <axa-icon icon="rocking_chair" icon-class="huge"> </axa-icon>
+          <axa-icon icon="broken_glass" icon-class="huge"> </axa-icon>
+        </div>
+        <axa-button data-modal-close>Hausrat und Privathaftpflicht</axa-button>
+        <div>
+          <axa-icon class="green-icon" icon="check_circle_outline"></axa-icon>
+          Schützt Sie zusätzlich gegen Forderungen anderer.
+        </div>
       </axa-col>
-    </axa-row>
-    <axa-row>
-      <axa-col><axa-button data-modal-close>Hausrat und Privathaftpflicht</axa-button></axa-col>
-      <axa-col><axa-button data-modal-close>Nur Hausrat</axa-button></axa-col>
-    </axa-row>
-    <axa-row>
-      <axa-col>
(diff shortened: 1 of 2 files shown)


raphaellueckl:
Please leave your comments, because I guess some stuff could be solved in multiple ways and the FOUC problem is more or less a quick fix. :)

(Sorry that I opened one PR for two different things)

src/components/o-modal/_preview.html
```diff
@@ -1,31 +1,31 @@
 <button type="button" onclick="openModal()">Open modal</button>
 
-<axa-modal class="axa-modal-preview">
+<!-- The style attribute is necessary to avoid the FOUC problem -->
+<axa-modal class="axa-modal-preview" style="display:none;">
```
LucaMele: I would use a modifier instead of s style attr `axa-modal-preview--hidden` for example

src/components/o-modal/index.js
```diff
@@ -48,6 +48,9 @@ class AXAModal extends BaseComponentGlobal {
   connectedCallback() {
     super.connectedCallback();
 
+    // FOUC workaround
+    this.removeAttribute('style');
```
LucaMele: and here classList.remove()

src/components/o-modal/_preview.html
AndyOGo: Be aware that only inline style will be available before the component has been downloaded.
If you add a modifier within the `<axa-modal>` CSS, this CSS will only be available after component download and will not be able to fix FOUC, because it is loaded too late

src/components/o-modal/index.js
AndyOGo: This only works if said modifier is externel to `<axa-modal>`'s CSS and loaded within the `<head>` before anything else 💣

src/components/o-modal/_preview.html
```diff
@@ -76,4 +65,8 @@ <h2>Prämie berechnen</h2>
     fill: green;
     color: white;
   }
+
+  axa-col {
```
AndyOGo: Please use a specific class here.

src/components/o-modal/index.js
AndyOGo: better to remove just the specific `display` style property by using these APIs:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/style
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Properties_Reference

src/components/o-modal/_preview.html
raphaellueckl: Ah shit, how did I miss that...

raphaellueckl:
All comments are handled, please review again (I went with the inline style @LucaMele, feel free to express your opinion).

src/components/o-modal/_preview.html
LucaMele: ahhh yes good point sorry

src/components/o-modal/index.js
LucaMele: agree with andy

LucaMele: @raphaellueckl please can u do as @AndyOGo said here (`wcNode.style.display = 'block';`)?

raphaellueckl: @LucaMele @AndyOGo Done, please have a look again.

src/components/o-modal/index.js
```diff
@@ -48,6 +48,9 @@ class AXAModal extends BaseComponentGlobal {
   connectedCallback() {
     super.connectedCallback();
 
+    // FOUC workaround
+    this.style.display = 'block';
```
AndyOGo: @raphaellueckl 
@LucaMele 
That works, though it trumps styles coming from a class.
Just if you the case to control `display` by a class, better to remove the inline style, like:

```js
this.style. removeProperty('display');
```

https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/removeProperty

raphaellueckl:
merci @AndyOGo

PR #855, opened 2019-03-07
Migrate Storybook v4.1 to v5

MKaHead:
Storybook v4.1 migration to 5.
- some part of themeing in v5 is strange, but not important for us yet.
- themeing get fixed soon by Storybook.

btw info:
I had to kill package.lock.json to update it properly.


LucaMele:
yeah node 8 has issues with the lock file

PR #856, opened 2019-03-07
Feature/finalise button

MKaHead:
Not really final, because we need more core feature ready. But l like to merge the actual state in develop-v2, because it's more stable than before.


MKaHead:
fixed svg stuff is only a mock for the moment

src/components/molecules/button/index.js
```diff
@@ -2,24 +2,40 @@ import { LitElement, html, css, unsafeCSS } from 'lit-element';
 import { classMap } from 'lit-html/directives/class-map';
 import buttonCSS from './index.scss';
 
+const icon = () => html`
```
LucaMele: why not `import` our new icon? and is it ok for @jojackpot to have a different arrow?

src/components/molecules/button/index.js
```diff
@@ -2,24 +2,40 @@ import { LitElement, html, css, unsafeCSS } from 'lit-element';
...
-      ghost: { type: Boolean },
+      secondary: { type: Boolean },
+      large: { type: Boolean },
+      inverted: { type: Boolean },
+      cta: { type: Boolean },
       motion: { type: Boolean },
       disabled: { type: Boolean },
 
-      onClick: { type: Function },
+      onClick: { type: String },
```
LucaMele: on line 48 we check for type function but here is string.. is this intended as not finished code or a n error?

MKaHead:
- fix button property
- change molecule to atom
- place the svg mock somewhere else but couldn't fix the initial problem in this branch. The svg (build, example...) stuff should be fixed in a different branch.

jackblackCH:
LGTM

PR #863, opened 2019-03-12
Feature/dropdown v2

jackblackCH:
a working version of a refactored component based on lit-element.
Let's add it for further development and as base discussion for

- unit tests
- typography and other materials
- other helpers
- icons

### DIFF
--- a/src/components/20-molecules/dropdown/index.js
+++ b/src/components/20-molecules/dropdown/index.js
@@ -37,11 +37,12 @@ class AXADropdown extends LitElement {
 
   connectedCallback() {
     super.connectedCallback();
+    window.axaComponents = window.axaComponents || {};
     this.isOpen = false;
     this.selectedItem = this.items.filter(item => item.isSelected)[0] || null;
     this.title = this.selectedItem.name;
 
-    window.openDropdownInstance = false;
+    window.axaComponents.openDropdownInstance = false;
     window.addEventListener('keydown', e => this.handleWindowKeyDown(e));
     window.addEventListener('click', e => this.handleWindowClick(e));
   }
@@ -122,8 +123,8 @@ class AXADropdown extends LitElement {
   }
 
   closeOpenDropdowns() {
-    if (window.openDropdownInstance && this !== window.openDropdownInstance) {
-      this.closeDropdown(window.openDropdownInstance);
+    if (window.axaComponents.openDropdownInstance && this !== window.axaComponents.openDropdownInstance) {
+      this.closeDropdown(window.axaComponents.openDropdownInstance);
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/components/20-molecules/dropdown/index.js
```diff
@@ -0,0 +1,224 @@
...
+    window.addEventListener('resize', debounce(() => this.handleViewportCheck(this.dropdown), 250));
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+    this.isOpen = false;
+    this.selectedItem = this.items.filter(item => item.isSelected)[0] || null;
+    this.title = this.selectedItem.name;
+
+    window.openDropdownInstance = false;
```
LucaMele: what if u have more then 1 dropdown on a page?

LucaMele: plus we should use one window object that is genreic like: `__configs_plib`.

src/components/20-molecules/dropdown/index.scss
```diff
@@ -0,0 +1,238 @@
+$dropdown-color: $color-prim-gray-mine-shaft !default;
+$dropdown-color--hover: $color-axa-blue !default;
+$dropdown-border-width: 1px;
+$dropdown-border-style: $dropdown-border-width solid $color-prim-gray-silver;
+$dropdown-height: 40px;
+
+@mixin normalize-native-select-styles() {
+  -moz-appearance: none;
```
LucaMele: here we should relay on autoprefixer

src/components/20-molecules/dropdown/index.js
LucaMele: example, every component does this in the constructor: `window._config_plig = window._config_plig || {};`

LucaMele: once SSR is a topic, we also have to add something like: 

`const win = global || window`

jackblackCH: sure but add it when SRR is a topic. (over engineering)

LucaMele: Yepp thats why i said: "once SSR is a topic"

src/components/20-molecules/dropdown/index.scss
jackblackCH: done.

src/components/20-molecules/dropdown/index.js
jackblackCH: This 

> what if u have more then 1 dropdown on a page?

This exactly why we save the previous instance so we can close it on the next click.

jackblackCH: > example, every component does this in the constructor: `window._config_plig = window._config_plig || {};`

I agree, scope axa components within a sub object makes sense, something like `window.axaComponents` or window.axa.components` since components basically independent we should discuss the pros and cons.

LucaMele:
LGTM

PR #868, opened 2019-03-18
beginnings of Architecture document

markus-walther:
Luca,

 is this going into the right direction?

Cheers, Markus


jackblackCH:
This document is 👍 🥇

PR #869, opened 2019-03-19
Introduce axa footer small

raphaellueckl:

 - Adds a customizable footer

## Type of change

### DIFF
--- a/src/components/20-molecules/footer-small/index.js
+++ b/src/components/20-molecules/footer-small/index.js
@@ -1,6 +1,6 @@
 import { LitElement, html, css, unsafeCSS } from 'lit-element';
-import footerSmallCSS from './index.scss';
 import { repeat } from 'lit-html/directives/repeat';
+import footerSmallCSS from './index.scss';
 
 class AXAFooterSmall extends LitElement {
   static tagName = 'axa-footer-small';
--- a/src/components/20-molecules/footer-small/index.js
+++ b/src/components/20-molecules/footer-small/index.js
@@ -23,10 +23,6 @@ class AXAFooterSmall extends LitElement {
     this.copyrightText = '';
   }
 
-  formatText(text, arrIndex, arrLength) {
-    return `${text}${arrIndex !== arrLength - 1 ? ' | ' : ''}`;
-  }
-
   render() {
     return html`
       <article class="m-footer-small">
--- a/src/components/20-molecules/footer-small/index.js
+++ b/src/components/20-molecules/footer-small/index.js
@@ -1,5 +1,6 @@
 import { LitElement, html, css, unsafeCSS } from 'lit-element';
 import footerSmallCSS from './index.scss';
+import { repeat } from 'lit-html/directives/repeat';
 
 class AXAFooterSmall extends LitElement {
   static tagName = 'axa-footer-small';
@@ -45,9 +46,16 @@ class AXAFooterSmall extends LitElement {
 
     return html`
       <article class="m-footer-small">
-        <div class="m-footer-small__side-aligned">
-          ${languageHtml}
-        </div>
+        <ul class="m-footer-small__side-aligned">
+          ${repeat(
+            this.languageLinks,
+            (languageItem, index) => html`
+              <li class="m-footer-small__list">
+                <a class="m-footer-small__link" href="${languageItem.link}">${languageItem.text}</a>
+              </li>
+            `
+          )}
+        </ul>
 
         <div class="m-footer-small__side-aligned">
           <div class="m-footer-small__disclaimer">
--- a/src/components/20-molecules/footer-small/index.js
+++ b/src/components/20-molecules/footer-small/index.js
@@ -1,0 +1,67 @@
+import { LitElement, html, css, unsafeCSS } from 'lit-element';
+import footerSmallCSS from './index.scss';
+
+class AXAFooterSmall extends LitElement {
+  static tagName = 'axa-footer-small';
+  static styles = css`
+    ${unsafeCSS(footerSmallCSS)}
+  `;
+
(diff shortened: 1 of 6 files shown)


src/components/20-molecules/footer-small/index.js
```diff
@@ -0,0 +1,67 @@
...
+
+  constructor() {
+    super();
+    this.languageLinks = [];
+    this.disclaimerLinks = [];
+    this.copyrightText = '';
+  }
+
+  formatText(text, arrIndex, arrLength) {
+    return `${text}${arrIndex !== arrLength - 1 ? ' | ' : ''}`;
```
LucaMele: @raphaellueckl please use array method join instead. Reason: better readability

https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Array/join

https://jsbin.com/kemifiwuhe/edit?js,console

LucaMele: And better would be use a pseudo element for pipeing as the pipe is a visual or design element  and not part of text.

```
.m-footer-small__link + .m-footer-small__link::before {
    content: " | ";
}
```

LucaMele: Plus relying on the index of map is dangerous as you make 2 function hardly coupled.

src/components/20-molecules/footer-small/index.js
```diff
@@ -0,0 +1,67 @@
...
+        html`
+          <a class="m-footer-small__link" href="${disclaimer.link}"
+            >${this.formatText(disclaimer.text, index, this.disclaimerLinks.length)}</a
+          >
+        `
+    );
+
+    return html`
+      <article class="m-footer-small">
+        <div class="m-footer-small__side-aligned">
```
LucaMele: for me it looks like this should be a `ul` and every `a` tag a `axa-link` component wrapped in a `li`.

src/components/20-molecules/footer-small/index.js
```diff
@@ -0,0 +1,67 @@
...
+          ${languageHtml}
+        </div>
+
+        <div class="m-footer-small__side-aligned">
+          <div class="m-footer-small__disclaimer">
+            <span class="m-footer-small__disclaimer--group-right">
```
LucaMele: dtto: `ul `li` and `axa-link` here instead

src/components/20-molecules/footer-small/index.js
```diff
@@ -0,0 +1,67 @@
...
+              ${disclaimerHtml}
+            </span>
+            <span footer-small__disclaimer--group-right>${this.copyrightText}</span>
```
LucaMele: typo in class name (class attribute missing)

LucaMele:
As a recap, for me we should never merge unfinished code. To me, is not design compliant if `axa-link` is missing. Now we have a link that looks like a link, but in reality isn't. Same as with the datepicker where we migrated also axa-dropdown, here we should in this PR migrate axa-link as without it, the axa-footer is not finished. Not doing this can lead to broken windows

src/components/20-molecules/footer-small/index.js
raphaellueckl: @LucaMele I learned about this method with css this weekend :D Some guy made a compilation of css tricks, I didn't know that before, but I will go for that one, thanks!

src/components/20-molecules/footer-small/index.js
```diff
@@ -1,5 +1,6 @@
 import { LitElement, html, css, unsafeCSS } from 'lit-element';
 import footerSmallCSS from './index.scss';
+import { repeat } from 'lit-html/directives/repeat';
```
LucaMele: WOW nice!

src/components/20-molecules/footer-small/index.scss
```diff
@@ -21,13 +19,32 @@
     justify-content: space-between;
   }
 
-  &__side-aligned {
-    padding-top: 5px;
-    padding-bottom: 5px;
+  span {
```
LucaMele: no globals. ever

raphaellueckl: I always use those for testing and then forget to clean up... thx.

LucaMele: and i will always be there and checking ahahahaha

src/components/20-molecules/footer-small/index.js
```diff
@@ -0,0 +1,60 @@
+import { LitElement, html, css, unsafeCSS } from 'lit-element';
+import footerSmallCSS from './index.scss';
```
jackblackCH: We can omit file extensions `.scss`

src/components/20-molecules/footer-small/index.scss
```diff
@@ -0,0 +1,60 @@
...
+  @include breakpoint($mediaquery-md-up) {
+    flex-direction: row;
+    justify-content: space-between;
+  }
+
+  &__list {
+    // ### Reset browser styles for ul
+    margin: 0;
+    padding: 0;
+    text-indent: 0;
```
jackblackCH: is `text-indent` needed?

raphaellueckl: dunno, i copied this from a random guy on github that claimed that this would reset an ul.

src/components/20-molecules/footer-small/index.js
raphaellueckl: That only works in case of "js" files afaik. Still, I tried it out and it does not apply the scss then.

raphaellueckl:
All comments handled, r4r

jackblackCH:
LGTM

PR #871, opened 2019-03-19
Bugfix/lint prettier fix

jackblackCH:
Fixes conflicts between formatting- and linting rules. This is a known side effect while using prettier and eslint out of the box.

Since prettier is here for formatting only and since this is depending only on the line length, some eslint rules like comma dangles and parens rules are in conflict with each other. Basically, it does not make sense to lint them, since it can change immediately by the line length.

That's why the community brought us the "eslint-prettier-plugin", that disables all rules which are taken over by prettier. (Code formatting).

https://github.com/prettier/eslint-config-prettier

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -107,6 +107,7 @@
     "HTMLImports": true,
     "XMLHttpRequest": true,
     "history": true,
+    "CustomEvent": true,
     "MutationObserver": true,
     "customElements": true,
     "document": true,
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,5 +1,5 @@
 {
-  "extends": "airbnb",
+  "extends": ["airbnb", "prettier"],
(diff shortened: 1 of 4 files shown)


.eslintrc
```diff
@@ -28,7 +28,21 @@
...
+    "import/no-extraneous-dependencies": [
+      "error",
+      {
+        "devDependencies": [
+          "scripts/**",
+          "rollup.config.js",
+          "**/_demo*.js",
+          "**/story.js",
+          "**/_story*.js",
+          "**/*.story.js",
```
jackblackCH: We have to version for the naming for 'stories'. This is a temp fix that works for all the existing versions. This will be cleaned up in a separate PR.

LucaMele: agree

PR #874, opened 2019-03-20
Feature/build improvements

MKaHead:
- Improve build and fix icon build
- button improvements
- update storybook to 5.0.3


src/components/10-atoms/button/index.js
```diff
@@ -44,7 +44,7 @@ class AXAButton extends LitElement {
       'a-button--large': this.large,
       'a-button--secondary': this.secondary,
       'a-button--inverted': this.inverted,
-      'a-button--motion': this.motion,
+      'a-button--motion': !this.static,
```
LucaMele: I would suggest to use still or similar, not a word that has a meaning in Javascript (like static has). Users cant get confused. If is a NON JS word, then maybe is clearer what we want to achive. is not a show stopper thought so ur desicion :)

src/components/10-atoms/button/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "@axa-ch/button",
+  "version": "1.0.0-beta.1",
```
LucaMele: not sure if is better to start with `0.0.0-beta.1` instead

src/components/10-atoms/button/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "description": "> TODO: description",
+  "author": "Pattern Warriors",
+  "homepage": "https://github.com/axa-ch/patterns-library/tree/master/src/components/10-atoms/button#readme",
+  "license": "ISC",
```
LucaMele: Carefull here, pelase add See Readme and then in Readme add "AXA Versicherungen AG"

LucaMele:
nice PR :)

PR #876, opened 2019-03-20
Removed rule that is no longer valid

raphaellueckl:
Please only merge if all of the patterns lib team accepted.

We changed the rules so that they are able to work together with prettier.


raphaellueckl:
> Why not valid? we decided to change stylelint and keep eslint in the last meeting. Adding prettier didnt change the linting stuff

AFAIK we removed a rule, but I don't know which one it was. But anyways, we will soon have a fundamental discussion about this stuff, so we can also wait with this PR.

PR #886, opened 2019-03-25
Feature/add dist build

LucaMele:
Adds a IIFE bundle parallel to every lib folder. Only for legacy. No common chunks

https://jsbin.com/ruxubequma/2/edit?html,js,output


jackblackCH:
Might be offtopic but npm scripts look a bit polluted with rimraf and cleanup scripts. is it possible to put them away from the "public" api

LucaMele:
agree with marc

rollup.config.js
```diff
@@ -20,45 +22,102 @@ const babelOptions = require('./.storybook/.babelrc'); // get the babelrc file
...
+    babel({
+      presets: babelOptions.presets,
+      plugins: [
+        ...babelOptions.plugins,
+        '@babel/plugin-transform-runtime'
+      ],
+      babelrc: false,
+      exclude: [
+        'node_modules/@skatejs/val/**',
+        'node_modules/@babel/**',
```
codepleb: Why are only those two node_modules folders ignored?

LucaMele: @codepleb  Cause they dont provide correctly UMD named exports and generate a warning in rollup build. The warning says that those modules are ignored and threated as external. This settings is only here to suppress that warning

raphaellueckl:
lgtm

PR #887, opened 2019-03-27
Fix demo content

LucaMele:
Fixes Layout issue in preview and modal and adapt window height to content height

### DIFF
--- a/src/components/o-modal/_preview.html
+++ b/src/components/o-modal/_preview.html
@@ -2,31 +2,36 @@
 
 <!-- The style attribute is necessary to avoid the FOUC problem -->
 <axa-modal class="axa-modal-preview" style="display:none;">
-  <h2>Prämie berechnen</h2>
-  <p><strong>Mehr als 80%</strong> unserer Kunden kombinieren die Hausratversicherung mit einer Privathaftpflichtversicherung.</p>
-  <p>Was möchten Sie?</p>
-
-  <axa-container>
-    <axa-row>
-      <axa-col>
-        <div>
-          <axa-icon icon="rocking_chair" icon-class="huge"> </axa-icon>
-        </div>
-        <axa-button data-modal-close>Nur Hausrat</axa-button>
-      </axa-col>
-      <axa-col>
-        <div>
-          <axa-icon icon="rocking_chair" icon-class="huge"> </axa-icon>
-          <axa-icon icon="broken_glass" icon-class="huge"> </axa-icon>
-        </div>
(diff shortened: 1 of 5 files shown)


src/components/o-modal/_preview.html
```diff
@@ -37,36 +42,106 @@ <h2>Prämie berechnen</h2>
...
+        right: 20px;
+
+    }
+
+    .axa-modal-preview .m-modal__content {
+        position: relative;
+
+    }
+
+    .axa-modal-preview h2 {
```
raphaellueckl: no globals, right?

src/components/o-modal/index.js
```diff
@@ -71,6 +73,33 @@ class AXAModal extends BaseComponentGlobal {
...
+    if (offsetHeight && vh) {
+      const marginTopBottom = (vh - offsetHeight) / 2;
+
+      if (marginTopBottom > 0) {
+        contentElementStyle.marginTop = `${marginTopBottom}px`;
+      } else {
+        contentElementStyle.marginTop = '0px';
+      }
+    }
+  }
```
raphaellueckl: Why styling with css?

src/components/o-modal/index.scss
```diff
@@ -36,18 +37,19 @@
   @include respond-up(md) {
     margin-top: 20vh;
     width: 80vw;
-    height: 60vh;
+    height: auto;
```
raphaellueckl: isn't this the default? Doesn't look like you override anything here.

src/components/o-modal/index.scss
```diff
@@ -36,18 +37,19 @@
...
   }
 
   @include respond-up(lg) {
     margin-top: 20vh;
     width: 70vw;
-    height: 60vh;
+    height: auto;
```
raphaellueckl: isn't this the default? Doesn't look like you override anything here.

src/components/o-modal/_preview.html
LucaMele: yes right but this is not global. is scoped in axa-preview. And this is copy paste from previous :)

src/components/o-modal/index.js
LucaMele: cause is calculated

src/components/o-modal/index.scss
LucaMele: im overriding as i set height 100% on the mobile screen

src/components/o-modal/index.scss
LucaMele: im overriding as i set height 100% on the mobile screen

PR #898, opened 2019-04-02
Develop v2 how we work

markus-walther:
Hi Luca,

does this look reasonable?

--Markus


LucaMele:
@markus-walther please add V2 label always on your PRs and name your branch in the following way in the future: `feature/any-name` for changes and `bugfix/ayn-name` for operations

PR #908, opened 2019-04-08
Feature/introduce footer tests

raphaellueckl:

 - Add tests for axa footer

### DIFF
--- a/src/components/20-molecules/footer-small/index.js
+++ b/src/components/20-molecules/footer-small/index.js
@@ -10,45 +10,40 @@ class AXAFooterSmall extends LitElement {
 
   static get properties() {
     return {
-      languageitems: { type: Array },
-      disclaimeritems: { type: Array },
-      copyrighttext: { type: String },
+      languageItems: { type: Array },
+      disclaimerItems: { type: Array },
+      copyrightText: { type: String },
     };
   }
 
   constructor() {
     super();
-    this.languageitems = [];
-    this.disclaimeritems = [];
-    this.copyrighttext = '';
+    this.languageItems = [];
+    this.disclaimerItems = [];
+    this.copyrightText = '';
   }
+
+  renderFooterList = items => html`
+    <ul class="m-footer-small__list">
+      ${repeat(
+        items,
+        item => html`
+          <li class="m-footer-small__list-item">
+            <axa-link class="m-footer-small__link" href="${item.link}" color="white">${item.text}</axa-link>
+          </li>
+        `
+      )}
+    </ul>
+  `;
 
   render() {
     return html`
       <article class="m-footer-small">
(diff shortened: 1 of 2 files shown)


src/components/20-molecules/footer-small/index.js
```diff
@@ -10,45 +10,40 @@ class AXAFooterSmall extends LitElement {
...
-              languageItem => html`
-                <li class="m-footer-small__list-item">
-                  <axa-link class="m-footer-small__link" href="${languageItem.link}" color="white">${languageItem.text}</axa-link>
-                </li>
-              `
-            )}
-          </ul>
-          <div>${this.copyrighttext}</div>
+          ${this.renderFooterList(this.disclaimerItems)}
+          <div class="js-footer-small__copyright">${this.copyrightText}</div>
```
LucaMele: Invalid HTML: A div should never contain directly text

LucaMele: consider please this construct: 

```
<footer>
 <small>&copy; Copyright 2058, Example Corporation</small>
</footer>
```

src/components/20-molecules/footer-small/index.js
```diff
@@ -10,45 +10,40 @@ class AXAFooterSmall extends LitElement {
 
   static get properties() {
     return {
-      languageitems: { type: Array },
-      disclaimeritems: { type: Array },
-      copyrighttext: { type: String },
+      languageItems: { type: Array },
```
LucaMele: Having attributes as camel case in not nice. Better in lower case. Why did u change it from `languageitems ` to `languageItems `? can u still add attribute `<axa-footer languageitems="..." ></axa-footer>`? if not, please revert camel case

raphaellueckl: Yes, that's why I did it. Marco gave me this hint. This way we are compliant with the standard from html AND js. :)

src/components/20-molecules/footer-small/index.js
raphaellueckl: "The <small> tag defines smaller text (and other side comments)."

I wrapped it in a span tag, not sure if that is the nicest idea, but at least better than putting it in the div. Can you have a look again?

raphaellueckl: @LucaMele I tested it out and it appeared smaller. I surely could reset the styles, but IMHO we should not do that. <small> is meant for smaller text on a website, but since we do not write a whole website but just a component, that does not make much sense imho. We need someting to "compare" smaller text to. If we just decorate the whole footer element with <small> (because it is meant for smaller text), it is imho equivalent to making a whole website with small-tags just to make the text appear smaller. It should be "smaller compared to X" and not just "smaller than... hm... smaller than what?". 

IMHO :)

PR #910, opened 2019-04-09
Feature/axa link with real icon

raphaellueckl:

 - axa-icon for axa-link

### DIFF
--- a/src/components/10-atoms/link/index.js
+++ b/src/components/10-atoms/link/index.js
@@ -33,30 +33,41 @@ class AXALink extends LitElement {
     this.motion = false;
   }
 
-  render() {
-    const classes = {
+  connectedCallback() {
+    super.connectedCallback();
+    this.classes = classMap({
       'a-link--red': this.color === 'red',
       'a-link--blue': this.color === 'blue',
       'a-link--white': this.color === 'white',
       'a-link--bold': this.bold,
       'a-link--deco': this.deco,
       'a-link--motion': this.motion,
-    };
+    });
+  }
 
-    const showArrow = html`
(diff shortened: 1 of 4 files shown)


src/components/10-atoms/link/index.js
```diff
@@ -33,30 +33,42 @@ class AXALink extends LitElement {
     this.motion = false;
   }
 
+  connectedCallback() {
+    super.connectedCallback();
+  }
+
+  showArrow = html`
```
LucaMele: might be nice to move vars at the top

src/components/10-atoms/link/index.js
```diff
@@ -33,30 +33,42 @@ class AXALink extends LitElement {
...
-        class="a-link ${classMap(classes)}"
+        class="a-link ${classes}"
         href="${this.href}"
         target="${this.external ? '_blank' : '_top'}"
         rel="${this.external ? 'noreferrer noopener' : ''}"
       >
-        ${this.arrowLeft === true ? showArrow : ''}
+        ${this.arrowLeft
+          ? html`
+              <div class="a-link__arrow">${this.showArrow}</div>
```
LucaMele: in HTML 4 this would be invalid, in HTML 5 its ok as long as there are no click events. I still would prefer to see a `<span>` element here so that we dont have inline (a tag) -> block (div tag) -> inline elements.

But as it is semantically correct, its up to u

MKaHead: I would like to see here `span`, too.

raphaellueckl: Done, merci!

PR #912, opened 2019-04-11
Use icon in datepicker

raphaellueckl:

 - Datepicker making use of the icon materials


lerna.json
```diff
@@ -1,8 +1,5 @@
 {
-  "packages": [
-    "src/components/00-materials",
-    "src/components/10-atoms/icon"
-  ],
+  "packages": ["src/components/00-materials", "src/components/10-atoms/icon", "src/components/20-molecules/datepicker"],
```
MKaHead: Hey @raphaellueckl the datepicker v2 is not yet rdy for alpha release. Could you remove it for just now, because if we run lerna publish command the datepicker will be triggered, too.

raphaellueckl: Done

PR #914, opened 2019-04-12
Feature/readme storyook

LucaMele:
Add Addon for Markdown. 

Unfortunatly existing markdowns addons only work with `storybook/react` or `storybook/vue`: https://github.com/tuchk4/storybook-readme

Or cannt be place in a Panel like: https://www.npmjs.com/package/@storybook/addon-notes

I Added a wrapper for the rendering as `storybook/html` addons cannot render HTML, just text. React wrapper is written in NO JSX syntax to avoid build complexity

### DIFF
--- a/.storybook/addons/markdown.js
+++ b/.storybook/addons/markdown.js
@@ -18,22 +18,22 @@ const e = React.createElement;
 addons.register(ADDON_ID, (api) => {
   const title = 'Readme';
   const channel = addons.getChannel();
-  let initalText = DEFAULT_MSG;
-  let initalTemplate = '';
+  let initialText = DEFAULT_MSG;
+  let initialTemplate = '';
 
   class Wrapper extends Component {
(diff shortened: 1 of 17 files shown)


.storybook/addons/markdown.js
```diff
@@ -0,0 +1,70 @@
...
+    constructor(props) {
+      super(props);
+      this.state = {
+        html: initalText,
+      };
+      this.onAddonAdded = this.onAddonAdded.bind(this);
+    }
+
+    onAddonAdded ({ html }) {
+      initalText = html;
```
MKaHead: Typo initial and why the overwrite here? not strong opinions here

LucaMele: @MKaHead thanks, cause of multiple re-rendering we need to store last one outside react. Is like setting default dynamically. Will fix typo

PR #917, opened 2019-04-13
The AXA TABLE

LucaMele:
I would merge it to test it in create react app and as its getting a big PR. still open after merge:

TODO:</br>

     - Allow cells formatting like axa-table</br>
     - Allow to add callback events onChange or onClick when row is pressed</br>
     - Add ui tests</br>

Add axa-table to v2


LucaMele:
Will use ARIA to declaer that is a table. Native table selector is not possibile to use unless a consumer passes all the table inclusiv of <table>. Browser will brake if we pass as local dom only <tr><td> for example.

Here the proof that Screen readers can see it as a table:

Bildschirmfoto 2019-04-13 um 10 31 56

src/components/30-organisms/table/README.md
```diff
@@ -0,0 +1,57 @@
+# Table
```
raphaellueckl: Do we have to define code in here? IMHO that can get outdated very fast and this is very annoying to any users who want to implement this table and this the documentation is always "real". We are consisten everywhere else, but if we happen to get into such a situation here, the users might think that the documentation might be outdated everywhere else too.

src/components/30-organisms/table/index.js
```diff
@@ -0,0 +1,39 @@
+import { LitElement, css, unsafeCSS } from 'lit-element';
+import defineOnce from '../../../utils/define-once';
+import tableCss from './index.scss';
+
+// subcomponents that can only be used with axa-table:
```
raphaellueckl: What does this comment mean? Axa-table is the definition and not a subcomponent that can be used withing axa-table, right? But I guess I'm just confused here. :)

src/components/30-organisms/table/index.js
```diff
@@ -0,0 +1,39 @@
...
+  static get styles() {
+    return css`
+      ${unsafeCSS(tableCss)}
+    `;
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+    const table = this.querySelector('table');
+    if (table) {
```
raphaellueckl: What if the table is not there, is there even a use case for that?

src/components/30-organisms/table/index.js
```diff
@@ -0,0 +1,39 @@
...
+
+// semantics are done with aria: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Table_Role
+class AXATable extends LitElement {
+  static tagName = 'axa-table';
+
+  static get tagName() {
+    return 'axa-table';
+  }
+
+  static get styles() {
```
raphaellueckl: I like how we write that now :) 👍

src/components/30-organisms/table/README.md
LucaMele: Yeah i agree.. so we need to make a mechnaism that reads react code and previews it as atm my addon does only wc code

src/components/30-organisms/table/index.js
LucaMele: legacy.. i remove it

src/components/30-organisms/table/index.js
LucaMele: defensive programming.. if table is not here, nothing happens

src/components/30-organisms/table-sortable/index.js
```diff
@@ -0,0 +1,206 @@
...
+      thead: { length: theadL },
+      tbody: {
+        0: { length: tbodyL },
+      },
+      tfoot: {
+        0: { length: tfootL },
+      },
+    } = this.model;
+
+    if (theadL !== tbodyL && (!tfootL || tfootL === tbodyL)) {
```
MKaHead: Could that return directly, for example like so `return theadL !== tbodyL && (!tfootL || tfootL === tbodyL)`

src/components/30-organisms/table/index.js
MKaHead: I would take the lit element firstUpdated() method for the connectedCallback, because there lifecycle callback really called once "Implement firstUpdated to perform one-time work after the element’s template has been created." @source lit element docs

src/components/30-organisms/table-sortable/index.js
LucaMele: nice catch! thanks -> done

LucaMele:
thanks @MKaHead, nice catchs! i implemented the PR comments

src/components/30-organisms/table-sortable/story.js
```diff
@@ -0,0 +1,73 @@
...
+const model = {
+  thead: [
+    { html: 'Title 0', sort: 'ASC' },
+    { html: 'Title 1', sort: 'ASC' },
+    { html: 'Title 3', sort: 'DESC' },
+    { html: 'Title 2' },
+  ],
+  tbody: [
+    [
+      { html: '<span>11 Gaa</span>' },
```
jackblackCH: Please use better "test" or even real case test strings in cells. (Lipsum, names..., longer, shorter)

src/components/30-organisms/table-sortable/ui.test.js
```diff
@@ -0,0 +1,217 @@
+import { Selector } from 'testcafe';
+
+fixture('Table Sortable - basic functionality').page(
+  'http://localhost:9999/iframe.html?id=organisms-table--table-sortable-default'
```
jackblackCH: take host and port from 'env'. It exists.

src/components/30-organisms/table-sortable/ui.test.js
```diff
@@ -0,0 +1,217 @@
...
+  await t
+    .click($columnOne)
+    .expect($columnOne.getAttribute('class'))
+    .eql('o-table-sortable__th--selected');
+
+  await t.expect($columnTwo.getAttribute('class')).eql('');
+});
+
+fixture('Table Sortable - innerscroll functionality').page(
+  'http://localhost:9999/iframe.html?id=organisms-table--table-sortable-innerscroll'
```
jackblackCH: same, take it from `env vars`

src/components/30-organisms/table/ui.test.js
```diff
@@ -0,0 +1,42 @@
+import { Selector } from 'testcafe';
+
+fixture('Table - basic functionality').page(
+  'http://localhost:9999/iframe.html?id=organisms-table--table-default'
```
jackblackCH: same

src/components/30-organisms/table-sortable/README.md
```diff
@@ -0,0 +1,47 @@
...
+    [
+      { html: '<span>2 Gaa</span>' },
+      { html: '<span>aHello 3</span>' },
+      { html: '<span>Cell 2</span>' },
+      { html: 'C' },
+    ],
+  ],
+};
+```
+
```
raphaellueckl: Isn't it sufficient to provide this as story? Instead as code that could get outdated at some point?

src/components/30-organisms/table-sortable/index.js
```diff
@@ -0,0 +1,202 @@
...
+import '../table';
+
+const ASC = 'ascending';
+const DESC = 'descending';
+
+const mapAsc = {
+  [ASC]: 'ASC',
+  [DESC]: 'DESC',
+};
+
```
raphaellueckl: What's the use of this upper block? Is it for documentation, so that people understand what asc and decs means? I would completely get rid of it if possible and just name it ASC and DESC. This const is only used once and I don't see a need for this structure.

src/components/30-organisms/table-sortable/index.js
```diff
@@ -0,0 +1,202 @@
...
+      thead: { length: theadL },
+      tbody: {
+        0: { length: tbodyL },
+      },
+      tfoot: {
+        0: { length: tfootL },
+      },
+    } = this.model;
+
+    return !(theadL !== tbodyL && (!tfootL || tfootL === tbodyL));
```
raphaellueckl: What does the L's stand for? tbodyL?

LucaMele:
@jackblackCH nice catch! should staring done

src/components/30-organisms/table-sortable/ui.test.js
LucaMele: button is also using loclahost. I will fix it but also button needs a fix here

src/components/30-organisms/table-sortable/index.js
```diff
@@ -0,0 +1,202 @@
...
+      this.model = { ...this.defaultModel, ...this.model };
+    }
+    if (!this.validateModel()) {
+      throw new Error('Invalid model provided');
+    }
+    return super.shouldUpdate(...args);
+  }
+
+  render() {
+    const { thead, tbody, tfoot } = this.model;
```
raphaellueckl: I don't like this. But I know half of the team does.

src/components/30-organisms/table-sortable/ui.test.js
LucaMele: and the scaffolding script too

src/components/30-organisms/table-sortable/index.scss
```diff
@@ -0,0 +1,65 @@
...
+      border-style: solid;
+    }
+  }
+
+  th[aria-sort="ascending"] {
+    &:after {
+      border-bottom-width: 4px;
+      border-bottom-color: $color-prim-gray-silver;
+      border-top: none;
+      top: -8px;
```
raphaellueckl: 😬

src/components/30-organisms/table-sortable/index.scss
```diff
@@ -0,0 +1,65 @@
...
+      }
+    }
+  }
+
+  th[aria-sort="descending"] {
+    &:after {
+      border-top-width: 4px;
+      border-top-color: $color-prim-gray-silver;
+      border-bottom: none;
+      top: -4px;
```
raphaellueckl: 😬

src/components/30-organisms/table-sortable/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "@axa-ch/table-sortable",
+  "version": "0.0.0",
+  "description": "> TODO: description",
```
raphaellueckl: I vote for writing something here. Just a little explanation, no rocket science stuff. Everything is better than a todo imho.

src/components/30-organisms/table-sortable/story.js
```diff
@@ -0,0 +1,73 @@
...
+    () => `
+      <h3>AXA Table Sortable Default</h3>
+      <p style="color: red; font-weight:bold;">WARNING: on mobile sortable not implemented, use innerscroll option till finished<p/>
+      <axa-table-sortable model='${JSON.stringify(model)}'>
+      </axa-table-sortable>
+
+
+
+      </br>
+      </br>
```
raphaellueckl: I do that myself, but not sure if we should try to avoid spamming those <br>'s everywhere.

src/components/30-organisms/table-sortable/ui.test.js
LucaMele: done

src/components/30-organisms/table-sortable/ui.test.js
LucaMele: done

src/components/30-organisms/table/ui.test.js
LucaMele: done

src/components/30-organisms/table-sortable/story.js
LucaMele: ok i will

src/components/30-organisms/table-sortable/ui.test.js
```diff
@@ -0,0 +1,217 @@
+import { Selector } from 'testcafe';
```
raphaellueckl: Why do you use $ for variable names everywhere in here? Just curious, since we don't do that in other tests.

I know for instance from angularjs that everything prefixed with $ is a framework related variable, but I don't see the connection to here. $router in angular for instance, is the angular router you can inject everywhere, but here, the $ prefixed variables are not injected, they were declared by us.

src/components/30-organisms/table/index.js
raphaellueckl: Not sure, but if the table is not there, then we have a huge problem, right? Should we really check that? Why don't we check it in all the other methods?

src/components/30-organisms/table/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "@axa-ch/table",
+  "version": "0.0.0",
+  "description": "> TODO: description",
```
raphaellueckl: Also here: I would put a description

src/components/30-organisms/table-sortable/story.js
LucaMele: done

src/components/30-organisms/table-sortable/index.js
jackblackCH: probably for lenght. I suggest to not use abbreviations

src/components/30-organisms/table-sortable/index.js
```diff
@@ -0,0 +1,202 @@
...
+  // and space complexity is O(log(n))
+  sort(arr, index, sortAs) {
+    return arr.sort((rowLx, rowRx) => {
+      const {
+        [index]: { html: cellRx },
+      } = rowRx;
+      const {
+        [index]: { html: cellLx },
+      } = rowLx;
+      const cleanCellRx = cellRx.replace(/<[^>]*>/g, '').trim();
```
jackblackCH: Rx and Lx? same. It's hard to figure out and maintain.

src/components/30-organisms/table-sortable/README.md
LucaMele: is not a code snipped, just a guide for on how to use it. Main reaosn: i want to specify that: ... here HTML5 conform table, having thead, tbody and tfoot

LucaMele: i would like to keep it

src/components/30-organisms/table-sortable/index.js
LucaMele: ?? the usage is to map 'ascending' to 'ASC', as i need it lower in the code...

LucaMele: cant get rid of it

src/components/30-organisms/table-sortable/index.js
LucaMele: length, i will rename it

LucaMele: i cant as prettier now makes it even less readable:

```
return !(
      theadLength !== tbodyLength &&
      (!tfootLength || tfootLength === tbodyLength)
    );
```

i leave it

src/components/30-organisms/table-sortable/index.js
LucaMele: u dont like what?

src/components/30-organisms/table-sortable/package.json
LucaMele: agree

LucaMele: done

src/components/30-organisms/table-sortable/story.js
LucaMele: necessary here and is just demo

src/components/30-organisms/table-sortable/ui.test.js
LucaMele: marco did and i guess is a best practices that comes from jQuery where $ represents a living DOM node

src/components/30-organisms/table/package.json
LucaMele: done

src/components/30-organisms/table-sortable/index.js
LucaMele: right and left, why is that hard to maintain?

LucaMele: most sort algorithm have this "right" and "left" naming, every eng knows that so why hard to maintain?

LucaMele:
Bildschirmfoto 2019-05-03 um 18 29 21

LucaMele:
Bildschirmfoto 2019-05-03 um 18 44 10

PR #919, opened 2019-04-17
Feature/submit button

MKaHead:
**General:**
-  Change all proposal class properties to static get methods
- Fix global collision css coming from the controlled input demos.

**Button Component**
- Button consumes now the icon component as dependency etc.
- Implement submit functionality
- Add more stories for submit functionality in native stories and react demo
- Import webcomponents polyfill, because every component is plug and play
- Improve React Wrapper - set default values for the props, because the webcomponent constructor don't set default values of the React props coming in.
- Add defineOnce function

**Icon Component**
- Add defineOnce function
- Import webcomponents polyfill and as dependency, because every component is plug and play

**Storybook**
- Update Storybook version

### DIFF
--- a/src/components/10-atoms/icon/index.js
+++ b/src/components/10-atoms/icon/index.js
@@ -2,37 +2,33 @@ import '@webcomponents/webcomponentsjs';
 import { LitElement, svg } from 'lit-element';
 // TODO fix that stuff
 /* eslint-disable import/no-extraneous-dependencies */
-import {
-  ArrowRightSvg,
-  CollapseSvg,
-  DownloadSvg,
-  EmailSvg,
-  ExpandSvg,
-  PhoneSvg,
-  PlusSvg,
-  SearchSvg,
-  UploadSvg
-} from '@axa-ch/materials';
-import defineOnce from "../../../utils/define-once";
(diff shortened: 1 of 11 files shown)


src/components/10-atoms/icon/index.js
```diff
@@ -28,11 +29,10 @@ class AXAIcon extends LitElement {
     'upload': UploadSvg
   };
 
-  static get properties() {
-    return {
-      icon: { type: String },
-    };
-  }
+  static properties = {
```
raphaellueckl: get (as discussued)

src/demo/demo-controlled-inputs/App.css
```diff
@@ -16,18 +16,18 @@
   font-size: 16px;
 }
 
-article {
-  padding: 1rem;
-  box-shadow: 1px 2px 3px #ddd;
-}
+/*article {*/
```
raphaellueckl: Pls remove outcommented code. If we need it in the future, we have version control to see how it was.

src/components/10-atoms/icon/index.js
MKaHead: fixes #872ae70

src/demo/demo-controlled-inputs/App.css
MKaHead: fixes in #f027467

PR #922, opened 2019-04-18
Add max 80 line width allowence

LucaMele:
Sets the max width to 80 as we decided in the team


LucaMele:
Make sure your IDE is compliant with it: https://editorconfig.org/#download

raphaellueckl:
I don't get what benefit we get out of that now. Are we now reminded by the editor that our lines are too long? And how do we validate it on the CI side?

PR #926, opened 2019-04-25
Feature/enable axa link for footer

raphaellueckl:

 - We create a version of link here to support the axa-footer-small


raphaellueckl:
Closed because we go with the native html tag solution.

PR #928, opened 2019-04-25
Feature/axa footer with html links

raphaellueckl:

 - Remove the axa link from the footer

Reason:
We would have needed to write a link item specifically for the footer, that isn't used anywhere else. This is not only overengineered, we also have to make sure that nobody else would use this type of link since it is otherwise not according to the styleguide.


LucaMele:
Makes sense :)

PR #930, opened 2019-04-26
Add 'npm run build' to 'npm test'

raphaellueckl:

We had the issue, that everything was working fine, but ```npm run build``` crashed with a dependency problem. To avoid that, we put it into the ```npm test```, to make sure that the local npm test does the same as the CI pipeline. We put it into test, because npm test should make sure that the build does not break. We should not diverge the things the CI and out local npm test do. To always be sure, that a build will be green with a local ```npm test```, we put it into here, because we have to make sure, that this does not fail.


jackblackCH:
This is "wrong", as it's super expensive to run a build after each test command

PR #931, opened 2019-04-29
Feature/footer small release including documentation

raphaellueckl:

 Release branch of footer small, which has been published as v1.0.0 to npm.

### DIFF
--- a/src/components/20-molecules/footer-small/package.json
+++ b/src/components/20-molecules/footer-small/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@axa-ch/footer-small",
-  "version": "1.0.0",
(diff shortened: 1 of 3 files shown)


src/components/20-molecules/footer-small/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "@axa-ch/footer-small",
+  "version": "1.0.0",
```
LucaMele: Are we released? is it not better to just release beta till is completly merged?

raphaellueckl: Yes we are already released. :/ Marek wanted to use it and is now doing tests with it.

PR #935, opened 2019-04-30
Feature/axa link with variants

raphaellueckl:

AXA link with all requested variants.

State:
- ✓ Implemented
- ✓ Design reviewed
- ✓ New tests written

### DIFF
--- a/src/components/10-atoms/link/README.md
+++ b/src/components/10-atoms/link/README.md
@@ -1,3 +1,5 @@
 # Link
 
-The link component meant for hyper- and internal links.
+The link component meant for hyper and internal links.
+
+The link has a `variant` property with different parameters that can be provided. Adding a `external` attribute to it will make sure that on-click the link will open in a new tab.
--- a/src/components/10-atoms/link/README.md
+++ b/src/components/10-atoms/link/README.md
@@ -1,3 +1,3 @@
 # Link
 
-TODO Description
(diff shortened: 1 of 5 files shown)


src/components/10-atoms/link/README.md
```diff
@@ -1,3 +1,5 @@
 # Link
 
-TODO Description
+The link component meant for hyper and internal links.
+
+The link has a `variant` property with different parameters that can be provided. Adding a `external` attribute to it will make sure that on-click the link will open in a new tab.
```
LucaMele: we should add a desc for all possible variants

raphaellueckl: I disagree. This desc is prone to degenerate, since we would always have to check the documentation when making changes. I hope that we can agree on keeping the variants visible from the examples. What do you think?

LucaMele: I think @MKaHead just speccified them all and would be better for user. If we have a automatic version, would be better but till then at least lets keep consistent between us. I opt for more docu, even if docu gets bad in time as its not sync with code but for that we can plan in next PI a script

raphaellueckl: Can you reconsider this review, since we had the discussion on the daily?

LucaMele: yepp ok

PR #936, opened 2019-05-02
feat(footer-small): add TS type declaration file index.react.d.ts

raphaellueckl:

Add typescript types for footer


raphaellueckl:
> We should add this to scaffolding too. Would be great if u can add it

I agree. This should be done in a separate branch. I want to do it. I never touched the scaffolding so far and I guess this gives a good entry point into it. :)

raphaellueckl:
@LucaMele Done: https://github.com/axa-ch/patterns-library/pull/948

PR #940, opened 2019-05-03
Feature/datepicker final release preparation

jackblackCH:
- Adds the readme and doc
- Smoke tests for react
- A unit test


src/components/20-molecules/datepicker/index.js
```diff
@@ -19,6 +19,7 @@ export class Datepicker extends LitElement {
 
   static get properties() {
     return {
+      'data-test-id': { type: String, reflect: true },
```
LucaMele: why reflect? we dont need to change it to outside

src/components/20-molecules/datepicker/index.react.js
```diff
@@ -0,0 +1,7 @@
+import withReact from '../../../utils/with-react';
+import AXADatepicker from './index';
+
+export default createElement => props =>
+  withReact(createElement)(AXADatepicker.tagName, {
+    ...props,
```
LucaMele: dont use deconstructor here, we agreed with @MKaHead that we explicitly whitelist props not pass all from react to CE

src/components/20-molecules/datepicker/react/story.jsx
```diff
@@ -0,0 +1,24 @@
...
+import ReactDOM from 'react-dom';
+import AXADatepickerReact from './AXADatepickerReact';
+import { withMarkdown } from '../../../../../.storybook/addons/markdown';
+import Readme from '../README.md';
+
+storiesOf('Molecules/Datepicker/React', module)
+  .addDecorator(withMarkdown(Readme))
+  .add('Datepicker as React Component', () => {
+    const div = document.createElement('div');
+    ReactDOM.render(
```
LucaMele: there we go :)

src/components/20-molecules/datepicker/index.js
jackblackCH: It's not passed down to the react component without, and then the smoke test fails.

src/components/20-molecules/datepicker/index.react.js
jackblackCH: Fixed. Is there a specific reason we do that?

src/components/20-molecules/datepicker/index.js
LucaMele: weird as reflect just makes sure that changes inside the component are synced up to the attr.. pretty sure u don need it

src/components/20-molecules/datepicker/index.react.js
LucaMele: I think so that not all props from the react comp go on the CE but only those that are whitelisted.. like a interface that is strict and protected. but i did the same changes as u did and talked to marco and he gave me another explanation that i cant remeber but made sense :)

src/components/20-molecules/dropdown/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "@axa-ch/dropdown",
+  "version": "1.0.0-alpha.2",
+  "description": "The dropdown for the AXA patterns library",
+  "author": "Pattern Warriors",
+  "homepage": "https://github.com/axa-ch/patterns-library",
```
LucaMele: this should be to the component, please see other

jackblackCH: done.

src/components/10-atoms/button/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@axa-ch/button",
-  "version": "0.0.0-beta.1",
+  "version": "1.0.0-alpha.1",
```
LucaMele: we need to be consisten here everyone does its own versioning.. till today, every axa project used 0.0.0 as start version that on npm is officially as WIP and then tagged it with beta. I would vote for that naming convention but am ok also with 1.0.0-alpha which says 1.0.0 is almost ready.. but consistence is important so we need to define which format we will use

PR #949, opened 2019-05-07
Add description to Table Sortable

LucaMele:


src/components/30-organisms/table-sortable/README.md
```diff
@@ -23,27 +129,29 @@ const model = {
...
       { html: '<span>Cell 2</span>' },
       { html: 'B' },
     ],
     [
-      { html: '<span>2 Gaa</span>' },
-      { html: '<span>aHello 3</span>' },
+      { html: '<span>2 Test</span>' },
+      { html: '<span>A Hello 3</span>' },
       { html: '<span>Cell 2</span>' },
       { html: 'C' },
```
raphaellueckl: I agree with what Marc wrote somewhere, that quality test data would be nice, but we can also do that later.

PR #956, opened 2019-05-07
[WIP] Set up CI with Azure Pipelines

raphaellueckl:


jackblackCH:
Missing empty end of line

jackblackCH:
We should work with consistent node versions in all local, dev, prod etc. environments. AFAIK AXA has to use node 8 for some reasons? @LucaMele

LucaMele:
@jackblackCH AXA has ONLY Node 6 and 8.8.1 on their Cloud Open Shift Platforms and Jenkins... but i think on azure we are free.. maybe @raphaellueckl check with Shelko if we are free to choose the Node Version we want

raphaellueckl:
old

PR #977, opened 2019-05-10
Set up CI with Azure Pipelines

raphaellueckl:


raphaellueckl:
not needed, was only required by azure

PR #1002, opened 2019-05-13
Introduce AXA Footer

raphaellueckl:
Introduces the big axa-footer.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "test-ie": "sh scripts/testcafe-ie.sh",
     "test-jest": "node scripts/test.unit.js --env=jsdom",
     "test-ui": "TEST_HOST_STORYBOOK_URL=http://localhost:6006 testcafe chrome ./**/ui.test.js",
+    "test-ui-hl": "TEST_HOST_STORYBOOK_URL=http://localhost:6006 testcafe 'chrome:headless' ./**/ui.test.js",
     "new": "node scripts/create-a-m-o/index.js"
   },
   "dependencies": {
--- a/src/components/30-organisms/footer/README.md
+++ b/src/components/30-organisms/footer/README.md
@@ -235,4 +235,4 @@ Structure:
 
 ### dynamic
 
-Dynamic specifies, if you want the footer to contain static links, that will just route you forward, or if you want to use callback to react on clicks yourself. If dynamic is not set, the callbacks you hand into the component will not have any effect.
+Set the dynamic property (boolean), if you want to use callbacks for links. Otherwise they will not work.
--- a/src/components/30-organisms/footer/README.md
+++ b/src/components/30-organisms/footer/README.md
@@ -198,9 +198,40 @@ Import the footer-defining script and use a footer like this:
 
 Title and links to display within the footer.
 
+Structure:
+
+```typescript
+[
+  {
+    title: string,
+    items: [
+      {
+        text: string,
+        link: string,
+        external: boolean,
+      },
+    ],
+  },
+];
+```
+
 ### social
 
 Social media icons and their corresponding links. Also coming with a customizable title.
+
+Structure:
+
+```typescript
+{
+  title: string,
+  icons: [
+    {
+      title: string,
+      link: string
+    },
+  ],
+};
+```
 
 ### dynamic
 
--- a/src/components/30-organisms/footer/demo.js
+++ b/src/components/30-organisms/footer/demo.js
@@ -7,7 +7,7 @@ import Readme from './README.md';
 
 storiesOf('Organisms/Footer/Demos', module)
   .addDecorator(withMarkdown(Readme))
-  .add('Footer - Dynamic Links', () => {
+  .add('Feature - Footer Callbacks', () => {
     const content = [
       {
         title: 'axa & you',
--- a/src/components/30-organisms/footer/demo.js
+++ b/src/components/30-organisms/footer/demo.js
@@ -4,9 +4,11 @@ import './index';
 
 import { withMarkdown } from '../../../../.storybook/addons/markdown';
 import Readme from './README.md';
+import withBodyReset from '../../../../.storybook/addons/reset-body';
 
 storiesOf('Organisms/Footer/Demos', module)
   .addDecorator(withMarkdown(Readme))
+  .addDecorator(withBodyReset())
   .add('Feature - Footer Callbacks', () => {
     const content = [
       {
--- a/src/components/30-organisms/footer/demo.js
+++ b/src/components/30-organisms/footer/demo.js
@@ -83,7 +83,6 @@ storiesOf('Organisms/Footer/Demos', module)
     wrapper.appendChild(clickedLink);
 
     footer.addEventListener('axa-link-click', ev => {
-      const clickedLink = document.getElementById('clicked-link');
       clickedLink.innerText = `Last clicked link: ${ev.detail}`;
     });
 
--- a/src/components/30-organisms/footer/index.js
+++ b/src/components/30-organisms/footer/index.js
@@ -153,17 +153,19 @@ class AXAFooter extends LitElement {
       <footer class="o-footer">
         <axa-container>
           <div class="o-footer__mobile">
-            ${repeat(
-              this.content,
-              (contentItem, index) =>
-                html`
-                  <div class="o-footer__accordion">
-                    ${this.getFooterHeader(contentItem, index)}
-                    ${this.getFooterContentList(contentItem)}
-                  </div>
-                `
-            )}
-            ${this.getFooterIcons()}
+            <div class="o-footer__collection">
+              ${repeat(
+                this.content,
+                (contentItem, index) =>
+                  html`
+                    <div class="o-footer__accordion">
+                      ${this.getFooterHeader(contentItem, index)}
+                      ${this.getFooterContentList(contentItem)}
+                    </div>
+                  `
+              )}
+              ${this.getFooterIcons()}
+            </div>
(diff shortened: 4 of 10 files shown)


src/components/30-organisms/footer/README.md
```diff
@@ -0,0 +1,3 @@
+# Footer
+
+TODO Description
```
LucaMele: TODO :)

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,176 @@
+/* eslint-disable import/no-extraneous-dependencies */
+import '@webcomponents/webcomponentsjs';
```
LucaMele: delete please

LucaMele:
First Review FB:

- Rebase PR
- Add axa-container
- Park your code, re-run `npm run new`, create the o-footer again and cherrypick your code back to the new fotter
- Replace main article with footer tag
- HandleClick can be optimised. Now is o(n) due to map. If you store last one and reset only last element it can be o(1).
- Please BEMify all classes (`mobile` -> `o-footer__mobile`)
- Maybe check if there is a way to avoid having a section for mobile and a section for tablet if it doesnt blow the logic too much. Cause now we have quite a lot of dead DOM nodes in every display size
- Links must also work with react callbacks not just `<a href="">`
- `<div class="o-footer__collection-container-list">` class name demands its correct HTML tag. Please refactor this to `ul / li` same as u did here: `o-footer__social-media-list`

src/components/30-organisms/footer/index.js
raphaellueckl: Are you sure? This is setup by the scaffolding script. Why don't we need to import it? Where is it injected?

raphaellueckl:
State:
======

First Review FB:
- ❓`<div class="o-footer__collection-container-list">` class name demands its correct HTML tag. Please refactor this to `ul / li` same as u did here: `o-footer__social-media-list`

--> I tried that. I put an `ul` with 2 columns, but they did not render properly. I wasted quite some time. The second column was basically rendered outside its box and overlapped other stuff. Maybe you have a better idea (display block and other consorts did not help here). My flexbox approach is far from ideal, since it also changes the ordering of the items.

- ❓HandleClick can be optimised. Now is o(n) due to map. If you store last one and reset only last element it can be o(1).

--> I tried it out and I wasn't able to come up with a better solution that reduces complexity. I tried moving stuff into temporary variables, but I always ended up having to loop somewhere. If you could provide an example, that would be helpful.

- ❌ Please BEMify all classes (`mobile` -> `o-footer__mobile`)
- ❌ Maybe check if there is a way to avoid having a section for mobile and a section for tablet if it doesnt blow the logic too much. Cause now we have quite a lot of dead DOM nodes in every display size
- ✅ Rebase PR
- ✅ Add axa-container
- ✅ Park your code, re-run `npm run new`, create the o-footer again and cherrypick your code back to the new footer
- ✅ Replace main article with footer tag
- ✅ Links must also work with react callbacks not just `<a href="">`

package.json
```diff
@@ -22,6 +22,8 @@
     "test-chrome": "sh scripts/testcafe-chrome.sh",
     "test-ie": "sh scripts/testcafe-ie.sh",
     "test-jest": "node scripts/test.unit.js --env=jsdom",
+    "test-ui": "TEST_HOST_STORYBOOK_URL=http://localhost:6006 testcafe chrome ./**/ui.test.js",
+    "test-ui-hl": "TEST_HOST_STORYBOOK_URL=http://localhost:6006 testcafe 'chrome:headless' ./**/ui.test.js",
```
LucaMele: If you refactor it, please change also the testscript `test calls test-chrome, instead should call test-ui-hl imho` to use `test-ui-hl` instead of calling directly then `sh` script

src/components/30-organisms/footer/README.md
```diff
@@ -0,0 +1,238 @@
...
+
+```bash
+npm install @axa-ch/footer
+```
+
+```js
+import '@axa-ch/footer';
+...
+
+const content = JSON.stringify([
```
LucaMele: I believe, this is not possibile here to have a model. The footer must be indexed via google bot and needs its content inside DOM children.

LucaMele: Please add aversion of purely DOM children *or* refactor this

LucaMele: have a look at this SEO content as an inspiration example: https://github.com/axa-ch/patterns-library/blob/develop-v1/src/components/o-commercial-hero-cover/_preview.html

src/components/30-organisms/footer/demo.js
```diff
@@ -0,0 +1,90 @@
+/* global document */
+import { storiesOf } from '@storybook/html';
+import './index';
+
+import { withMarkdown } from '../../../../.storybook/addons/markdown';
+import Readme from './README.md';
+
+storiesOf('Organisms/Footer/Demos', module)
+  .addDecorator(withMarkdown(Readme))
```
LucaMele: Please add `withBodyReset` for better demo effect and effective MQ matching

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,161 @@
...
+      <footer class="o-footer">
+        <axa-container>
+          <div class="o-footer__content">
+            <div class="o-footer__collection">
+              ${repeat(
+                this.content,
+                (contentItem, index) =>
+                  html`
+                    <div class="o-footer__main">
+                      <strong class="o-footer__title-desktop"
```
LucaMele: This has to be a title `<h1>` or `<h2>` and therefore `<footer>` tag wont work anymore due to the `h1` tags. 

This is the outline of the footer ATM which is not possible to have for SEO:

Bildschirmfoto 2019-05-28 um 09 26 02

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,161 @@
...
+                        >${contentItem.title}</strong
+                      >
+                      <button
+                        class="o-footer__accordion-button"
+                        @click="${ev => this.handleClick(ev, index)}"
+                      >
+                        <strong class="o-footer__title">
+                          ${contentItem.title}
```
LucaMele: dtto title

package.json
raphaellueckl: This is just for local testing and not meant to overrule the other scripts. But usually, when you develop, you have an instance of storybook running and then it's convenient when you run the tests on top without needing to fire up storybook again although it is already running.

src/components/30-organisms/footer/index.scss
```diff
@@ -0,0 +1,233 @@
...
+      @include typo(primary, medium, bold);
+
+      display: none;
+
+      text-transform: uppercase;
+      line-height: 1.5;
+    }
+
+    &-list {
+      // Reset browser styles for ul
```
jackblackCH: no need for these kinda comments

src/components/30-organisms/footer/index.scss
```diff
@@ -0,0 +1,233 @@
...
+    display: none;
+  }
+
+  &__accordion-button {
+    @include typo(primary, default, medium);
+    display: flex;
+    justify-content: space-between;
+    color: $color-prim-white;
+    background-color: $color-prim-blue-dark-indigo;
+    padding: 20px 30px 20px 30px;
```
jackblackCH: can be shortened as: `padding: 20px 30px`

src/components/30-organisms/footer/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@axa-ch/footer",
+  "version": "0.0.1-beta.5",
+  "description": "The big version of the footer.",
+  "author": "Pattern Warriors",
+  "homepage": "https://github.com/axa-ch/patterns-library/tree/master/src/components/30-organisms/footer#readme",
+  "license": "Copyright 2019 AXA Versicherungen AG",
+  "main": "lib/index.js",
```
jackblackCH: missing build scripts (check out other package.json)

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,162 @@
...
+    this.iconMap = new Map();
+    this.iconMap.set('facebook', FacebookSvg);
+    this.iconMap.set('instagram', InstagramSvg);
+    this.iconMap.set('twitter', TwitterSvg);
+    this.iconMap.set('xing', XingSvg);
+    this.iconMap.set('youtube', YoutubeSvg);
+    this.iconMap.set('linkedin', LinkedinSvg);
+  }
+
+  handleClick(e, index) {
```
jackblackCH: What is this needed for and what is the difference between the handleLinkClick?

jackblackCH: Was is the index needed for and is it reliable?

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,162 @@
...
+    e.stopPropagation();
+    const oldState = this.content[index].active;
+    this.content = this.content.map(c => {
+      c.active = false;
```
jackblackCH: what is `c`? an item?

src/components/30-organisms/footer/index.js
jackblackCH: We have a separate @axa-ch/polyfill package that includes it. Consumers must include it. It comes with inconvenient side effects if we include them in our packages.

jackblackCH:
Closed. As discussed. Newer PR is open. Week the the branch.

PR #1012, opened 2019-05-16
Add social media svg icons

raphaellueckl:
Adds social media buttons to materials


cge:
Hi @raphaellueckl , I see these icons have `width` and `height` set to `100%`, whereas the existing icons in the library use fixed `px` values. What is the preferred approach?

raphaellueckl:
@cge We should use % and not px. We are aware that we currently are not consistent there, but thanks for the hint! The important part is the viewBox.

PR #1025, opened 2019-05-20
Add fix to reset rest css that can be used on consumers website

### LINKED ISSUE #1021: Button eventuell broken
dede

type='button'

LucaMele:
Reset common reset.css/normalise.css out there.

they add styling like so: `[type="button"]` and add appearance to 'button'

fixes #1021

### DIFF
--- a/src/components/10-atoms/button-link/index.js
+++ b/src/components/10-atoms/button-link/index.js
@@ -43,6 +43,15 @@ class AXAButtonLink extends LitElement {
     this.onClick = () => {};
   }
 
+  firstUpdated() {
+    const { style } = this;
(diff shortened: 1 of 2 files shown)


src/components/10-atoms/button-link/index.js
```diff
@@ -43,6 +43,15 @@ class AXAButtonLink extends LitElement {
     this.onClick = () => {};
   }
 
+  firstUpdated() {
+    const { style } = this;
```
raphaellueckl: Are you sure this works? Isn't it `this.styles`?

raphaellueckl: plural

PR #1029, opened 2019-05-22
Feature/add input file

LucaMele:
Ready for review


LucaMele:
First Review
- no fix ID on the `for` and `id` attribute. either put input inside label or generate a uuid (Ask @MKaHead as he have a similar issue)

src/components/10-atoms/button/README.md
```diff
@@ -71,6 +71,7 @@ Import the button-defining script and use a button like this:
 | `type="button"` | Default button type (default if omitted) |
 | `type="submit"` | Submit button submit for forms           |
 | `type="reset"`  | Reset button for forms                   |
+| `type="file"`   | Upload files for forms                   |
```
LucaMele: please remove this

src/components/10-atoms/input-file/index.js
```diff
@@ -0,0 +1,97 @@
...
+import styles from './index.scss';
+
+const TYPE = 'file';
+class AXAInputFile extends NoShadowDOM {
+  static get tagName() {
+    return 'axa-input-file';
+  }
+
+  static get styles() {
+    return styles;
```
raphaellueckl: @LucaMele Why don't we use the unsafe-wrapper here? If the tests are green, why do we need that wrapper at all?

src/components/10-atoms/input-file/index.js
```diff
@@ -0,0 +1,97 @@
...
+    this.icon = '';
+    this.refId =
+      this.id ||
+      `input-file-${Math.random()
+        .toString(36)
+        .slice(2)}`;
+    this.large = false;
+    this.motionOff = false;
+    this.disabled = false;
+    this.accept = 'image/jpg, image/jpeg, application/pdf, image/png';
```
raphaellueckl: Do we really want to restrict the file-list by default?

src/components/10-atoms/input-file/index.react.d.ts
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { Icon } from '@axa-ch/icon/lib/index.d';
+
+type Variant = 'secondary' | 'red' | 'inverted';
+
+interface AXAButtonProps {
```
raphaellueckl: I guess that's a copy pasta issue. Better chose `AXAInputFileProps` here.

src/components/10-atoms/input-file/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "@axa-ch/input-file",
+  "version": "0.0.0-alpha.0",
+  "description": "> TODO: description",
```
raphaellueckl: todo

src/components/10-atoms/input-file/react/story.jsx
```diff
@@ -0,0 +1,114 @@
...
+import ReactDOM from 'react-dom';
+import AXAInputFileReact from './AXAInputFileReact';
+import { withMarkdown } from '../../../../../.storybook/addons/markdown';
+import Readme from '../README.md';
+
+storiesOf('Atoms/Input File/React', module)
+  .addDecorator(withMarkdown(Readme))
+
+  /* Default */
+  .add('InputFile - Default', () => {
```
raphaellueckl: @LucaMele We have this redundancy all over the place. This is imho fine for this pull request, but in the future, we should think about Removing the component name in here, since the wrapping package (`Atoms/Input File/React`) already contains the name.

src/components/10-atoms/input-file/ui.test.js
```diff
@@ -0,0 +1,17 @@
...
+
+const TAG = 'axa-input-file';
+const CLASS = '.a-input-file';
+
+test('should render input-file', async t => {
+  const $axaElem = await Selector(TAG);
+  await t.expect($axaElem.exists).ok();
+  const $axaElemEl = await $axaElem.find(CLASS);
+  await t.expect($axaElemEl.exists).ok();
+});
```
raphaellueckl: Not sure if we can even write any tests for the file upload, since a lot of stuff is handled by the browser, but if possible, there might be some tests that would make sense, e.g. validating the color of the upload button or whatever. I didn't test the file upload myself, maybe it's so tiny, we don't need more. @LucaMele You know better if we should add some here. :)

src/components/10-atoms/input-file/index.js
LucaMele: @raphaellueckl because with the use of noShadow thats the requirement. Markus said he did document it somewhere

src/components/10-atoms/input-file/index.js
LucaMele: yes, cause everything open leaves space for zip which it get forgotten can have limitation of WAF or ESG

LucaMele: User can always override if they wan to

src/components/10-atoms/input-file/index.react.d.ts
LucaMele: @NoahLiechti yes please change that

src/components/10-atoms/input-file/package.json
LucaMele: @NoahLiechti yes please add a todo

src/components/10-atoms/input-file/react/story.jsx
LucaMele: @raphaellueckl @NoahLiechti : agree, fine for the PR and needs to be redefined. Also we have 3 times same style and we should start thinking of making a baseButton.. But not for this PR

src/components/10-atoms/input-file/ui.test.js
LucaMele: @raphaellueckl @NoahLiechti yes u cant really test native file-upload. But noah please add some test where u test the different attributes. Set multiple and check if multiple is set for example

LucaMele: also ok if you combine more then one attribute in one test as they are tiny

src/components/10-atoms/input-file/index.js
raphaellueckl: Weird

src/components/10-atoms/input-file/index.js
raphaellueckl: Leave space open for zip? What does that mean?

Is it bad if a zip file gets uploaded?

src/components/10-atoms/input-file/ui.test.js
NoahLiechti: Yes I'll write more tests. Thanks for the feedback!

src/components/10-atoms/input-file/package.json
NoahLiechti: I will. Thanks for the feedback!

src/components/10-atoms/input-file/index.react.d.ts
NoahLiechti: I will. Thanks for the feedback!

src/components/10-atoms/input-file/react/story.jsx
NoahLiechti: @raphaellueckl @LucaMele should I make an issue?

src/components/10-atoms/input-file/index.js
LucaMele: without security on the backend, yes is really bad

raphaellueckl: But security belongs to the backend. The backend has to make sure, that it is secured against attacks. Doing the security in the frontend while leaving the backend-endpoints naked is a flaw. But doesn't matter for this PR I guess. Just saying that we should never do security work because a backend might potentially not handle it. Then the backend needs an update. One could also just enable the zips in the frontend with the developer tools and still be able to update it, so that's not a proper security measure.

src/components/10-atoms/input-file/react/story.jsx
raphaellueckl: @NoahLiechti  No,  I guess that's ok.

src/components/10-atoms/input-file/demo.js
```diff
@@ -0,0 +1,27 @@
+/* global document */
+import { storiesOf } from '@storybook/html';
+// import { html, render } from 'lit-html';
```
raphaellueckl: Plz remove if not necessary

src/components/10-atoms/input-file/demo.js
```diff
@@ -0,0 +1,27 @@
...
+import { withMarkdown } from '../../../../.storybook/addons/markdown';
+import Readme from './README.md';
+import './index';
+
+storiesOf('Atoms/Input File/Demos', module)
+  .addDecorator(withMarkdown(Readme))
+
```
raphaellueckl: Empty line?

src/components/10-atoms/input-file/demo.js
```diff
@@ -0,0 +1,27 @@
...
+  .add('Feature - InputFile clickable', () => {
+    const inputFile = document.createElement('axa-input-file');
+    inputFile.setAttribute('multiple', 'multiple');
+    let counter = 0;
+    inputFile.innerHTML = `You have chosen ${counter} files`;
+
+    inputFile.addEventListener('input', e => {
+      const {
+        target: { files },
+      } = e;
```
raphaellueckl: I'm not a fan of that. If you use that variable everywhere, why not. But using destructuring for only one place where is it being used makes the code too unreadable without any benefit imho. I leave it up to you if you want to change it or not, because this is very opinionated. But you need 3 lines of code that are completely unnecessary.

src/components/10-atoms/input-file/index.scss
```diff
@@ -0,0 +1,172 @@
+.a-input-file {
```
raphaellueckl: @LucaMele @NoahLiechti  Did you guys test this on internet explorer? (I mean the whole component, not the commented line, since you are using flexbox and stuff)

src/components/10-atoms/input-file/ui.test.js
```diff
@@ -0,0 +1,89 @@
...
+  `${host}/iframe.html?id=atoms-input-file--input-file-default`
+);
+
+test('should set input file property disabled', async t => {
+  const setDisabled = ClientFunction(() => {
+    document.querySelector('axa-input-file').disabled = true;
+  });
+  await setDisabled();
+  const $axaElm = await Selector('axa-input-file');
+  await t.expect(await $axaElm.hasAttribute('disabled')).ok();
```
raphaellueckl: I guess this test can be removed. It doesn't really add value imho. This will always be true and if not, the element has a problem which will be detected by other tests.

LucaMele: True, @NoahLiechti sorry here you need to get the value of the input inside axa-input-file. this is valid for all of them

src/components/10-atoms/input-file/index.scss
LucaMele: pretty sure he did. @NoahLiechti ?

src/components/10-atoms/input-file/demo.js
LucaMele: it doesnt change that much but ok if you use it only once and the line you use it wont get to long makes no difference. but its up to you noah here

src/components/10-atoms/input-file/demo.js
LucaMele: agree

src/components/10-atoms/input-file/index.scss
NoahLiechti: @LucaMele Nein ich habe es auf Safari, Chrome und Firefox getestet, nicht aber auf IE (habe leider kein Browserstack). Könnte ich das an dich delegieren Luca?

LucaMele: @NoahLiechti. Ah ok dann am Montag kommst du zu uns und wir testen es

src/components/10-atoms/input-file/index.scss
```diff
@@ -0,0 +1,173 @@
...
+  	opacity: 0;
+  	overflow: hidden;
+  	position: absolute;
+  	z-index: -1;
+  }
+}
+
+.a-input-file--large {
+  min-height: 40px;
+  @include _text(default);
```
jackblackCH: It's very common to have @includes on the top of rules to be sure to override anything we imported. (preventing unusual side effects)

PR #1031, opened 2019-05-23
Update readme

raphaellueckl:
Add communication channels, refactor text

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,37 +2,28 @@
 
 [![Build Status](https://travis-ci.org/axa-ch/patterns-library.svg?branch=develop)](https://travis-ci.org/axa-ch/patterns-library)
 
-[Pattern Library Preview](https://patterns.axa.ch)
+## [>> Pattern Library](https://patterns.axa.ch)
 
 [Check out our Contribution File](https://github.com/axa-ch/patterns-library/blob/develop/CONTRIBUTION.md#rules-and-lintings)
 [Check out our Architecture File](https://github.com/axa-ch/patterns-library/blob/develop/ARCHITECTURE.md)
 [Check out our Code of Conduct](https://github.com/axa-ch/patterns-library/blob/develop/CODE_OF_CONDUCT.md)
 
-- We only export holistic and self contained WebComponents based on CustomElement, ShadowDom and HTML Templates.
-- We do not support Built-ins
+## Communication
+
+Use the following channels for different kind of requests/reports:
+- Bug, small change, "wish": https://github.com/axa-ch/patterns-library/issues
+- Questions, ask for help, request presentations, etc: Slack #patterns-lib-devs
+- Feature requests: Slack @martin.stuedle
+
+## What we deliver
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,14 +2,21 @@
 
 [![Build Status](https://travis-ci.org/axa-ch/patterns-library.svg?branch=develop)](https://travis-ci.org/axa-ch/patterns-library)
 
-[Pattern Library Preview](https://patterns.axa.ch)
+## [>> Pattern Library](https://patterns.axa.ch)
```
LucaMele: Please leave preview as it is a preview

LucaMele: People will not realise is a link to the preview

README.md
```diff
@@ -2,14 +2,21 @@
...
 [Check out our Code of Conduct](https://github.com/axa-ch/patterns-library/blob/develop/CODE_OF_CONDUCT.md)
 
-- We only export holistic and self contained WebComponents based on CustomElement, ShadowDom and HTML Templates.
-- We do not support Built-ins
+## Communication
+
+Use the following channels for different kind of requests/reports:
+- Bug, small change, "wish": https://github.com/axa-ch/patterns-library/issues
+- Questions, ask for help, request presentations, etc: Slack #patterns-lib-devs
+- Feature requests: Slack @martin.stuedle
```
LucaMele: I would say: components requests instead of feature

AndyOGo: @LucaMele @raphaellueckl 
Awesome to add these channels for communication.

Would you agree to add this information also inside an Github Issue template, so that every time if someone wants to create an issue and somehow missed this info, he will get it directly if he creates a new issue?
https://help.github.com/en/articles/creating-issue-templates-for-your-repository

README.md
raphaellueckl: I find that "Peview" is an unfortunate name. It's basically saying: "Well, we are not ready yet, but here you can take a glimpse at the current progress." That's not actually the case, since we have components released that can be used. Therefore we shouldn't put the prototype label on ourselves. Otherwise people won't take us seriously when we tell them that we are ready and on and then put a preview label. Can't we find another word for out "Storybook" presentation? Because I didn't know any better, that's why I removed it. Something like "Pattern Library Live" or whatever. I don't like that one too, but just as example. :)

README.md
raphaellueckl: @AndyOGo A very neat idea! This will maybe help some people that don't know better when they start an issue. This would lead to a cleaner workflow.

raphaellueckl: @LucaMele I think "Component Request" is a subset of "Feature". I would keep it as simple as possible. Otherwise we might get other big request and the complaint "But it is not a component what we request!! >:-() "

README.md
LucaMele: Well why change the fact. It is a preview of our components. The client does not use our components from storybook but from npm and storybook is a pre-view of what he gets from npm. It doesn’t mean at all that they are not finished. Pre view mean before viewing so i dont think there is a better matching name. Live is less specific. Please leave preview

LucaMele: C9BBDA22-98E9-445E-BD22-6C0A66A72A72

README.md
LucaMele: Ok good point

README.md
LucaMele: But if u cant live with preview use sometimes like: live, url, link, visible etc

PR #1037, opened 2019-05-27
Add deployment status badge

raphaellueckl:
Add deployment status badge


raphaellueckl:
I tweaked the azure pipeline a bit, when the build here was running. I guess this is the reason why it failed. I restarted it and it worked, but it does not update here as long as there is not commit.

PR #1039, opened 2019-05-28
[WIP] Add favicon to our storybook deployment

raphaellueckl:
This adds a favicon to our build. This is way better than the storybook icon we currently have.

EDIT: This breaks the build and should not be merged as of now.


raphaellueckl:
This has serious sideeffects. Closing this here, since I don't have a fast solution.

PR #1047, opened 2019-05-29
Feature/atom title section

jackblackCH:
Allows consumers to include a section title as a component.
i.e `<axa-title-section>Kundenbewertungen</axa-title-section>`

In many times as an app builder, the title of a section is staying outside of the next component.
That's why we deliver a title as a component that comes with all the needed stylings - out of the box.

An example of a independent section title and the following component (rating) from axa.ch
Screenshot 2019-05-29 at 22 31 53

### DIFF
--- a/src/components/10-atoms/title-section/README.md
+++ b/src/components/10-atoms/title-section/README.md
@@ -1,0 +1,68 @@
+# Title section
+
+This component comes with all needed styles for an AXA headline
+
+## Usage
+
+```bash
+npm install @axa-ch/title-section
+```
+
+```js
+import '@axa-ch/title-section';
+...
+<axa-title-section></axa-title-section>
+```
+
+### React
+
+Create a React-ified title-section with the createElement function from your React version and then use it like this:
+
+```js
+import { createElement } from 'react';
+import createAXATitleSectionReact from '@axa-ch/title-section/lib/index.react';
+
+const AXATitleSectionReact = createAXATitleSectionReact(createElement);
+
+export default AXATitleSectionReact;
+```
+
+```js
+<AXATitleSectionReact onClick={handler} />
+```
+
+### Pure HTML pages
+
+Import the title-section-defining script and use a title-section like this:
+
+```html
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <title>Your awesome title</title>
+  </head>
+  <body>
+    <axa-title-section></axa-title-section>
+    <script src="node_modules/@axa-ch/title-section/dist/index.js"></script>
+  </body>
+</html>
+```
+
+## Properties
+
+### Variant
+
+| Attribute       | Details         |
+| --------------- | --------------- |
+| `variant="foo"` | Desc of Variant |
+
+### Bar
+
+The attribute `bar` specifies...
+
+### onClick
+
+The function-valued attribute `onClick` can be used as a callback prop for React and other frameworks.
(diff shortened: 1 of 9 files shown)


src/components/10-atoms/title-section/index.js
```diff
@@ -0,0 +1,44 @@
...
+    this.variant = '';
+  }
+
+  render() {
+    const classes = {
+      'a-title-section--white': this.variant === 'white',
+    };
+
+    return html`
+      <h2 class="a-title-section ${classMap(classes)}">
```
LucaMele: why h2? and if h2, where is h1?

LucaMele: it should be `<h1>` and in the storybook examples title should be surrounded via a `article`

src/components/10-atoms/title-section/story.js
```diff
@@ -0,0 +1,17 @@
...
+storiesOf('Atoms/Title section', module)
+  .addDecorator(withMarkdown(Readme))
+  .add(
+    'Title section',
+    () => '<axa-title-section>This is a section title</axa-title-section>'
+  )
+  .add(
+    'Title section White',
+    () =>
+      `<div style="background: #00008f; padding: 50px; 100px;"><axa-title-section variant="white">This is the section title in white to be used on backgrounds</axa-title-section></div>`
```
LucaMele: use `article` here instead of div so that HTML also in examples is correct once `<axa-title-section >` renders h1

jackblackCH: done.

src/components/10-atoms/title-section/index.js
jackblackCH: It is a section title. Sections can exist mulitple times on a page, so it is mostly an H2 or H3.

H1 is a page title. Our story only previous the component itself. As a consumer you will have your h1 "somewhere"

LucaMele: Per HTML5 you can have a h1 per self contained block element like article and section therefore please use h1, even if user has it somewhere on the page

LucaMele: You are referring to HTML4, which we don’t support

LucaMele: https://html.spec.whatwg.org/multipage/sections.html#headings-and-sections

LucaMele: Quote: “Authors are also encouraged to explicitly wrap sections in elements of sectioning content, instead of relying on the implicit sections generated by having multiple headings in one element of sectioning content.”

Sectioning content: https://html.spec.whatwg.org/multipage/dom.html#sectioning-content-2

src/components/10-atoms/title-section/README.md
```diff
@@ -0,0 +1,64 @@
...
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <title>Your awesome title</title>
+  </head>
+  <body>
+    <axa-title-section></axa-title-section>
```
LucaMele: Please add Text here like in the example above: `<axa-title-section>This is a title for a section</axa-title-section>`

src/components/10-atoms/title-section/index.react.js
```diff
@@ -0,0 +1,14 @@
...
+import AXATitleSection from './index';
+
+export default createElement => ({
+  /* props here, same as in the constructor of index.js */
+  children,
+}) =>
+  withReact(createElement)(
+    AXATitleSection.tagName,
+    {
+      /* props here, same as in the constructor of index.js */
```
LucaMele: variant is missing here as attr

jackblackCH:
Tested with VoiceOver, Windows and MacOS. Heading Elements were found by voice over software.

src/components/10-atoms/title-section/index.react.js
jackblackCH: Nice. Done.

src/components/10-atoms/title-section/README.md
jackblackCH: Done.

PR #1050, opened 2019-06-03
[WIP] Feature/axa footer with childfragments

raphaellueckl:
This should once become an alternative to the model based footer approach.

TODO:
- ~~IE11: Fix removing empty elements from DOM~~
- ~~IE11: Fix title styling~~
- ~~Links in accordion should max to parent~~
- Documentation has to be completely rewritten with new approach
- Tests need to be adapted


jackblackCH:
As discussed. Closing this. Reopend in #1083

PR #1055, opened 2019-06-03
Title and Text as Components

jackblackCH:
As Discussed
- Add Two title components, providing out-of-the-box responsive titles as components.
- Add a text component, providing out-of-the-box responsive text as component
    - based on certain naming and architecture discussions in the team.
    - We challenged having seven separate atoms vs. two grouped by the typo primary and secondary
      - The size will can be adjusted by modifier/variant.
- Behind the curtain: typo styles are simplified again (approved by @TommyHoliday) 
- Title-section has merged into this (as discussed)


src/components/10-atoms/text/ui.test.js
```diff
@@ -0,0 +1,20 @@
...
+
+test('should render text', async t => {
+  const $axaElem = await Selector(TAG);
+  await t.expect($axaElem.exists).ok();
+  const $axaElemShadow = await Selector(
+    () => document.querySelector('axa-text').shadowRoot
+  );
+  const $axaElemShadowEl = await $axaElemShadow.find(CLASS);
+  await t.expect($axaElemShadowEl.exists).ok();
+});
```
raphaellueckl: Shouldn't you test for font-sizes, line-heights and what not here?

src/components/10-atoms/title-primary/index.js
```diff
@@ -0,0 +1,60 @@
...
+
+  render() {
+    // eslint-disable-next-line prefer-destructuring
+    const variant = this.variant;
+    const isSize1 = variant.includes('size-1');
+    const isSize2 = variant.includes('size-2');
+    const isSize3 = variant.includes('size-3');
+    const isSize4 = variant.includes('size-4');
+    const isSize5 = variant.includes('size-5');
+    const isSize6 = variant.includes('size-6');
```
raphaellueckl: I would make the sized mutually exclusive, so that one text can't be all the sizes at once. A switch statement would be appropriate imho.

src/components/10-atoms/title-primary/index.react.d.ts
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+type Variant = 'size-2' | 'size-3' | 'size-4' | 'size-5' | 'size-6';
```
raphaellueckl: What about `size-1`?

src/components/10-atoms/title-secondary/index.js
```diff
@@ -0,0 +1,60 @@
...
+
+  render() {
+    // eslint-disable-next-line prefer-destructuring
+    const variant = this.variant;
+    const isSize1 = variant.includes('size-1');
+    const isSize2 = variant.includes('size-2');
+    const isSize3 = variant.includes('size-3');
+    const isSize4 = variant.includes('size-4');
+    const isSize5 = variant.includes('size-5');
+    const isSize6 = variant.includes('size-6');
```
raphaellueckl: Should imho also be mutually exclusive.

src/components/10-atoms/title-primary/index.react.d.ts
jackblackCH: Size 1 is the default.

src/components/10-atoms/text/ui.test.js
jackblackCH: Hmm. Yes. You are #@# right.

src/components/10-atoms/title-primary/index.js
jackblackCH: I disagree. Variants are modifiers that can and should be able to be combined. Indeed for sizes, it does not make too much sense. Still, we should shot not prevent any "users" fault. We should show the modifiers that a user puts on it. Preventing it results also in even more unexpected behavior.

src/components/10-atoms/text/ui.test.js
jackblackCH: I added tons of responsive tests that cover font size check, based on our design rules. So this should prevent us from surprised in components by changing font sizes.

src/components/10-atoms/title-primary/index.react.d.ts
jackblackCH: I removed size one from the docs.

src/components/10-atoms/title-secondary/index.js
jackblackCH: dito. top.

PR #1065, opened 2019-06-06
feature/checkbox-focus-blur

### LINKED ISSUE #1041: AXACheckbox: add onFocus and onBlur handlers
Add `onFocus` and `onBlur` handlers to `AXACheckbox` component.

These are needed by form frameworks which evaluate whether a field was touched and so whether to do validation, enable submit button, etc...

MKaHead:
fixes #1041

- Add blur and focus to checkbox
- Update readme and all boilerplate
- Improve stories and demos


src/components/10-atoms/checkbox/index.react.d.ts
```diff
@@ -8,6 +8,8 @@ interface AXACheckboxProps {
   disabled?: boolean;
   error?: string;
   onChange?: (value: any) => void;
+  onFocus?: () => void;
```
MKaHead: @MarekLacoAXA do you want value: any here, too?

raphaellueckl: Why any? Can't it be `Function` or something like that?

src/components/10-atoms/checkbox/index.react.d.ts
```diff
@@ -8,6 +8,8 @@ interface AXACheckboxProps {
   disabled?: boolean;
```
raphaellueckl: AFAIK, the lowercase types belong to javascript and the uppercase ones to typescript.

MarekLacoAXA: no, no.

PR #1077, opened 2019-06-11
Feature/axa dropdown formready

### LINKED ISSUE #1111: Datepicker: unusable in react
When using as 'inputfield' variant: 

Datepicker component doesn't behave like 'controlled' which is react standard:

Datepicker doesn't have "value" prop, so it can't display the pre/selected date.

Datepicker doesn't survive re-render. So in practice every click on Date causes dismiss of the opened calendar (*), and the input shows no date selected.

*) onChange event in Datepicker causes react state change which causes parent component rerender which rerenders the Datepicker, which appears as blank ==> buggy behavior

Another problem:
onAXADateChange cb is called on first render (seems that with the "Monday" day of current week)

markus-walther:
This actually includes both dropdown and datepicker, because the latter includes the former.

IE 11 tests have not yet been done.

All tests work.

Branch has been rebased on latest develop.

Dropdown has extensive refactoring, datepicker less so.

Constructive comments welcome, bring them on!

fixes #1111


jackblackCH:
There is now a react/demo.jsx and react/story.jsx, can we merge them? or what is the difference?

jackblackCH:
I suggest we explain our thoughts in the README and not in the storzy it self. i.e this within a description to the readme `With the React wrapper - controlled <axa-datepicker> component with callback prop and date-error validation!`

jackblackCH:
Git commits should start with an uppercase

src/components/00-materials/icons/filled-tick-animated.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="166" height="151" viewBox="-20 -10 210 170" style="overflow:visible"><path d="M.3 96l62.4 54.1L165.6.3" stroke-linejoin="round" style="fill:none;stroke:#1cc54e;stroke-miterlimit:10;stroke-width:40px;margin:40px;stroke-dasharray:500;stroke-dashoffset:500;animation:filled-tick-animated 1.2s linear forwards"/></svg>
```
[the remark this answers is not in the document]
markus-walther: done

src/components/00-materials/icons/filled-tick.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 69.99 59.97" width="100%" height="100%">
```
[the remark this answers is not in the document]
markus-walther: ditto

src/components/20-molecules/datepicker/README.md
```diff
@@ -1,31 +1,22 @@
...
-- Due to limitations of ShadowDom, sent input values won't be visible within native form data (WIP)
-
 ## Install
 
 ```bash
 npm install @axa-ch/datepicker
 ```
 
-## Usage Native
+## Native Example Usage
```
raphaellueckl: Nice!

src/components/20-molecules/datepicker/index.scss
```diff
@@ -26,17 +42,23 @@
...
-    margin-bottom: 1px;
-    border: solid 1px $color-prim-gray-mercury;
     width: 197px;
   }
 
   &__input-button {
     background: $color-prim-white;
-    cursor: pointer;
-    border: 0;
+    cursor: pointer;border: 1px solid;
```
raphaellueckl: forgotten line break

src/components/20-molecules/datepicker/index.scss
```diff
@@ -26,17 +42,23 @@
...
+    border-color: $color-prim-gray-mercury;
+    border-left: none;
     line-height: 0;
     padding: 1px 3px 2px;
+    margin-left: 0; /* suppress Safari 2px margin */
```
raphaellueckl: AFAIK such comments are not wanted by the team. Comments about the user agent style sheets that gets overridden.

src/components/20-molecules/datepicker/react/demo.jsx
```diff
@@ -0,0 +1,15 @@
...
+import ReactDOM from 'react-dom';
+import DemoControlledDatepickerReact from './DemoDatepickerReact';
+import { withMarkdown } from '../../../../../.storybook/addons/markdown';
+import Readme from '../README.md';
+
+storiesOf('Molecules/Datepicker/React/Demos', module)
+  .addDecorator(withMarkdown(Readme))
+  .add('Controlled-component React-ified Datepicker with inputfield', () => {
+    const div = document.createElement('div');
+    ReactDOM.render(<DemoControlledDatepickerReact />, div);
```
raphaellueckl: Just question, since this is handled differently anyways in our codebase:

We could make up the example in here instead of just calling the prepared demo-component. Since in our documentation, we show the user how to use skatejs and tell him to create a STATELESS react component and then use it everywhere and providing the state from the outside.

Meaning: The `DemoControlledDatepickerReact` component should only be what we suggest as skatejs wrapper, and the rest should be in here within the story, including the props, childfragments, etc

What do you think?

src/components/20-molecules/datepicker/react/story.jsx
```diff
@@ -38,6 +38,7 @@ storiesOf('Molecules/Datepicker/React', module)
         labelbuttoncancel="Cancel"
         labelbuttonok="OK"
         allowedyears={[2019, 2020]}
+        // eslint-disable-next-line no-console
```
raphaellueckl: I don't like console.log statements in stories, since a user probably does not have the console open when he is testing our demos. IMHO 'alert' is the least we should provide, but I even better like a div containing some state text, that changes upon events that happen.

src/components/20-molecules/datepicker/story.js
```diff
@@ -20,7 +19,7 @@ storiesOf('Molecules/Datepicker', module)
     'Datepicker Input',
     () => `
       <axa-datepicker inputfield year="2019" month="0" day="1" data-test-id="datepicker-input-2019"></axa-datepicker>
-      <axa-datepicker inputfield year="2020" month="1" day="2" data-test-id="datepicker-input-2020"></axa-datepicker>
-      <axa-datepicker inputfield year="2021" month="2" day="3" data-test-id="datepicker-input-2021"></axa-datepicker>
+      <axa-datepicker inputfield placeholder="Bitte Datum wählen" year="2020" month="1" day="2" data-test-id="datepicker-input-2020" invaliddatetext="Ungültiges Datum"></axa-datepicker>
```
raphaellueckl: We should keep the documenation in english, including all values.

src/components/20-molecules/datepicker/ui.test.js
```diff
@@ -155,3 +162,72 @@ test('should render datepicker as reactified component', async t => {
...
+  );
+  await t.expect(datepickerForm.exists).ok();
+  await t.typeText(
+    `axa-datepicker[data-test-id="datepicker-forms"] .js-datepicker__input`,
+    '29.2.2020'
+  );
+  await t.click('#datepicker-forms-submit');
+  await t
+    .wait(50)
+    .expect((await Selector('#datepicker-forms-content')).innerText)
```
raphaellueckl: Is this wait statement really needed? Testcafe anyway has a build in timeout to wait for elements to appear.

src/components/20-molecules/datepicker/utils/date.js
```diff
@@ -1,7 +1,9 @@
 export const getStartOfWeek = date => {
   const iDayOfWeek = date.getDay();
   const iDifference = date.getDate() - iDayOfWeek + (iDayOfWeek === 0 ? -6 : 1);
-  return new Date(date.setDate(iDifference));
+  const newDate = new Date(date);
+  newDate.setDate(iDifference);
```
raphaellueckl: What does the `i` stand for?

src/components/20-molecules/datepicker/utils/date.js
```diff
@@ -16,55 +18,74 @@ const getWeekdays = (date, locale) => {
...
 };
 
-const ALL_DATE_SEPERATORS = / |,|\.|-|\//;
+const ALL_DATE_SEPARATORS = / |,|\.|-|\//;
+
 const clearStringFromIEGeneratedCharacters = string =>
   string.replace(/[^\x00-\x7F]/g, ''); // eslint-disable-line no-control-regex
+
+const zeroFill = (number, width) => {
+  const _number = number | 0; // coerce number to integer
```
raphaellueckl: Using underscore here is confusing, since in javascript, the underscore is used for private methods and properties. This here is neither, it's a local variable.

src/components/20-molecules/datepicker/utils/date.js
```diff
@@ -16,55 +18,74 @@ const getWeekdays = (date, locale) => {
...
+  const n_ = Math.abs(_number);
+  const zeros = Math.max(0, width - Math.floor(n_).toString().length);
+  const zeroString = (10 ** zeros).toString().slice(1);
+  return `${_number < 0 ? '-' : ''}${zeroString}${_number}`;
+};
```
raphaellueckl: Understanding what this method does is a pain in the ass. What does `zeroFill` even mean? I would add a more declarative method name.

src/components/20-molecules/dropdown/react/DemoDropdownReact.jsx
```diff
@@ -0,0 +1,66 @@
+import React, { useState } from 'react';
+import AXADropdownReact from './AXADropdownReact';
+import AXACheckboxReact from '../../../10-atoms/checkbox/react/AXACheckboxReact';
+
+const DemoDropdown = () => {
```
raphaellueckl: As mentioned before, I would only have the wrapper in here and do all the rest with those arrays and properties outside of here, if possible. Because the user won't use the wrapper like this, he will keep it generic.

src/components/20-molecules/dropdown/react/story.jsx
```diff
@@ -2,32 +2,14 @@
...
-          },
-          { name: 'Item 1', value: 'Item 1', url: '#', isSelected: false },
-          { name: 'Item 2', value: 'Item 2', url: '#', isSelected: false },
-          { name: 'Item 3', value: 'Item 3', url: '#', isSelected: false },
-        ]}
-        onAXAValueChange={value => console.log(`value changed ${value}`)}
-      />,
-      div
-    );
+    ReactDOM.render(<DemoDropdownReact />, div);
```
raphaellueckl: Oh, now I see, that you did this on purpose. ö.ö

Would be worth a discussion, maybe you have a good reason to do that. :) Because all other stories and demos aren't handled like that.

src/components/20-molecules/dropdown/ui.test.js
```diff
@@ -11,13 +11,127 @@ test('should render correctly', async t => {
...
+
+  const getErrorMessage = ClientFunction(() => {
+    const errorDiv = document.querySelector(
+      'axa-dropdown[data-test-id="dropdown-error"] > .m-dropdown__error'
+    );
+    return errorDiv.innerText;
+  });
+
+  await t
+    .wait(50)
```
raphaellueckl: You use waits almost everywhere. I don't like it. If not explicitly needed, I would propose to not use them, because 20 waits like that will already make the build one second slower, and probably for no reason.

src/components/20-molecules/datepicker/index.scss
markus-walther: done

src/components/20-molecules/datepicker/index.scss
markus-walther: I just checked with Luca: that is not the position of the team. Mere logic would suggest that one needs to know whether an otherwise unneeded CSS property is imposed just to fix a browser quirk or bug. That is the case here. If I don't comment, we have information loss - so I comment. Hope that makes sense to you.

src/components/20-molecules/datepicker/react/demo.jsx
markus-walther: It's not handled differently in our codebase, the approach was suggested by Marco, see his Button for more examples. It looks clean and simple.

src/components/20-molecules/datepicker/story.js
markus-walther: Definitely a no from me here. We should use realistic values from the domain we develop for, which is AXA.

src/components/20-molecules/datepicker/ui.test.js
markus-walther: Good question. After the click logic runs which takes time, the innerText will not appear instantaneously. That motivates the wait - note that it's not about the element which needs to exist, but its text which has to have a certain expected value. With that background info, is it OK for you?

src/components/20-molecules/datepicker/utils/date.js
markus-walther: I suspect index or integer, but it's code that I did not touch.

I changed it to `dayOfWeek` and `daysSinceBeginningOfMonth`, thanks for the feedback!

src/components/20-molecules/datepicker/utils/date.js
markus-walther: Hmm, at best it's a practice by some people, not anything enforced by JS syntax/semantics.

But it's been deleted anyway by refactoring zeroFill below.

src/components/20-molecules/datepicker/utils/date.js
markus-walther: Fair point. I could have added the stackoverflow URL...

Instead I refactored it and named it `addLeadingZeroes`.

src/components/20-molecules/dropdown/react/DemoDropdownReact.jsx
markus-walther: See my feedback before.

src/components/20-molecules/dropdown/react/story.jsx
markus-walther: `git grep 'console.log' | grep story | wc -l` gives 12 occurrences, so not true.

src/components/20-molecules/dropdown/ui.test.js
markus-walther: Not true. I use them when there is a reason. You could leigitimately ask for that reason, though - that would be a fair point.

src/components/20-molecules/datepicker/ui.test.js
markus-walther: I will insert a comment to that effect into the code

src/components/20-molecules/dropdown/react/story.jsx
markus-walther: OK, I see your point and will see what I can do here.

src/components/20-molecules/dropdown/ui.test.js
markus-walther: OK, will document reasons for wait as comments everywhere - or remove if no such reason can be found.

src/components/20-molecules/datepicker/react/story.jsx
markus-walther: Deleted since underlying change callback was problematic in triggering behaviour - thanks for spotting this!

src/components/20-molecules/dropdown/ui.test.js
markus-walther: done

src/components/20-molecules/datepicker/ui.test.js
markus-walther: done

LucaMele:
Tests are failing and the tick icon is not visible at all even if dropdown reserves the space

PR #1078, opened 2019-06-12
Fix table crashing on update

LucaMele:


src/components/30-organisms/table-sortable/index.js
```diff
@@ -53,16 +53,21 @@ class AXATableSortable extends LitElement {
...
       },
+      tfoot
     } = this.model;
-
-    return !(theadL !== tbodyL && (!tfootL || tfootL === tbodyL));
+    let tfootL = 0;
+    if (tfoot && tfoot[0]) {
+      tfootL = tfoot.length;
+    }
+    return !tbodyL || !((theadL !== tbodyL) && (!tfootL || tfootL === tbodyL));
```
raphaellueckl: This shit is complicated as fuck. Why not move it to a method with a proper name (that's what we would do in the java context, going all for readability)? Because understanding what this thing is doing is very hard.

LucaMele: It is in a method with proper name: `validateModel() {}`

LucaMele:
```js
validateModel() {
    const {
      thead: {
        length: theadL
      },
      tbody: {
        0: {
          length: tbodyL
        }
      },
      tfoot
    } = this.model;
    let tfootL = 0;
    if (tfoot && tfoot[0]) {
      tfootL = tfoot.length;
    }
    return !tbodyL || !((theadL !== tbodyL) && (!tfootL || tfootL === tbodyL));
  }
```

LucaMele: i try to rename the variables instead

PR #1081, opened 2019-06-12
Feature/improve icons

MKaHead:
- Improve icon consistency - all icons now have 24x24 px.
- Icon component - use as default size for the icons 16x16 px.


src/components/10-atoms/icon/index.js
```diff
@@ -40,6 +47,7 @@ class AXAIcon extends LitElement {
   static get properties() {
     return {
       icon: { type: String },
+      size: { type: String, reflect: true },
```
raphaellueckl: Reflect doesn't make sense here imho.

PR #1083, opened 2019-06-13
Feature/footer v3

raphaellueckl:
The axa footer


src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,279 @@
...
+      .forEach(el => {
+        const label = el.querySelector('slot').assignedNodes()[0];
+        // Second part of IF-statement is an IE11 workaround
+        if (this._listElementHasNoContent(label)) {
+          el.style.display = 'none';
+        }
+      });
+  }
+
+  _listElementHasNoContent(label) {
```
markus-walther: All helper methods without `this` don't belong into a class in my opinion - move to arrow function above class.

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,279 @@
...
+
+  static get styles() {
+    return css`
+      ${unsafeCSS(styles)}
+    `;
+  }
+
+  static get properties() {
+    return {
+      dynamic: { type: Boolean },
```
markus-walther: What does dynamic mean? 1 line of comment would help a lot for the rest of the file...

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,279 @@
...
+  updated() {
+    this._removeEmptyListElements();
+  }
+
+  _removeEmptyListElements() {
+    this.shadowRoot
+      .querySelectorAll('.o-footer__main-content-panel-list-item')
+      .forEach(el => {
+        const label = el.querySelector('slot').assignedNodes()[0];
+        // Second part of IF-statement is an IE11 workaround
```
markus-walther: workaround for what bug? If comment explains it, the reader could learn to avoid the bug in their own code...!

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,279 @@
...
+            detail: href,
+            bubbles: true,
+            cancelable: true,
+          })
+        );
+      }
+    }
+  };
+
+  _extractNestedHref(ev) {
```
markus-walther: I don't think 'Nested' is the right word, since you presumably don't have 'href-in-href'...

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,279 @@
...
+    let element;
+    if (!ev.target || !ev.target.href) {
```
markus-walther: This looks super-complicated and cries out for simplification and comments. I guess at least a for-loop would help, and object destructuring.

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,279 @@
...
+        else return undefined;
+      }
+    }
+    return element === undefined ? ev.target.href : element.href;
+  }
+
+  _handleAccordionClick = index => {
+    this._accordionActiveIndex =
+      index === this._accordionActiveIndex ? -1 : index;
+    this.requestUpdate();
```
markus-walther: if _accordionActiveIndex were a declared property, wouldn't the manual requestUpdate be superfluous? The latter one should be used sparingly at best, because it's a procedural breakaway from the nicely declarative style of lit-element....

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,279 @@
...
+                  name="column-0-title-desktop"
+                  class="o-footer__title-desktop"
+                ></slot>
+                <button
+                  class="o-footer__accordion-button"
+                  @click="${() => this._handleAccordionClick(0)}"
+                >
+                  <slot name="column-0-title" class="o-footer__title"></slot>
+                  <span
+                    class="o-footer__accordion-button-caret o-footer__accordion-button-caret${this
```
markus-walther: replace this with proper use of `classMap`! It's error-prone to construct multi-class strings manually.

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,279 @@
...
+                      ._accordionActiveIndex === 0
+                      ? '--open'
+                      : ''}"
+                  >
+                    ${svg([CaretSvg || ''])}
```
markus-walther: This expression should be lifted to the beginning of render() and given a proper name

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,279 @@
...
+                  </span>
+                </button>
+                <ul
+                  class="o-footer__main-content-panel o-footer__main-content-panel${this
```
markus-walther: ditto for `classMap`

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,279 @@
...
+                    ._accordionActiveIndex === 0
+                    ? '--open'
+                    : ''}"
+                >
+                  <li class="o-footer__main-content-panel-list-item">
```
markus-walther: uh, use `map` or `repeat` to avoid this major violation of DRY...

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,279 @@
...
+                  name="column-1-title-desktop"
+                  class="o-footer__title-desktop"
+                ></slot>
+                <button
+                  class="o-footer__accordion-button"
+                  @click="${() => this._handleAccordionClick(1)}"
+                >
+                  <slot name="column-1-title" class="o-footer__title"></slot>
+                  <span
+                    class="o-footer__accordion-button-caret o-footer__accordion-button-caret${this
```
markus-walther: ditto for `classMap`

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,279 @@
...
+                      ._accordionActiveIndex === 1
+                      ? '--open'
+                      : ''}"
+                  >
+                    ${svg([CaretSvg || ''])}
+                  </span>
+                </button>
+                <ul
+                  class="o-footer__main-content-panel o-footer__main-content-panel--short o-footer__main-content-panel${this
```
markus-walther: ditto for `classMap`

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,279 @@
...
+                    ._accordionActiveIndex === 1
+                    ? '--open'
+                    : ''}"
+                >
+                  <li class="o-footer__main-content-panel-list-item">
```
markus-walther: ditto for `map`/`repeat`

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,279 @@
...
+                </ul>
+              </div>
+            </div>
+            <div class="o-footer__social-media">
+              <slot
+                name="column-2-social-title"
+                class="o-footer__social-media-title"
+              ></slot>
+              <ul class="o-footer__social-media-list">
+                <li class="o-footer__social-media-item">
```
markus-walther: ditto for `map`/`repeat` - I think you should make a function factory since you have several slightly different versions of this same pattern

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,279 @@
...
+                  @click="${() => this._handleAccordionClick(1)}"
+                >
+                  <slot name="column-1-title" class="o-footer__title"></slot>
+                  <span
+                    class="o-footer__accordion-button-caret o-footer__accordion-button-caret${this
+                      ._accordionActiveIndex === 1
+                      ? '--open'
+                      : ''}"
+                  >
+                    ${svg([CaretSvg || ''])}
```
markus-walther: ditto for extracting expression, which - since repeated here - now also violates DRY

src/components/30-organisms/footer/index.scss
```diff
@@ -0,0 +1,314 @@
+.o-footer {
+  text-align: initial;
+
+  color: $color-prim-white;
+  background-color: $color-prim-blue-dark-indigo;
+
+  border-top: 1px solid rgba(255, 255, 255, 0.25);
```
markus-walther: undeclared color, shouldn't this be added to global colors after discussion with designers?

src/components/30-organisms/footer/index.scss
```diff
@@ -0,0 +1,314 @@
...
+    padding: 10px;
+  }
+
+  ::slotted(a) {
+    padding: 10px;
+  }
+
+  &__content {
+    @include breakpoint($mediaquery-sm-up) {
+      display: grid;
```
markus-walther: nice!

src/components/30-organisms/footer/index.scss
```diff
@@ -0,0 +1,314 @@
...
+
+        ::slotted(a) {
+          display: block;
+          text-decoration: none;
+          color: $color-prim-white;
+          padding: 20px 30px !important; // Needed to dominate body reset
+        }
+
+        ::slotted(a:hover) {
+          color: rgba(255, 255, 255, 0.65);
```
markus-walther: undeclared color, shouldn't this be added to global colors after discussion with designers?

src/components/30-organisms/footer/index.scss
```diff
@@ -0,0 +1,314 @@
...
+        &--short {
+          width: 165px;
+        }
+
+        &-list-item {
+          @include typo(primary, default, default);
+
+          width: 165px;
+          margin-top: 8px;
+          padding: 5px 0 !important;
```
markus-walther: why this !important? If absolutely unavoidable, give reason in comment

src/components/30-organisms/footer/react/story.jsx
```diff
@@ -0,0 +1,196 @@
+import AXAFooterReact from './Footer';
+
+import { storiesOf } from '@storybook/html';
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { withMarkdown } from '../../../../../.storybook/addons/markdown';
+import Readme from '../README.md';
+// import withBodyReset from '../../../../../.storybook/addons/reset-body';
+
+function setNewlyClickedLink(link) {
```
markus-walther: Arrow function is more idiomatic in our code

src/components/30-organisms/footer/ui.test.js
```diff
@@ -0,0 +1,236 @@
...
+
+  await t.expect($contactLink.textContent).eql('Contact');
+  await t.expect($contactLink.visible).ok();
+
+  const $result = Selector('#clicked-link');
+  await t.expect($result.innerText).contains(' -');
+
+  await t.click($contactLink);
+
+  // Travis cannot handle the following block of code, for whatever reason.
```
markus-walther: I wouldn't leave large blocks of commented-out code in, that's what version control is there  for

src/components/30-organisms/footer/ui.test.js
```diff
@@ -0,0 +1,236 @@
...
+  // await t.expect($axaWorldwideLink.visible).ok();
+  // await t.expect($axaWorldwideLink.textContent).eql('axa & you');
+
+  const $facebookButton = FooterAccessor.getSlotNode('column-social-item-0');
+
+  await t.expect($facebookButton.visible).ok();
+  await t.click($facebookButton);
+
+  await t.expect($result.innerText).contains('https://www.facebook.com/axach/');
+});
```
markus-walther: Nice + comprehensive tests!

src/components/30-organisms/footer/story.js
```diff
@@ -0,0 +1,48 @@
+/* global document */
+// TODO fix that stuff
+/* eslint-disable import/no-extraneous-dependencies */
+import { storiesOf } from '@storybook/html';
+import './index';
+import { withMarkdown } from '../../../../.storybook/addons/markdown';
+// import withBodyReset from '../../../../.storybook/addons/reset-body';
```
markus-walther: don't leave commented-out code in, version control handles that

src/components/30-organisms/footer/demo.js
```diff
@@ -0,0 +1,62 @@
...
+        <svg xmlns="http://www.w3.org/2000/svg" width="25" height="25" viewBox="0 0 25 25"><path d="M18.797 4.185c-3.25-.216-10.469-.216-13.719 0-3.516.23-3.922 2.323-3.953 7.815-.031 5.492.438 7.585 3.953 7.815 3.234.216 10.469.216 13.719 0 3.515-.23 3.922-2.323 3.953-7.815.031-5.492-.438-7.585-3.953-7.815zM9.234 15.554V8.446L16.438 12l-7.204 3.554z" fill="#FFFFFF"/></svg>
+      </a>
+      <a slot="column-social-item-5" href="https://www.linkedin.com/company/axa/" target="_blank">
+      <svg xmlns="http://www.w3.org/2000/svg" width="25" height="25" viewBox="0 0 25 25"><path d="M6.519 4.183A2.146 2.146 0 0 1 4.354 6.3a2.147 2.147 0 0 1-2.121-2.16A2.147 2.147 0 0 1 4.376 2a2.14 2.14 0 0 1 1.527.641c.403.41.625.966.616 1.542zm0 3.9H2.216V22h4.32L6.519 8.083zm6.878 0h-4.27V22h4.287v-7.3c0-4.05 5.2-4.383 5.2 0V22h4.303v-8.817c0-6.833-7.692-6.666-9.503-3.216l-.017-1.884z" fill="#FFFFFF"/></svg>
+      </a>
+    </axa-footer>`;
+
+    wrapper.innerHTML = footerMarkup;
+
+    const lastClicked = document.createElement('span');
```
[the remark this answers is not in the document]
raphaellueckl: I would like to keep especially those things simple, because a user might look into the sourcecode and search for the "way of least resistance" to bring something to work. There I would keep it as "accessible" as possible.

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,279 @@
...
+  }
+
+  constructor() {
+    super();
+    this.dynamic = false;
+    this.onItemClick = () => {}; // Only for react
+    this._accordionActiveIndex = -1;
+  }
+
+  updated() {
```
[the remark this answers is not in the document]
raphaellueckl: Good point!

src/components/30-organisms/footer/index.js
raphaellueckl: Why that?
What about making them static? You probably have an issue with the performance aspect of that, right? `static` would resolve the performance problems.

The underscore notation was explicitly created to have private methods and properties. It would make me wonder if that is considered bad practice.

raphaellueckl: But I am not sure. Maybe your approach would be more appropriate. Sounds somehow logical, but I never heard of a pattern like that.

src/components/30-organisms/footer/index.js
raphaellueckl: Hm, it is explained in the readme. We use this terminology also in other places. But I agree, it's not that readable. But before we start writing comments everywhere, I would vote, that we find another name that is more readable.

src/components/30-organisms/footer/index.scss
raphaellueckl: Good find, I probably forgot to clean up. :$

src/components/30-organisms/footer/index.js
raphaellueckl: I went with the static approach, please have a look. :) Nice feedback!

src/components/30-organisms/footer/ui.test.js
raphaellueckl: Right, usually that's my feedback. Should also apply to myself. :)

src/components/30-organisms/footer/index.js
LucaMele: What with: preventdefault ? which is what natively happens

src/components/30-organisms/footer/demo.js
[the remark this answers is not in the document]
markus-walther: OK, your choice here, Marco and I used this way in other components, that's why I suggested it.

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,279 @@
...
+  static get properties() {
+    return {
+      dynamic: { type: Boolean },
+    };
+  }
+
+  constructor() {
+    super();
+    this.dynamic = false;
+    this.onItemClick = () => {}; // Only for react
```
LucaMele: `// Only for react` -> not true, all those languages having functions as callbacks.

src/components/30-organisms/footer/index.js
markus-walther: Writing comments is good, I requested 1 comment, not "everywhere". Just do it :-)

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,279 @@
...
+    `;
+  }
+
+  updated() {
+    this._removeEmptyListElements();
+  }
+
+  _removeEmptyListElements() {
+    this.shadowRoot
+      .querySelectorAll('.o-footer__main-content-panel-list-item')
```
LucaMele: never use style classes for JS selection. If u want to change style or class name on style u need to refactor as well JS. Please add another class named like so: `.js-footer__main-content-panel-list-item` for example. We always prefix with `js` for dom modifications

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,279 @@
...
+      .forEach(el => {
```
LucaMele: Does this worrk in IE and EDGE? pretty sure they cant handle DOM NODE List as ARRAY

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,279 @@
...
+        const label = el.querySelector('slot').assignedNodes()[0];
```
LucaMele: https://developer.mozilla.org/de/docs/Web/API/HTMLSlotElement/assignedNodes
Assigned nodes seems not to be IE compatible. Does one of our polyfill take care of it?

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,279 @@
...
+            bubbles: true,
+            cancelable: true,
+          })
+        );
+      }
+    }
+  };
+
+  static _listElementHasNoContent(label) {
+    return !label || label.toString() === '[object Text]';
```
LucaMele: Please do not relay on toString method instead use node Types to identify the type.

Reason: Type is more reliable. toString can be overridden or syntax of the to string methods could slightly change between browsers

REF: https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType

src/components/30-organisms/footer/index.js
markus-walther: There is some confusion here in my opinion.

(Arrow) functions outside of class scope is just basic JS scoping rules, no 'best practice' needed at all. You have a module scope, use it!

Everything in a class that's non-static is added to the instance and thus consumes memory per  *instance*, so basic perf consideration.

"The underscore notation was explicitly created to have private methods and properties": citation needed.
There is nothing private here, just a convention by some people. You don't have privacy in JS.

And: I recommend to not use "I never heard of X".

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,279 @@
...
+  }
+
+  static _extractNestedHref(ev) {
+    let element;
+    if (!ev.target || !ev.target.href) {
```
LucaMele: Please use deconstruction. Doing `ev.target` twice is duplication of code. Is like in JAVA terms calling the getter method of a class twice instead of storing it locally.

Something like so:

```js
static _extractNestedHref({target}) {
    let element;
    if (!target || !target.href) {
       ...
    }
```

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,279 @@
...
+      element = ev;
```
LucaMele: why does an event become a element?

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,279 @@
...
+      while (!element.href) {
+        if (element.target && element.target.parentNode)
```
LucaMele: again, use target directly from my example abve.

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,279 @@
...
+          element = element.target.parentNode;
+        else if (element.parentNode) element = element.parentNode;
```
LucaMele: use always brackets please. It should be in our lint rules as well

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,279 @@
...
+        else return undefined;
+      }
+    }
+    return element === undefined ? ev.target.href : element.href;
+  }
```
LucaMele: Here how i would get the params for this function:

```js
static _extractNestedHref({target: { href, parentNode } }) {
 ...
  }
// OR
static _extractNestedHref(ev) {
const {target: { href, parentNode } } = ev;
 ...
  }
```

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,279 @@
...
+
+  static _extractNestedHref(ev) {
+    let element;
+    if (!ev.target || !ev.target.href) {
+      element = ev;
+      while (!element.href) {
+        if (element.target && element.target.parentNode)
+          element = element.target.parentNode;
+        else if (element.parentNode) element = element.parentNode;
+        else return undefined;
```
LucaMele: return undefined? Sounds hacky. I would refactor the whole method

src/components/30-organisms/footer/index.scss
```diff
@@ -0,0 +1,309 @@
...
+  @include breakpoint($mediaquery-sm-up) {
+    padding: 40px 0;
+  }
+
+  @include breakpoint($mediaquery-lg-up) {
+    padding: 50px 0;
+  }
+
+  &__main {
+    ::slotted(h2) {
```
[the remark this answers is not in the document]
LucaMele: Plus, is still a "working draft" technology. I would inject CSS into the dom with a style tag (like cookie disclaimer).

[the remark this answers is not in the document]
LucaMele: Lets discuss it anway. 

But as example something like so:

```js
      // Necessary to inject in the local dom a style tag for resetting
      // the p tag as its is within 2 shadow roots (axa-cookie-disclaimer and axa-container).
      // Property ::slotted has not beign used due to IE and EDGE compatibility
      // and its standard-ification is still "Working Draft" the 21.06.2019
      const resetStyle = document.createElement('style');
      resetStyle.textContent = yourStyleImportedFromAnyDotSCSS;
      this.appendChild(resetStyle);
    }
```

src/components/30-organisms/footer/index.scss
```diff
@@ -0,0 +1,309 @@
...
+      @include breakpoint($mediaquery-sm-up) {
+        display: none; // mobile title
+      }
+
+      padding: 0;
+      margin: 0;
+    }
+  }
+
+  ::slotted(h2) {
```
LucaMele: dtto slotted

src/components/30-organisms/footer/index.scss
```diff
@@ -0,0 +1,309 @@
...
+  }
+
+  .o-footer__social-media-item ::slotted(a) {
+    display: block;
+    width: 25px;
+    height: 25px;
+    padding: 10px;
+  }
+
+  ::slotted(a) {
```
LucaMele: dtto slotted

src/components/30-organisms/footer/story.js
LucaMele: Why no nobodyReset?? Footer would be broken without

src/components/30-organisms/footer/story.js
```diff
@@ -0,0 +1,48 @@
...
+import Readme from './README.md';
+
+storiesOf('Organisms/Footer', module)
+  .addDecorator(withMarkdown(Readme))
+  // .addDecorator(withBodyReset())
```
LucaMele: why no bodyReset? this brakes footer as it is a 100% with component

src/components/30-organisms/footer/index.js
raphaellueckl: After the previous discussion, I will go with your approach.

src/components/30-organisms/footer/index.js
raphaellueckl: @LucaMele I don't like "preventDefault" in this context. Using "native terminology" to express functionality dedicated to your component does not sound good to me. Example: Someone reads this and thinks "hmmm preventdefault, which event gets prevented?" It opens unnecessary other questions. In the end, in this case, ONLY clickevents get prevented. Not all events.

@markus-walther The fact, that you noticed it and were confused tells me, that we should find something better. Having an ugly property name and use comments and readmes to explain it is IMHO bad, if we could agree on a better name. How about a boolean with the name `clickevent`?

package.json
```diff
@@ -20,6 +20,8 @@
     "test-chrome": "sh scripts/testcafe-chrome.sh",
     "test-ie": "sh scripts/testcafe-ie.sh",
     "test-jest": "node scripts/test.unit.js --env=jsdom",
+    "test-ui": "TEST_HOST_STORYBOOK_URL=http://localhost:6006 testcafe chrome -c 4 ./**/ui.test.js",
+    "test-ui-hl": "TEST_HOST_STORYBOOK_URL=http://localhost:6006 testcafe 'chrome:headless' -c 4 ./**/ui.test.js",
```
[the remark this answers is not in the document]
raphaellueckl: Good point! The new ones I introduced make faster testing possible. Usually, during development of a feature, an instance of storybook is already running, so those commands will only execute the tests without also starting another storybook instance.

But from here, we probably can get rid of `test-chrome`.

src/components/10-atoms/input-file/demo.js
```diff
@@ -17,8 +17,6 @@ storiesOf('Atoms/Input File/Demos', module)
       inputFile.querySelector('.a-input-file').innerHTML = `
         <span class="a-input-file__flex-wrapper">You have chosen ${counter} files</span>
       `;
-      // eslint-disable-next-line no-console
-      console.log('counter!!', counter);
```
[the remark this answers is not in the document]
raphaellueckl: Hm, if I remember correctly, I had eslint issues and thought that's a bad approach. My bad, I think I discussed with marco and he said he is going to refactor it anyways or whatever, but I definitely should not just delete it.

src/components/30-organisms/footer/demo.js
```diff
@@ -0,0 +1,60 @@
...
+import Readme from './README.md';
+
+storiesOf('Organisms/Footer/Demos', module)
+  .addDecorator(withMarkdown(Readme))
+  .add('Feature - Footer Callbacks', () => {
+    const wrapper = document.createElement('div');
+
+    const footerMarkup = `
+    <axa-footer dynamic>
+      <h2 slot="column-0-title-desktop">axa & you</h2>
```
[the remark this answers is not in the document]
raphaellueckl: Ah right, `title-desktop`. That one was still on my imaginary todo-list. Maybe I can rename it to `title-big-screen`.

src/components/30-organisms/footer/index.js
raphaellueckl: But none other can use it, since skatejs is enabling this. Skatejs has this onItemClick as props that are handed over. Litelement does currently not keep it as a prop. Only the default value is set here.

src/components/30-organisms/footer/index.js
raphaellueckl: I once started to do that. But then decided not to do it. I want to discuss this bilateral.

package.json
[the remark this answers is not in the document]
raphaellueckl: After discussion with @LucaMele we decided to clean the testing mess up in another PR.

src/components/30-organisms/footer/demo.js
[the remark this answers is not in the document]
raphaellueckl: @jackblackCH Do you have an opinion about the naming system here? `column-x...` etc

src/components/30-organisms/footer/index.js
raphaellueckl: Right, thx

src/components/30-organisms/footer/index.js
raphaellueckl: Attribute is now renamed and commented.

src/components/10-atoms/input-file/demo.js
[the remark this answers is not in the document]
raphaellueckl: Done

src/components/30-organisms/footer/index.js
raphaellueckl: Because the event becomes an element otherwise, but I also think that this sucks.

src/components/30-organisms/footer/index.scss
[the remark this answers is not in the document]
raphaellueckl: As discussed, we will watch this feature status and don't change it here.

src/components/30-organisms/footer/story.js
raphaellueckl: Will come again after marc merged his branch

package.json
jackblackCH: Do we really need new global scripts?

scripts/testcafe-chrome.sh
```diff
@@ -6,7 +6,7 @@ export $(egrep -v '^#' .env | xargs)
 
 npx start-storybook -p $TEST_HOST_STORYBOOK_PORT -c .storybook -s ./src/static --ci --quiet > /dev/null 2>&1 &
 npx wait-on $TEST_HOST_STORYBOOK_URL -t 30000
-npx testcafe "chrome:headless" ./**/ui.test.js
+npx testcafe "chrome:headless" -c 4 ./**/ui.test.js
```
jackblackCH: Very nice, I tried as well. saves more than a minute.

src/components/30-organisms/footer/demo.js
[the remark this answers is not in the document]
jackblackCH: Never use mobile or desktop in names. We always describe what it is in terms of content. The behavior of what's happening on the viewport can be seen in the CSS.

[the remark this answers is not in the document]
jackblackCH: Just call it title-body, title-head, title__head, title__head-1,

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,284 @@
...
+          <div class="o-footer__content">
+            <div class="o-footer__collection">
+              <div class="o-footer__main">
+                <slot
+                  name="column-0-title-desktop"
+                  class="o-footer__title-desktop"
+                ></slot>
+                <button
+                  class="o-footer__accordion-button"
+                  @click="${() => this._handleAccordionClick(0)}"
```
jackblackCH: What's the '0' for in the function argument?

src/components/30-organisms/footer/index.js
jackblackCH: yes!

src/components/30-organisms/footer/index.scss
jackblackCH: Hmm, it's a white with .25% transparency, if it's an exception I think it's better to add it here. If not, yes then add it to globals as a `$white-transparent` or similar

src/components/30-organisms/footer/story.js
jackblackCH: Merged.

package.json
raphaellueckl: Yes, but we will remove some others in other PRs. These here make sense, I used them a lot and always reverted them, but I think the way we develop, those will be the most requested ones during development.

src/components/30-organisms/footer/index.js
raphaellueckl: The index of the accordion button.

src/components/30-organisms/footer/demo.js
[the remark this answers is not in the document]
raphaellueckl: @jackblackCH This is not possible. The reason, that title-desktop exists is exactly the viewport, nothing else. The content, the sizes (more or less), etc are the same.

src/components/30-organisms/footer/index.js
raphaellueckl: done

src/components/30-organisms/footer/index.js
raphaellueckl: I can do that, but I don't see a reason. Isn't that readable?

src/components/30-organisms/footer/index.js
```diff
@@ -0,0 +1,285 @@
...
+  return !label || label.nodeType === 3;
+};
+
+const _extractNestedHref = ev => {
+  const { target } = ev;
+  let element;
+  if (!target || !target.href) {
+    element = ev;
+    while (!element.href) {
+      if (element.target && element.target.parentNode) {
```
LucaMele: @raphaellueckl here still not refactored. Please use deconstruction for duplication of code. And we need to clean up element = ev.

LucaMele: if you need help, lets do it together

src/components/30-organisms/footer/index.js
raphaellueckl: done

src/components/30-organisms/footer/index.js
raphaellueckl: Done!

src/components/30-organisms/footer/index.js
raphaellueckl: Good point!

src/components/30-organisms/footer/index.js
raphaellueckl: Yes, this is probably a shadycss feature.

src/components/30-organisms/footer/index.js
raphaellueckl: Yes

src/components/30-organisms/footer/index.scss
raphaellueckl: I would say we don't add it. It would, however, make sense to introduce the variable @jackblackCH is mentioning with 0.65 opacity, since that is quite common.

src/components/30-organisms/footer/index.scss
raphaellueckl: Yes, I agree.

raphaellueckl: I created an issue for this request: https://github.com/axa-ch/patterns-library/issues/1109

src/components/30-organisms/footer/index.js
raphaellueckl: Ok I will join you later in SB

PR #1085, opened 2019-06-14
Bugfix/markdown

jackblackCH:
I fixed/improved the broken markdown rendering by using the community plugin which got patched and can be used with storybook/html. It's looking really nice now.

Works like a charm and our READMEs are rendered with GFM.

Disclaimer: Also while refactoring, I figured out, that our self-written storybook plugins are somehow not "chainable", probably missing some essential plugin code rules and so are breaking other storybook plugins from working correctly when added together. 

So I temporarily removed the hardly used reset-body, will be readded again, but it needs a refactoring - in a separate PR.

Screenshot 2019-06-15 at 01 08 50

Screenshot 2019-06-15 at 02 07 58


LucaMele:
Nice! Glad they added it!! Nice Job!

But i think the body reset must be part of this as otherwise footer small  top content bar and footer will look broken due to mediaquery being viewport based and not being element based.

Lets not have better readme for the price of broken previews.

jackblackCH:
Well, the preview is not broken. It's simply a feature of Storybook. 

If you want to ensure our clients that the padding is not included in a component, I can understand but let's do it consistently everywhere then. 

We can add this to `preview-head.html`
```html
<style type="text/css">
  body {
    margin: 0;
    padding: 0;
  }
</style>
```

LucaMele:
@jackblackCH no because most of the components look better with the default padding. There is a reason why i lost half a day on my life making this addon so please let it be part of the pr. a PR should always leave the code cleaner behind. Fixing one broken window and breaking another one is beside the point in my opinion.

To summarise:
* Most components looks better with the page padding.
* 100% needs a body reset as otherwise they are broken ( not just looks bad, but broken due to media query triggering later )

The body reset is mandatory and I insist and it must be optional with default disabled. Same as now. Reason is not purely aesthetic but it actually can make 100% components that have media query brake. If we leave this feature out, we will have a downgrade in quality and are going backwards

jackblackCH:
OK I could simplify the desired `no-border` plugin. same way, less code, less magic.
I readded the no-border solution for the footer.

src/components/20-molecules/cookie-disclaimer/story.js
```diff
@@ -1,52 +1,25 @@
...
-  .addDecorator(withNoBorder)
+  .addDecorator(story => {
+    return withNoBorder(story);
+  })
   .addParameters({
     readme: {
       sidebar: Readme,
     },
   })
-  .add('Cookie disclaimer - default', () => {
```
LucaMele: All this mess code is here to show what happens when u click on it and shows u when u clicked as deno. Please leave this as is great demo value

LucaMele: Maybe ask for the reason why someone does something before deleting. If u leave ur example, once someone clicked on accept he will always see blank when he opens storybook. Thats why i took the effort of making a nice example and not just because it looks bad means is bad. Is a demo anyway.

jackblackCH:
As discussed, while touching the cookie disclaimer, I fixed the mobile view.

package.json
```diff
@@ -83,6 +83,7 @@
     "rollup-plugin-sass": "^1.1.0",
     "rollup-plugin-uglify": "^6.0.2",
     "sass-loader": "^7.1.0",
+    "storybook-readme": "^5.0.4-alpha.2",
```
LucaMele: alpha?

CONTRIBUTION.md
```diff
@@ -92,12 +92,15 @@ Example:
...
 import withBodyReset from '../../../../.storybook/addons/reset-body';
 import Readme from './README.md';
 storiesOf('Molecules/Top content bar', module)
-  .addDecorator(withMarkdown(Readme))
-  .addDecorator(withBodyReset())
+  .addParameters({
+    readme: {
+      sidebar: Readme,
+    },
+  })  .addDecorator(withBodyReset())
```
LucaMele: here maybe a new line

src/components/20-molecules/cookie-disclaimer/index.js
```diff
@@ -32,6 +32,8 @@ class AXACookieDisclaimer extends LitElement {
     this.onClick = () => {};
   }
 
+  //  With this, it removes it itself if you use the static <axa-cookie-disclaimer> tag.
+  //  Needs to be re-evaluated.
```
LucaMele: i dont think it needs to be reevaluated, can u remove this text please

src/components/20-molecules/top-content-bar/story.js
```diff
@@ -1,13 +1,14 @@
 /* global document */
 import { storiesOf } from '@storybook/html';
 import './index';
-import { withMarkdown } from '../../../../.storybook/addons/markdown';
-import withBodyReset from '../../../../.storybook/addons/reset-body';
 import Readme from './README.md';
```
LucaMele: Here u forgot to add noBodyReset

src/components/20-molecules/cookie-disclaimer/index.js
jackblackCH: True. Removed.

src/components/20-molecules/top-content-bar/story.js
jackblackCH: I readded.

CONTRIBUTION.md
jackblackCH: There were more mistakes here. Good catch. I updated.

PR #1096, opened 2019-06-20
Refine the document

raphaellueckl:
If you think this is an improvement, please merge it. If you think some things need to change, but overall it is an improvement, please merge and add your improvements in a separate PR or add a commit directly in here.

### DIFF
--- a/CONTRIBUTION.md
+++ b/CONTRIBUTION.md
@@ -1,6 +1,11 @@
 # Contribution
 
-## Install the AXA Pattern Library
+## What you need to contribute
+
+- [Node](https://nodejs.org/en/) (v10.15.3), which is the same version that runs on the pipeline.
+- [Prettier](https://prettier.io/) - A code styling tool. You don't explicitly need it, but without fitting the ruleset, you will not be able to merge.
+
+## Install and run the AXA Pattern Library
 
 ```js
 npm install // automatically runs `npm run bootstrap as postinstall hook`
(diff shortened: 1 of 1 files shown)


CONTRIBUTION.md
```diff
@@ -15,18 +20,13 @@ Open your code editor of choice and edit your desired component/package in
...
 - We use [prettier](https://prettier.io/).
-  - Make sure you install eslint checker and prettier in your IDE
 
 ### Style Rules
 
 - We follow DRY, KISS, YAGNI principles when we write SCSS. This includes variables, mixins, etc.
 - We remove the linting of ordering and grouping of CSS properties. Only includes have to be at the beginning.
 - We won't have settings for vertical rhythm or global spacing.
 
-### JS Rules
```
LucaMele: This is still valid rule, even with prettier

raphaellueckl: but we removed some of them. the built ins being the most obvious ones.

LucaMele: We removed one or two and Prettier overrides the formatting rules. Still, Eslint is for code style, prettier for formatting and settings are alomst same as plib 1

Bildschirmfoto 2019-06-24 um 09 33 54

LucaMele: @raphaellueckl maybe update the text like so:

We use the plibv1 eslint styling for code style, instead prettier takes care of all the formatting rules in the plib2

raphaellueckl: I reverted it. I think it's better to go with the former approach imho, but we can also use your sentence. I think mentioning the linting rules here is enough.

PR #1104, opened 2019-06-24
[WIP]: Feature/component cleanup knobs

jackblackCH:
(At the moment, this PR actually contains changes of the open #1085 PR, this files won't show up once #1085 is merged)

Here is an example of a button component, where props are turned into knobs.
Knobs are controls in a side panel letting you dynamically adjust all attributes of a component and see the live result in the preview. (Re-renders the component on knob change)

The idea behind is to have fewer stories. Makes it easier to test all "versions/variants" of a component and also way easier to maintain and test stories/components.

Have a look at the prepared button component. (atoms/button).
Instead of a huge list of stories, where each story represents a component with a set prop, there is only one story with all the props as knobs in the panel.

Once we agree on that example, I will refactor the other components.

Concept: 
All Props are Knobs
Variants aka (CSS modifiers) will be grouped in a variant option. (Check out the example or the screenshot)

props

### DIFF
--- a/src/components/10-atoms/button/story.js
+++ b/src/components/10-atoms/button/story.js
@@ -1,6 +1,12 @@
 /* global document */
 import { storiesOf } from '@storybook/html';
-import { boolean, radios, text, withKnobs } from '@storybook/addon-knobs';
+import {
+  boolean,
+  select,
+  radios,
+  text,
+  withKnobs,
+} from '@storybook/addon-knobs';
 import './index';
 import Readme from './README.md';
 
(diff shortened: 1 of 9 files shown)


src/components/10-atoms/button/story.js
```diff
@@ -1,65 +1,65 @@
...
-
-  /* Icon */
-  .add(
-    'Button - icon',
-    () => `<axa-button icon="arrow-right">Icon</axa-button>`
-  );
+  return `
+  <div style='${variants === 'inverted' ? blueBackgroundStyle : ''}'>
+    <axa-button
+      ${disabled ? 'disabled' : ''}
```
MKaHead: Maybe prettier to write like that <axa-button ?disabled=${disabled} ... > only works for Booleans

danyball:
@MKaHead Machst du noch den Rebase?

danyball:
This branch is too old to rebase. We will open a new PR.

PR #1105, opened 2019-06-25
Update README.md

LucaMele:


README.md
```diff
@@ -59,3 +59,17 @@ We are is dedicated to building a welcoming, diverse, safe community. We expect
...
+```js
+import { createElement } from 'react';
+
+import createAXAButtonReact from '@axa-ch/button/lib/index.react';
+
+export const AXAButton = createAXAButtonReact(createElement);
+
+AXAButton.displayName = 'AXAButton';
+```
+In future the patterns library will take care of this automatically
```
raphaellueckl: Here is a corrected version of this text:

=============================


## Testing with Enzyme
Our wrapped Custom Element at the moment renders an `<Unknow>` tag in React. If enzyme selections like `wrapper.find(AXAButton)` do not work, please add a display name to your mounted react component like this:

 ```js
import { createElement } from 'react';
 import createAXAButtonReact from '@axa-ch/button/lib/index.react';
 export const AXAButton = createAXAButtonReact(createElement);
 AXAButton.displayName = 'AXAButton';
```
In the future, the pattern library will take care of this automatically.

LucaMele: Is more a react thing then an enzym
Enzy is just the one we know that relies on this react naming

LucaMele: @raphaellueckl apart of the title, done

PR #1127, opened 2019-06-28
Feature/table height scroll

LucaMele:


src/components/30-organisms/table-sortable/index.js
```diff
@@ -48,26 +49,23 @@ class AXATableSortable extends LitElement {
...
       },
-      tfoot
+      tfoot,
     } = this.model;
     let tfootL = 0;
     if (tfoot && tfoot[0]) {
       tfootL = tfoot.length;
     }
-    return !tbodyL || !((theadL !== tbodyL) && (!tfootL || tfootL === tbodyL));
+    return !tbodyL || !(theadL !== tbodyL && (!tfootL || tfootL === tbodyL));
```
raphaellueckl: I would extract this part into a named method, since I have no idea what this is doing.

Issue #1114, opened 2019-06-27
Follow up Dropdown issues


LucaMele:
- Not native has bigger bottom border -> Feedback Jonas: needs to be fixed.
- Dropdown native has different padding with dropdown not native what makes it look broken on screen resize -> Feedback Jonas: needs to be fixed.

- Selected option is not respected, default title is displayed instead on the "enhanced version":
![Bildschirmfoto 2019-07-01 um 09 59 10](https://user-images.githubusercontent.com/275235/60420030-fcf1d000-9be6-11e9-985c-e4dae7b061d2.png)
<img width="442" alt="Bildschirmfoto 2019-07-01 um 09 59 25" src="https://user-images.githubusercontent.com/275235/60420031-fcf1d000-9be6-11e9-8252-1db20ceb62d5.png">

- Imported in a react project as client, following effect found:
<img width="1150" alt="Bildschirmfoto 2019-07-01 um 09 58 09" src="https://user-images.githubusercontent.com/275235/60419958-d2a01280-9be6-11e9-87eb-6c9c896b0881.png">
<img width="1133" alt="Bildschirmfoto 2019-07-01 um 09 58 13" src="https://user-images.githubusercontent.com/275235/60419959-d338a900-9be6-11e9-8b2a-5c980411b9ae.png">
<img width="1199" alt="Bildschirmfoto 2019-07-01 um 09 58 17" src="https://user-images.githubusercontent.com/275235/60419960-d338a900-9be6-11e9-9e2d-83c4a9da25a0.png">

markus-walther:
Fixed in minor release 2.1.0.

PR #1132, opened 2019-07-02
Bugfix/axa dropdown bugs

markus-walther:
Rebased, and tested with IE as well as symlinked create-react-app. All tests pass.

Claims to fix all issues in https://github.com/axa-ch/patterns-library/issues/1114 plus
associated follow-up problems in datepicker.

Please verify carefully that claims hold and things work - however, major new findings would be out of scope here and should lead to new issue(s).

Thanks, Markus

### DIFF
--- a/src/components/20-molecules/datepicker/index.js
+++ b/src/components/20-molecules/datepicker/index.js
@@ -133,7 +133,6 @@ class AXADatepicker extends NoShadowDOM {
     this.month = this.startDate.getMonth();
     this.day = this.startDate.getDate();
     this.allowedyears = [this.year];
-    this.inputplaceholder = 'Please select a date';
     this.outputdate = '';
     this.onChange = EMPTY_FUNCTION;
     this.handleWindowKeyDown = this.handleWindowKeyDown.bind(this);
--- a/src/components/20-molecules/datepicker/index.js
+++ b/src/components/20-molecules/datepicker/index.js
@@ -51,6 +51,24 @@ const applyEffect = self =>
     }, 0 /* execute after render() */);
   });
 
+const parseAndFormatAllowedYears = ({ allowedyears, year }) => {
+  let result = [year];
+  allowedyears.forEach(years => {
+    if (typeof years === 'string') {
+      const splitYears = years.split('-');
+      const generatedYears = range(
+        parseInt(splitYears[0], 10),
+        parseInt(splitYears[1], 10)
+      );
+      result = result.concat(generatedYears);
+    } else {
+      result.push(years);
+    }
+  });
+
+  return result.filter((item, index) => result.indexOf(item) >= index).sort();
+};
+
 // CE
 class AXADatepicker extends NoShadowDOM {
   static get tagName() {
@@ -151,6 +169,9 @@ class AXADatepicker extends NoShadowDOM {
       } = this;
       // controlledness is a React-only concept
       this.state.isControlled = isReact && isControlled;
+    }
+    if (changedProperties.has('allowedyears')) {
+      this.allowedyears = parseAndFormatAllowedYears(this);
     }
     return true;
   }
@@ -331,24 +352,8 @@ class AXADatepicker extends NoShadowDOM {
     this.startDate.setHours(0);
     this.startDate.setMinutes(0);
     this.startDate.setSeconds(0);
-    // the chosen start year is in the allowed years
-    let allowedYearsFinal = [this.year];
-    this.allowedyears.forEach(years => {
-      if (typeof years === 'string') {
-        const splitYears = years.split('-');
-        const generatedYears = range(
-          parseInt(splitYears[0], 10),
-          parseInt(splitYears[1], 10)
-        );
-        allowedYearsFinal = allowedYearsFinal.concat(generatedYears);
-      } else {
-        allowedYearsFinal.push(years);
-      }
-    });
 
-    this.allowedyears = allowedYearsFinal
-      .filter((item, index) => allowedYearsFinal.indexOf(item) >= index)
-      .sort();
+    this.allowedyears = parseAndFormatAllowedYears(this);
 
     this.date = this.startDate;
     this.store = new Store(this.locale, this.date, this.allowedyears);
--- a/src/components/20-molecules/datepicker/index.js
+++ b/src/components/20-molecules/datepicker/index.js
@@ -110,14 +110,15 @@ class AXADatepicker extends NoShadowDOM {
   set value(newValue) {
     const {
       state: { isControlled, value },
+      state,
     } = this;
     // first value coming in indicates controlledness?
     if (!isControlled && newValue !== undefined) {
       // yes, remember in model state
-      this.state.isControlled = true;
+      state.isControlled = true;
     }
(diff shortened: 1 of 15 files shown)


src/components/20-molecules/datepicker/utils/date.js
```diff
@@ -93,7 +93,9 @@ const parseLocalisedDateIfValid = (locale = 'en-UK', inputValue = '') => {
   );
 
   const isValid =
-    !Number.isNaN(dateAsUnixEpochInteger) && dateAsUnixEpochInteger >= 0;
+    // IE 11 doesn't support Number.isNaN (bleeding-edge ES6)
+    // eslint-disable-next-line no-restricted-properties
+    !window.isNaN(dateAsUnixEpochInteger) && dateAsUnixEpochInteger >= 0;
```
MKaHead: Maybe better readable if the comparison only with Booleans.

src/components/20-molecules/dropdown/index.js
```diff
@@ -103,15 +103,16 @@ class AXADropdown extends NoShadowDOM {
 
   set value(newValue) {
     const {
-      state: { isControlled, value },
+      state: { isControlled, value, firstTime },
     } = this;
     // first value coming in indicates controlledness?
-    if (!isControlled && newValue !== undefined) {
+    if (!isControlled && newValue !== undefined && firstTime) {
```
MKaHead: ShouldUpdate lifecycle is to late, for the adaptations?

src/components/20-molecules/dropdown/index.js
```diff
@@ -356,7 +329,6 @@ class AXADropdown extends NoShadowDOM {
   firstUpdated() {
     // dropdowns cannot meaningfully be initially open, since the underlying
     // <select> cannot be force-opened natively
-    this.open = false;
     this.dropdown = this.querySelector('.js-dropdown__content');
     this.select = this.querySelector('.js-dropdown__select');
     this.addEventListener('focus', this.onFocus, false);
```
MKaHead: Why not bind focus and blur eventListener directly in template? Then there no need to clean them on disconnected callback.

src/components/20-molecules/dropdown/react/DemoUncontrolledDropdownReact.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
```
MKaHead: Maybe we need demo that the onFocus and onBlur Props Callbacks work in react environment.

MKaHead:
The `expand-svg` width height is incorrect. Because the view box is 24x24. The fix for that problem is to the `expand-svg` to 10x10.

src/components/20-molecules/dropdown/index.js
markus-walther: Unfortunately, a property-changing operation is needed *before* render() - currently implemented in shouldUpdate - hence moving the isControlled logic to after render() is too late.

src/components/20-molecules/dropdown/index.js
markus-walther: done

src/components/20-molecules/dropdown/react/DemoUncontrolledDropdownReact.jsx
markus-walther: done

markus-walther:
I also added a commit for the expand-svg problem

src/components/20-molecules/datepicker/index.js
```diff
@@ -332,24 +352,8 @@ class AXADatepicker extends NoShadowDOM {
...
-        allowedYearsFinal = allowedYearsFinal.concat(generatedYears);
-      } else {
-        allowedYearsFinal.push(years);
-      }
-    });
 
-    this.allowedyears = allowedYearsFinal
-      .filter((item, index) => allowedYearsFinal.indexOf(item) >= index)
-      .sort();
+    this.allowedyears = parseAndFormatAllowedYears(this);
```
MKaHead: Maybe more readable, if you can only pass stuff thats needed in the function. Then the function have no reason to destruct a object.

src/components/20-molecules/datepicker/utils/date.js
LucaMele: please remove `window.` or something like this `(window || global).isNaN(..)` so that it could be used in node unit tests

src/components/20-molecules/dropdown/index.js
```diff
@@ -354,9 +335,6 @@ class AXADropdown extends NoShadowDOM {
   }
 
   firstUpdated() {
-    // dropdowns cannot meaningfully be initially open, since the underlying
-    // <select> cannot be force-opened natively
-    this.open = false;
     this.dropdown = this.querySelector('.js-dropdown__content');
     this.select = this.querySelector('.js-dropdown__select');
     this.addEventListener('focus', this.onFocus, false);
```
MKaHead: i'ts obsolete, because you moved to template?

src/components/20-molecules/dropdown/index.js
```diff
@@ -103,15 +103,16 @@ class AXADropdown extends NoShadowDOM {
...
     } = this;
     // first value coming in indicates controlledness?
-    if (!isControlled && newValue !== undefined) {
+    if (!isControlled && newValue !== undefined && firstTime) {
       // yes, remember in model state
       this.state.isControlled = true;
     }
     // update state
     this.state.value = newValue;
+    this.state.firstTime = false;
```
LucaMele: this get state set firstTime -> `this get state` is duplicated more then once. Please do something like this:

`const { state } = this`
`state.isControlled `
`state.value`
`state.firstTime `

src/components/20-molecules/dropdown/index.js
```diff
@@ -369,6 +347,14 @@ class AXADropdown extends NoShadowDOM {
   updated() {
     this.updateTitle();
   }
+
+  disconnectedCallback() {
+    this.removeEventListener('focus', this.onFocus, false);
+    this.removeEventListener('blur', this.onBlur, false);
+    window.removeEventListener('resize', this.handleResize);
```
LucaMele: dtto with window. better maybe `(window || global)`

src/components/20-molecules/dropdown/index.js
markus-walther: yes, obsolete. Removed.

src/components/20-molecules/datepicker/utils/date.js
markus-walther: I removed `window`.

src/components/20-molecules/dropdown/index.js
markus-walther: I did something equivalent, but simpler...

src/components/20-molecules/dropdown/index.js
markus-walther: This does not make sense here, since most of the 10 instances of `window` relate to DOM API functions, and in node unit testing require `jsdom` anyway, which provides a `window` object.

Not done.

PR #1160, opened 2019-07-12
add migration notes to scaffolding

danyball:


scripts/create-a-m-o/functions.js
```diff
@@ -145,6 +145,10 @@ const createFiles = (store, a, m, o, done) => () => {
     ### onClick
 
     The function-valued attribute \`onClick\` can be used as a callback prop for React and other frameworks.
+    
+    ### Migration Notes (v1.0 -> v2.0)
```
LucaMele: I would remove `(v1.0 -> v2.0)`

scripts/create-a-m-o/functions.js
```diff
@@ -145,6 +145,10 @@ const createFiles = (store, a, m, o, done) => () => {
...
+    
+    You don't have to pay attention to anything for upgrading to version 2.
```
LucaMele: also here, better no version added. somethink like: `You don't have to pay attention to anything for upgrading to newer version.`

PR #1161, opened 2019-07-12
cleanup table with clean history

### LINKED ISSUE #1128: Cleanup Table
from this PRs comments: https://github.com/axa-ch/patterns-library/pull/1127

TODOs:
- Add better test data (texts that make sense)
- Fix @raphaellueckl comment: "I would extract this part into a named method, since I have no idea what this is doing."

danyball:
fixes #1128


src/components/30-organisms/table-sortable/unit.test.js
```diff
@@ -0,0 +1,73 @@
+import AXATableSortable from './index';
+
+describe('validateModel()', () => {
+  test('should return true if no elements are given', () => {
```
LucaMele: Please add a test where we check if footer is optional, something like so:

```
AXATableSortable.prototype.model = {
      thead: [],
      tbody: [[]],
    };

     expect(AXATableSortable.prototype.validateModel()).toBe(true);
```

danyball:
These tests failed with one-liner-if, but i they have to be green?

```
 ✕ should return true if tbody has no elements, but thead and tfoot each have one (26ms)
    ✕ should return true if thead has no elements, but tfoot and tbody each have one (1ms)
```

danyball:
@LucaMele i refactored the validation logic (see last comment) and fixed a bug in line 97 `tfoot[0].length`. The PR is ready to review again.

LucaMele:
Hello @danyball , tbody is mandatory and tfoot makes no sense to exist without tbody, therefore the test should be: `should return false if tbody has no elements, but thead and tfoot each have one`

LucaMele:
So to answer:

```
 ✕ should return true if tbody has no elements, but thead and tfoot each have one (26ms) -> FALSE
 ✕ should return true if thead has no elements, but tfoot and tbody each have one (1ms) -> I guess TRUE
 ✕ should return true if tbody is set but nothing else (6ms) -> I guess TRUE
```

PR #1166, opened 2019-07-15
feature/embedded-textarea-input

### LINKED ISSUE #1147: All form components must implement a flag that disable any place reservation
At the moment, Textarea, Inputs, Dropdown, etc reserve place for checkmark and error messages.

Now this place reservation makes sense only if components are used in sync inside a form. But the plib is more flexible then that and must allow devs to flag this place reservation off.

Some related issues
#1146

MKaHead:
fixes #1147

- for review try out src/components/10-atoms/[textarea|input-text/react/DemoContUncontTextarea.jsx


src/components/10-atoms/input-text/index.scss
```diff
@@ -80,20 +80,36 @@ axa-input-text {
 
   .a-input-text__error {
     @include typo();
-    margin-top: 10px;
+    display: block;
```
NoahLiechti: use `$border-width` and `$border-type` on line 23

src/components/10-atoms/input-text/index.js
```diff
@@ -142,8 +145,16 @@ class AXAInputText extends NoShadowDOM {
...
-    const errorMessageClasses = {
-      'a-input-text__error': true,
+    const checkWrapperClasses = {
+      'a-input-text__check-wrapper': true,
+      'a-input-text__check-wrapper--reservation': !embedded,
+      'a-input-text__check-wrapper--hidden': embedded && !checkMark,
+    };
+
+    const errorMessageWrapperClasses = {
+      'a-input-text__error-wrapper': true,
```
NoahLiechti: this class is never used in css

src/components/10-atoms/textarea/index.js
```diff
@@ -256,7 +271,7 @@ class AXATextarea extends NoShadowDOM {
                 <span>${modelCounter}</span>
```
NoahLiechti: you could use if/else here, because when for example `this.showCounter` is true than `this.showCounterMax` and `this.showError` are never going to be true

src/components/10-atoms/input-text/index.scss
MKaHead: fix in c9feb9c

src/components/10-atoms/input-text/index.js
MKaHead: #83c3c9c

src/components/10-atoms/input-text/story.js
```diff
@@ -23,6 +23,11 @@ storiesOf('Atoms/Input text', module)
     () =>
       '<axa-input-text checkmark placeholder="Pre-filled content"></axa-input-text>'
   )
+  .add(
+    'Input text - embedded',
+    () =>
+      '<axa-input-text embedded></axa-input-text>'
+  )
```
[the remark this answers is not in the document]
MKaHead: fix in ed2bc23

src/components/10-atoms/textarea/index.js
MKaHead: `this.showError` have higher priority then the other two function, because of readability of markup i don't change

PR #1176, opened 2019-07-17
feature/popup

### LINKED ISSUE #1214: Colors differs on axa-dropdown and axa-input-text
Bildschirmfoto 2019-08-08 um 10 54 30

Either text or dropdown has wrong border color. Please @kenaxa tell which color is correct

### LINKED ISSUE #1215: Dropdown, label not present. conflicting with zepelin
There is no way to add a label property to dropdown. I think in a form it should make sense to add label or not add label the same way to all components. input type text has such property named `label` but axa dropdown hasnt. Here both needs to be synced

MKaHead:
- Add Popup functionality to input text only mobile devices design
- Input-text improvements and design adaptions, remove embedded etc

Other:
- Update dependencies
- Update Dropdown styles to fixes #1214
- Add label to dropdown fixes #1215


src/components/20-molecules/popup/popup-mixin/index.js
```diff
@@ -0,0 +1,33 @@
...
+
+    constructor() {
+      super();
+      this._open = false;
+    }
+
+    handlePopupButtonClick = () => {
+      this._open = !this._open;
+      
+      // (next minor)
```
raphaellueckl: outcommented code

PR #1179, opened 2019-07-18
feature/update input-text types

branislav-remen:
Hi all,
I found that value is missing in type definition and improved on* event handler types.

### DIFF
--- a/src/components/10-atoms/input-text/index.react.d.ts
+++ b/src/components/10-atoms/input-text/index.react.d.ts
@@ -15,9 +15,9 @@ interface AXAInputTextProps {
   checkMark?: boolean;
   embedded?: boolean;
   isReact?: boolean;
-  onFocus?: React.FocusEventHandler<HTMLInputElement>;
-  onBlur?: React.FocusEventHandler<HTMLInputElement>;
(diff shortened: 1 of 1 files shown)


branislav-remen:
> ```
> onFocus?: React.FocusEventHandler<HTMLInputElement>;
> onBlur?: React.FocusEventHandler<HTMLInputElement>;
> ```
> 
> Please check that the event object that this web component is sending on parameter of these callbacks is compatible with SyntheticEvent.
> 
> ```
>     node_modules/@types/react/index.d.ts:
>     type EventHandler<E extends SyntheticEvent<any>> = { bivarianceHack(event: E): void }["bivarianceHack"];
> ```

it's compatible
image

MKaHead:
Hi @MarekLacoAXA, @branislav-remen, it's the native FocusEvent not React SyntheticEvent.

branislav-remen:
> Hi @MarekLacoAXA, @branislav-remen, it's for example focus - the native FocusEvent not the React SyntheticEvent(FocusEvent). It is someway compatible because they wrap the native ones. But we should use then the native types.

Hi Marco, I changed types for events to avoid SyntheticEvent. For onChange I used only UIEvent, not InputEvent, which has weaker browser support right now.

PR #1183, opened 2019-07-22
WIP: Feature/1056 react classname

### LINKED ISSUE #1056: Map "class" property from React
## Current behavior:

- The class name "myClass" does not appear in DOM.

Example:
```
import createAXAButton from '@axa-ch/button/lib/index.react';
const Button = createAXAButton(React.createElement);
...
<Button className="myClass">...</Button>
```

Example renders as:
```
<axa-button type="button" style="-webkit-appearance: none;">....</axa-button>
```

## Wished  behavior (in all components where it makes sense):

- The class name "myClass" woud appear in DOM.

Example would render with **class="myClass"** :
```
<axa-button class="myClass" type="button" style="-webkit-appearance: none;">....</axa-button>
```

Thanks! :pray:

danyball:
fixes #1056

### DIFF
--- a/src/components/10-atoms/button/README.md
+++ b/src/components/10-atoms/button/README.md
@@ -103,6 +103,10 @@ Based on the string-valued attribute `icon`, interpreted as icon name, an icon w
 
 The function-valued attribute `onClick` can be used as a callback prop for React and other frameworks.
 
+### className
+
+With this string attribute you can set a CSS class to the component.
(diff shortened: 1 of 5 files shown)


danyball:
We have a weird npm/lerna bug and no logs - so recreate a PR: https://github.com/axa-ch/patterns-library/pull/1185

PR #1188, opened 2019-07-24
WIP: Feature/policy feature

danyball:
New Component for v2


danyball:
@LucaMele Branch is ready to review. The images i am using (in `src/components/20-molecules/policy-features/policy-features-item/index.js`) are "mocked" until the designers have a final release of the new images.

Please have a closer look at this branch, because it was my first lit-element component ;)

src/components/20-molecules/policy-features/policy-features-item/ui.test.js
```diff
@@ -0,0 +1,115 @@
...
+}).before(async t => {
+  await t.resizeWindow(lgWindowWidth, defaultWindowHeight);
+});
+
+test('should set width of item with screen size md-up', async t => {
+  await t
+    .expect(await firstArticleElementInDocument.getStyleProperty('width'))
+    .eql('210px');
+}).before(async t => {
+  await t.resizeWindow(mdWindowWidth, defaultWindowHeight);
```
LucaMele: if u want u can add this before on fixure level so that u dont have to copy it on every test like so: https://github.com/axa-ch/patterns-library/blob/1f5565ed8e264ff1e81819e2c417cdeb50534868/src/components/30-organisms/table-sortable/ui.test.js#L10

src/components/20-molecules/policy-features/story.js
```diff
@@ -0,0 +1,91 @@
+/* global document */
+import { storiesOf } from '@storybook/html';
+import { radios, select, text, withKnobs } from '@storybook/addon-knobs';
+import './policy-features-item/index';
+import './index';
+import Readme from './README.md';
+
+const story = storiesOf('Molecules/Policy features', module);
+story.addDecorator(withKnobs);
```
LucaMele: Please add noBorder cause this component is full with component so that demo is same as real case.

https://github.com/axa-ch/patterns-library/blob/1f5565ed8e264ff1e81819e2c417cdeb50534868/src/components/30-organisms/footer/story.js#L10

src/utils/requests.js
```diff
@@ -0,0 +1,9 @@
+export function xhrCall(url) {
```
LucaMele: nice

src/components/20-molecules/policy-features/policy-features-item/index.scss
```diff
@@ -0,0 +1,54 @@
+.policy-features-item {
```
LucaMele: Please use `m-` prefix

LucaMele:
Please make sure that width of `.m-policy-features` does not change depending on if embedding system has `box-sizing: border-box;` or not.

LucaMele:
Bildschirmfoto 2019-08-12 um 12 54 33

src/components/20-molecules/policy-features/README.md
```diff
@@ -0,0 +1,109 @@
...
+    icon="email"
+    description="A text which can be as long as you want."
+  ></axa-policy-features-item>
+</axa-policy-features>
+```
+
+### React
+
+You can use this component directly in your JSX without further React-ification (see Usage above).
+
```
raphaellueckl: Why is testimonials a molecule and not an organism?

src/components/20-molecules/policy-features/README.md
```diff
@@ -0,0 +1,109 @@
...
+    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <title>Your awesome title</title>
+  </head>
+  <body>
+    <axa-policy-features title="Your main title" axaStyle="dark-indigo">
+      <axa-policy-features-item
+        title="The title of item"
+        icon="email"
+        description="A text which can be as long as you want."
+      ></axa-policy-features-item>
```
raphaellueckl: I prefer to have copy&paste examples. This blows up the whole readme, but Marc Illien back then convinced me, that this is better. It should be as easy as possible for the user. So I would include the child fragments here too.

src/components/20-molecules/policy-features/README.md
```diff
@@ -0,0 +1,109 @@
...
+</html>
+```
+
+## Properties
+
+### Variant
+
+| Attribute                | Details                                                          |
+| ------------------------ | ---------------------------------------------------------------- |
+| `axaStyle="default"`     | background is `dark-indigo` and font color is set automatically. |
```
raphaellueckl: Why is this property called "axaStyle"? If I am correct, we do not use this terminology in other places.

src/components/20-molecules/policy-features/README.md
```diff
@@ -0,0 +1,109 @@
...
+| `axaStyle="white"`       | background is `white` and font color is set automatically.       |
+
+### title
+
+The attribute `title` (optional) specifies the main title at the top.
+
+### axaStyle
+
+The attribute `axaStyle` (optional) specifies the background and the font color of this component and all its children.
+
```
raphaellueckl: Unnecessary linebreak. If you use prettier, this whole readme would look way better without any efforts from you. It also automatically improves the tables. :)

src/components/20-molecules/policy-features/policy-features-item/index.js
```diff
@@ -0,0 +1,65 @@
+import { LitElement, html, css, unsafeCSS, svg } from 'lit-element';
+// eslint-disable-next-line import/no-extraneous-dependencies
```
raphaellueckl: Further down below you excluse this rule for the whole file, so you could just use that expression instead of both.

src/components/20-molecules/policy-features/policy-features-item/ui.test.js
```diff
@@ -0,0 +1,115 @@
...
+const defaultWindowHeight = 1000;
+const lgWindowWidth = 992;
+const mdWindowWidth = 768;
+const smWindowWidth = 576;
+
+// Selectors
+const firstSvgElementInDocument = Selector(
+  () => document.querySelector('axa-policy-features-item').shadowRoot
+).find('svg');
+const firstH1ElementInDocument = Selector(
```
raphaellueckl: I would love to have a better semantic here. What exacly is the first h1 element? I guess you want to test the title of the first item or whatever, or the title of the policy feature. I like to name my variables around the content they contain as value.

src/components/20-molecules/policy-features/policy-features-item/ui.test.js
```diff
@@ -0,0 +1,115 @@
...
+  await t
+    .expect(await firstSvgElementInDocument.getStyleProperty('width'))
+    .eql('42px')
+    .expect(await firstSvgElementInDocument.getStyleProperty('height'))
+    .eql('42px');
+}).before(async t => {
+  await t.resizeWindow(smWindowWidth, defaultWindowHeight);
+});
+
+test('should render title and description with size medium-1 on screen md', async t => {
```
raphaellueckl: I would call this `should render title and description with correct typography on tablet view`. This makes it more abstract and if anyone decides that medium-1 will be renamed at some point in time, it won't diverge with this description.

src/components/20-molecules/policy-features/policy-features-item/ui.test.js
```diff
@@ -0,0 +1,115 @@
...
+  await t
+    .expect(await firstH1ElementInDocument.getStyleProperty('font-size'))
+    .eql('18px')
+    .expect(await firstPElementInDocument.getStyleProperty('font-size'))
+    .eql('18px');
+}).before(async t => {
+  await t.resizeWindow(mdWindowWidth, defaultWindowHeight);
```
raphaellueckl: Cool! I use to do that in another way, but your approach is better.

src/components/20-molecules/policy-features/ui.test.js
```diff
@@ -0,0 +1,69 @@
+import { Selector } from 'testcafe';
+
+const host = process.env.TEST_HOST_STORYBOOK_URL || 'http://localhost:9999';
+const TAG = 'axa-policy-features';
+const CLASS = '.m-policy-features';
+const defaultWindowHeight = 1000;
+const huge2WindowWidth = 992;
+const hugeWindowWidth = 768;
+const large1WindowWidth = 576;
```
raphaellueckl: With those sizes, you never test the mobile mode, because this goes from 0-575.

Moreover I would use the devices as terminology, but that's a personal preference. Like `mobileViewPortWidth`, `tabledViewPortWidth`, etc.

src/components/20-molecules/policy-features/index.scss
```diff
@@ -0,0 +1,49 @@
...
+      @include typo-title(huge, secondary);
+    }
+
+    @include breakpoint($mediaquery-lg-up) {
+      @include typo-title(huge-2, secondary);
+    }
+  }
+
+  &__slot {
+    display: flex;
```
raphaellueckl: Is that according to the styleguide? Because there are situations, where we get 2 rows because of the screen size, an the first row contains 3 items and the second only 1.

src/components/20-molecules/policy-features/README.md
```diff
@@ -0,0 +1,109 @@
...
+You have to set `axa-policy-features-item` as childs of `axa-policy-features`. It renders a icon on top, a headline and a description text. Usage see examples above.
+
+## Properties
+
+### Variant
+
+| Attribute                | Details                                                          |
+| ------------------------ | ---------------------------------------------------------------- |
+| `icon="email"`     | The email icon of our icon-set. |
+| `icon="https://url-to-svg-icon.svg"`     | Loads the url and sets content as icon. |
```
raphaellueckl: This table would look better, if you use prettier. It formats those correctly out of the box.

src/components/20-molecules/policy-features/README.md
```diff
@@ -0,0 +1,109 @@
...
+    title="The title of item"
+    icon="email"
+    description="A text which can be as long as you want."
+  ></axa-policy-features-item>
+</axa-policy-features>
+```
+
+### React
+
+You can use this component directly in your JSX without further React-ification (see Usage above).
```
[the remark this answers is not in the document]
danyball: Because `axa-icon` also do not have a wrapper. And marcus told be i do not need it for this component.

src/components/20-molecules/policy-features/policy-features-item/ui.test.js
danyball: Every test needs another screen size - so i have to write it that way

danyball:
> Please make sure that width of `.m-policy-features` does not change depending on if embedding system has `box-sizing: border-box;` or not.

Ich bekomm das nicht nachgestellt?! Wo hast du den style gesetzt?

src/components/20-molecules/policy-features/index.scss
```diff
@@ -0,0 +1,49 @@
...
+    flex-wrap: wrap;
+    justify-content: space-around;
+  }
+}
+
+.m-policy-features__style-dark-indigo,
```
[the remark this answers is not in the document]
danyball: It would be a breaking change if there is no possibility to set axastyle="default". I copied it from V1. Delete "default" anyway? I also think its cleaner.

src/components/20-molecules/policy-features/index.js
```diff
@@ -0,0 +1,65 @@
...
+import styles from './index.scss';
+
+export const STYLE_WHITELIST = [
+  'default',
+  'dark-indigo',
+  'axa-blue',
+  'wild-sand',
+  'white',
+];
+export const DEFAULT_AXA_STYLE = 'dark-indigo';
```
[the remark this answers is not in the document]
danyball: Yes right, will remove this const. I was used to do it that way to import it in tests.

src/components/20-molecules/policy-features/README.md
danyball: Marco, Luca and me had a discussion about that. For Marco and me this component has not much complexity for an organism.

src/components/20-molecules/policy-features/README.md
danyball: This example has 1 child fragment. Or what do you mean?

src/components/20-molecules/policy-features/README.md
danyball: Have copied it from V1. What name is better?

src/components/20-molecules/policy-features/policy-features-item/ui.test.js
danyball: Yes, "correct size" is better than "medium-1". Will change that. "Tablet view" is readable very good, but do we have a specification for that? Or what is "sm"? A tablet or a big smartphone?

src/components/20-molecules/policy-features/ui.test.js
danyball: I do not test xs explicitly, because large1 is the smallest width defined in the scss. Every screen size smaller than large1 has the same behavior.

src/components/20-molecules/policy-features/index.scss
danyball: This situation exists in V1 too. I will ask a designer... good point.

danyball: Designer said its ok.

src/components/20-molecules/policy-features/README.md
danyball: variant

src/components/20-molecules/policy-features/index.js
```diff
@@ -0,0 +1,62 @@
...
+  static get properties() {
+    return {
+      title: { type: String },
+      variant: { type: String },
+    };
+  }
+
+  constructor() {
+    super();
+    this.variant = 'dark-indigo';
```
MKaHead: I would like more to see `dark indigo` as default styling. Because in best case the user have nothing to configure and we don't have to preselect something that is default.

src/components/20-molecules/policy-features/index.scss
```diff
@@ -0,0 +1,48 @@
...
+    }
+  }
+
+  &__slot {
+    display: flex;
+    flex-wrap: wrap;
+    justify-content: space-around;
+  }
+}
+
```
MKaHead: I would recommend to create a mini mixing that living in that file. Because the styleing of the variants are redundant and simple.

raphaellueckl:
Why does the 'react demo' exist? Does that provide a benefit or can we remove it?

raphaellueckl:
@LucaMele I don't want to disturb you, therefore I merged, since you agreed in the daily. You still had changes requested here, so if you do not agree, we can revert the merge commit. For now it's merged and Dani will release it tomorrow.

PR #1198, opened 2019-08-05
Add box sizing reset

LucaMele:
Reset box sizing to content box in case a parent app has hacks like: 

```
* { box-sizing: border-box; }
```

additionally add hover state


src/components/30-organisms/footer/child.scss
```diff
@@ -1,4 +1,9 @@
 axa-footer {
+
+  * {
```
markus-walther: what about :before, :after?

PR #1229, opened 2019-08-13
docu(input-text): improve TS types - onChange prop

MarekLacoAXA:
Change to typescript type definition for AXAInput.


MKaHead:
implement that change in #1240

PR #1235, opened 2019-08-16
Feature/slot to elements and default class name

raphaellueckl:
Gives classname a default and introduces 'slot' element to some elements that are "ready only".

### DIFF
--- a/src/components/10-atoms/input-file/index.react.d.ts
+++ b/src/components/10-atoms/input-file/index.react.d.ts
@@ -14,6 +14,8 @@ interface AXAInputFileProps {
   multiple?: boolean;
   capture?: boolean;
   className?: string;
+  slot?: string;
+
   onChange?: () => void;
 }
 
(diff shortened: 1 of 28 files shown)


src/components/10-atoms/input-file/index.react.d.ts
```diff
@@ -14,6 +14,8 @@ interface AXAInputFileProps {
   multiple?: boolean;
   capture?: boolean;
   className?: string;
+  slot?: string;
+
```
MKaHead: obsolete line

MKaHead:
> wouldn't it make sense to change the script `patterns-library/scripts/create-a-m-o/functions.js` to automatically add `className = ''`instead of just `className` for future components?

yes, should be issue (feature request)

NoahLiechti:
> > wouldn't it make sense to change the script `patterns-library/scripts/create-a-m-o/functions.js` to automatically add `className = ''`instead of just `className` for future components?
> 
> yes, should be issue (feature request)

oke I mad one #1239

PR #1236, opened 2019-08-16
Feature/improve readme

raphaellueckl:
I made some changes, because it annyoed me that I had to search for the components manually in the repo.

How it looks now: https://github.com/axa-ch/patterns-library/blob/feature/improve-readme/README.md


README.md
```diff
@@ -14,58 +14,75 @@
...
+| [AXA Table Sortable](./src/components/30-organisms/table-sortable)                 | [![npm version](https://img.shields.io/npm/v/@axa-ch/table-sortable.svg?style=flat)](https://www.npmjs.com/package/@axa-ch/table-sortable)                 |
+| [AXA Text](./src/components/10-atoms/text)                                         | [![npm version](https://img.shields.io/npm/v/@axa-ch/text.svg?style=flat)](https://www.npmjs.com/package/@axa-ch/text)                                     |
+| [AXA Textarea](./src/components/10-atoms/textarea)                                 | [![npm version](https://img.shields.io/npm/v/@axa-ch/textarea.svg?style=flat)](https://www.npmjs.com/package/@axa-ch/textarea)                             |
+| [AXA Top Content Bar](./src/components/20-molecules/top-content-bar)               | [![npm version](https://img.shields.io/npm/v/@axa-ch/top-content-bar.svg?style=flat)](https://www.npmjs.com/package/@axa-ch/top-content-bar)               |
 
 ## Alpha-Released Components
-- 
+
+- [AXA Popup](./src/components/20-molecules/popup) [![npm version](https://img.shields.io/npm/v/@axa-ch/popup.svg?style=flat)](https://www.npmjs.com/package/@axa-ch/popup)
```
NoahLiechti: In my opinion you could use here also a table like you used it for the released components. Keeping in mind, that there could be more than one alpha-released component.

README.md
```diff
@@ -14,58 +14,75 @@
...
+
+## Unreleased Components
+
+- [AXA Title Primary](./src/components/10-atoms/title-primary) [![npm version](https://img.shields.io/npm/v/@axa-ch/title-primary.svg?style=flat)](https://www.npmjs.com/package/@axa-ch/title-primary)
```
NoahLiechti: Same as above

README.md
MKaHead: Good input!

raphaellueckl: I thought about that too, but imho we should never have too much alpha components that are not finished and production ready, otherwise we are having issues in our process. That would be an extra table that would mostly be empty, which also looks weird. I would keep the table for really finished and released stuff and the other sections more as "informative".

README.md
raphaellueckl: I thought about that too, but imho we should never have too much alpha components that are not finished and production ready, otherwise we are having issues in our process. That would be an extra table that would mostly be empty, which also looks weird. I would keep the table for really finished and released stuff and the other sections more as "informative".

raphaellueckl:
@NoahLiechti I will merge this. If you are against it, feel free to open a discussion. We can always revert. :)

README.md
NoahLiechti: Oke I'm happy with that 👍

PR #1246, opened 2019-08-19
Features/images categories

danyball:


raphaellueckl:
All items don't have a width and height and the fill is set to `none`. Was that on purpose? Because we usually use `fill="currentColor"`.

src/components/00-materials/scripts/export-svg.js
```diff
@@ -14,80 +14,84 @@ const camelCase = _camelCase =>
...
+        const contents = fs
+          .readFileSync(iconPath, 'utf8')
+          .replace(/(\r\n|\n|\r)/gm, '');
+
+        let fileName = path.basename(icon);
+        fileName = cleanFileName(fileName);
+        const className = toClassName(fileName);
+
+        // Generate a js file forEach svg file found
+        fs.writeFileSync(
```
raphaellueckl: It doesn't make sesne to use synced methods here. But I leave it up to you to correct that, sicne I'm not sure how often we use that.

raphaellueckl:
Why is the script not recursive?

danyball:
> Why is the script not recursive?

Its not that easy with `fs` and we do not need it for now.

src/components/00-materials/scripts/export-svg.js
danyball: :D

danyball:
> All items don't have a width and height and the fill is set to `none`. Was that on purpose? Because we usually use `fill="currentColor"`.

For images these attributes are not relevant for now.

MKaHead:
- recursive script nice to have, but we should enable the images. We can open a new feature issue for the script.
- the images don't need current color because they have fixed color

PR #1260, opened 2019-08-26
Add favicons + References

raphaellueckl:
Adds a favicon to storybook


raphaellueckl:
Does not work in production, needs more time than I initially thought.

PR #1296, opened 2019-09-04
backgroundColor of images

### LINKED ISSUE #1255: images should have transparent background
Alle images haben weisse Hintergrundfarbe. Evtl. Zusammenhang mit currentcolor.

Bildschirmfoto vom 2019-08-21 14-19-53

danyball:
fixes #1255


danyball:
The images now have currentColor set. But color is the foreground --> the SVG sets the "foreground"-color as its fill-color. So we have to manually set the svg-color.

Problems:
- just `color` get through shadowdom, `background` doesnt (parent-wc sets color and child-wc inherited that). 
- cant get computedStyle of parent-wc shadowdom element (where background is set)

src/components/20-molecules/policy-features/policy-features-item/index.scss
```diff
@@ -18,6 +18,22 @@
 
   &__icon {
     max-width: 100%;
+    color: $color-prim-blue-dark-indigo; // default background color for svgs
+
+    // because of fill=currentColor we need to set color to the backgroundColor of the variant
+    &__style {
```
raphaellueckl: `style-axa-blue` is not an element, it should be a modifier.

`__icon--axa-blue` for instance. I would leave out the "style" in the property name.

danyball:
> Do we somewhere have a list of all icons to check if they still work and look correctly?

No. We have 300+ images - how to check if they look correct? i thinks thats impossible automatically.

src/components/20-molecules/policy-features/policy-features-item/index.scss
danyball: Oh, the "__" and "--" have a deeper meaning? Will fix that...

raphaellueckl:
@danyball We could check those icons on one big page where we would load all at the same time. Then we would instantly see, if there is a problem with them. That should be possible, but probably has a big impact on performance.

danyball:
> @danyball We could check those icons on one big page where we would load all at the same time. Then we would instantly see, if there is a problem with them. That should be possible, but probably has a big impact on performance.
> 
> EDIT: Sorry, I accidentially closed it.

I created a feature request for that: https://github.com/axa-ch/patterns-library/issues/1301 (we talked about that idea with marco a few days ago)

PR #1311, opened 2019-09-09
Axa ch/feature/arrow left icon

### LINKED ISSUE #1310: Missing arrow left icon for button
Design decision by @TommyHoliday 
We need arrow left icon for back button as seen in our design here:

Screen Shot 2019-09-09 at 11 16 47 AM

Invision prototype:
https://projects.invisionapp.com/share/K9TNWSMUC2F#/screens/381146365

salle18:
Resolves #1310


LucaMele:
But this wont fix the issue yet, correct? at it adds the icon only. It wont position it on the "left"

salle18:
> But this wont fix the issue yet, correct? at it adds the icon only. It wont position it on the "left"

All icons are automatically positioned on the left except `arrow-right`, check this https://github.com/axa-ch/patterns-library/blob/d83725777846fba733772a2559b4ac145f0d7e96/src/components/10-atoms/button/index.js#L56

I would like to create new issue where we can make button a bit more general e.g. allow positioning of icons on both left and right, allow custom icons in a slot so we don't have to add icon to the patterns library first etc.

LucaMele:
@salle18 true! I never went deeply to the button logic. Thanks!

Any PR for improvement is welcome.

ATM you can pass a icon URL instead of a reference of the icon tag: https://www.npmjs.com/package/@axa-ch/icon#icon-resource-path

raphaellueckl:
@salle18 @LucaMele Not sure about that. We should not allow everything, because stuff like that is more or less strictly regulated by the styleguide and we shouldn't create sandbox-components all over the way. AFAIK, the positioning of the icons needs to be aligned with our designers, if there is a request apart from the standard.

We don't just provide components for easier lifes. We also have to make sure that we have a similar experience and look&feel all over the frontends around here.

PR #1315, opened 2019-09-10
Fix errors in axa-input-text README and improve wording overall

markus-walther:


src/components/10-atoms/input-text/README.md
```diff
@@ -63,66 +64,80 @@ The type of input element to display.
...
 ### refId
 
-The string-valued `refId` sets the reference for label and input, if no `refId` is set a random id number will be created.
+The string-valued `refId` sets the reference ID for label and input. If no `refId` is set, a random ID will be created.
 
-### name\*
+### name
 
-The string-valued `name` of the element for purposes of form submission.
+The string-valued `name` of the element can be set for purposes of form submission.
```
LucaMele: Why do you make a `required` attribute to a `optional` one? the `*` meant that is mandatory to be set and now the text `can be set` implies optional

raphaellueckl: If name is mandatory, I wonder why it is not in the examples. The examples should be copy&paste.

Why is it mandatory? And if the star implies that, we should provide a legend imho.

LucaMele: I cant answer the questions. But u recon star implies required and it make sense form a Semantic Prospective

PR #1316, opened 2019-09-10
change size to 24x24 like the other icons

NoahLiechti:


NoahLiechti:
Bildschirmfoto 2019-09-10 um 12 45 01

src/components/00-materials/icons-raw/axa-logo-open.svg
```diff
@@ -1 +1 @@
...
...6 1.457h6.872s-1.027-1.781-1.26-2.376c-.133-.33-2.284-7.063-2.284-7.063l1.791-2.54h11.39s2.594 8.345 2.903 9.448c.202.83.315 1.678.336 2.531h7.171c-1.086-.769-3.817-10.42-5.143-14.128l-3.474-12.04c-.169-.458-.297-2.115-.297-2.115h-5.549c-.297.791-.804 1.518-1.149 2.157-.427.792-7.476 10.035-9.025 11.998-.228.292-.36.455-.36.447 0-.018-.049-.191-.118-.447-.256-.89-.78-2.676-.78-2.676s6.006-7.633 6.62-8.41c1.824-2.308 2.865-3.069 2.865-3.069h-5.687a3.783 3.783 0 0 1-.575 1.736c-.273.419-4.928 6.255-4.928 6.255s-1.464-5.28-1.741-6.135a5.947 5.947 0 0 1-.223-1.856h-6.826c.257.292 1.045 2.241 1.045 2.241l3.365 10.36-1.968 2.517c-.084-.126-3.645-11.808-3.797-12.295-.17-.551-.35-2.823-.35-2.823h-5.844s-1.135 2.47-1.922 3.613c-.325.473-4.03 5.437-7.978 10.542-4.393 5.854-9.176 11.791-9.28 11.867a1.6 1.6 0 0 0-.13.128V1.9h76.046v76.047zM22.64 49.503c.256 1.022 1.727 5.555 1.727 5.555h-7.595s4.796-5.838 5.61-7.59c.09-.195.108-.044.108.049a8.11 8.11 0 0 0 .15 1.986zm9.876 11.64l2.022-2.622s1.177 3.57 1.37 3.777l-2.09 2.872-1.302-4.027zm19.304-11.64c.257 1.022 1.729 5.555 1.729 5.555h-7.595s4.796-5.838 5.61-7.59c.09-.195.106-.044.106.049a8.111 8.111 0 0 0 .15 1.986z" fill="#00008F"/></g></svg>
```
LucaMele: Im not sure but i think also the viewbox should be adapted so that matches 1:1 to its size. otherwise we have a "zoomed" out SVG. Please double check my hypothesis.

So to be clear:
IS: `<svg viewBox="0 0 80 80" xmlns="http://www.w3.org/2000/svg" width="24" height="24" >`
SHOULD: `<svg viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" width="24" height="24" >`

src/components/00-materials/icons-raw/axa-logo.svg
```diff
@@ -1 +1 @@
...
...011-8.11551,12.1045H0v-7.949a5.9865,5.9865,0,0,1,.493-.4785c.386-.2835,18.213-22.4125,34.59-44.2335,14.717-19.0285,28.5255-37.5345,29.7365-39.297,2.9345-4.2625,7.163-13.4665,7.163-13.4665h21.782s.674,8.4665,1.309,10.522c.566,1.8165,13.8375,45.3615,14.15,45.8305l7.339-9.385-12.544-38.6135s-2.9395-7.2655-3.8965-8.354h25.4445a22.1658,22.1658,0,0,0,.83,6.9185c1.035,3.184,6.4895,22.8665,6.4895,22.8665s17.3535-21.753,18.369-23.3155a14.1,14.1,0,0,0,2.1435-6.4695h21.201s-3.8815,2.8365-10.6835,11.44c-2.285,2.8955-24.673,31.348-24.673,31.348s1.9535,6.66,2.9055,9.97552c.259.952.4395,1.5965.4395,1.665,0,.02948.493-.576,1.34252-1.665,5.77648-7.3195,32.051-41.7725,33.643-44.722,1.284-2.3825,3.1735-5.0925,4.282-8.0415H202.539s.4785,6.1765,1.1085,7.8855Zm-31.475-32.6125c-3.037,6.5335-20.913,28.296-20.913,28.296h28.3105s-5.488-16.8995-6.445-20.708a30.23266,30.23266,0,0,1-.557-7.4025C185.51748,169.6875,185.454,169.126,185.122,169.8485Zm-108.7745,0c-3.037,6.5335-20.913,28.296-20.913,28.296H83.745s-5.483-16.8995-6.44-20.708a30.23266,30.23266,0,0,1-.557-7.4025C76.748,169.6875,76.68,169.126,76.34749,169.8485ZM118.97,235.835l7.793-10.703c-.718-.77152-5.10748-14.082-5.10748-14.082l-7.53452,9.775Z"/></svg>
```
LucaMele: Im not sure but i think also the viewbox should be adapted so that matches 1:1 to its size. otherwise we have a "zoomed" out SVG. Please double check my hypothesis.

So to be clear:
IS: `<svg viewBox="0 0 283.46698 283.46699" xmlns="http://www.w3.org/2000/svg" width="24" height="24" >`
SHOULD: `<svg viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" width="24" height="24" >`

src/components/00-materials/icons-raw/axa-logo-open.svg
NoahLiechti: Thats with viewBox `0 0 24 24`.
Bildschirmfoto 2019-09-10 um 13 51 40
In my opinion this only works when the icon is actually 24x24 big. So either let the viewbox be as it is, or make a new icon with the size of 24x24.

LucaMele:
@danyball i dont know if the issue should be fixed in this PR. The title says only: "Bugfix/axa icons too big"

src/components/00-materials/icons-raw/axa-logo-open.svg
LucaMele: ok but seems then is not exported correctly from the designers

NoahLiechti:
@danyball I would prefer it in icons, because it has a dependency to icon-component. It is a feature that it can be used like this `<axa-icon icon="axa-logo"></axa-icon>` instead of 
`<axa-icon icon="/svg/logo-axa.svg"></axa-icon>`. But due the fact that the axa-logo from images has a different size that the logo from icon, in my opinion we can accept both.

PR #1324, opened 2019-09-11
Axa-textarea feature: show error when character limit reached

### LINKED ISSUE #1320: axa-textarea: max-char-entered should put the textarea into an error state
When entering too many characters, an error text "Max number of chars is reached"  is shown by &lt;axa-textarea&gt;

The error text below the textarea should be in red and the checkmark should not be shown.

Bildschirmfoto 2019-09-11 um 15 21 32

This feature request is approved by Martin and needed for AEM Forms.

markus-walther:
Fixes #1320

### DIFF
--- a/src/components/10-atoms/textarea/README.md
+++ b/src/components/10-atoms/textarea/README.md
@@ -1,6 +1,6 @@
 # Textarea
 
-The &lt;axa-textarea&gt; component is a wrapper for the HTML &lt;textarea&gt; element with custom styling and additional functionality, amongst other things a counter.
+The &lt;axa-textarea&gt; component is a wrapper for the HTML &lt;textarea&gt; element with custom styling and additional functionality, e.g. a character counter.
 
 ## Usage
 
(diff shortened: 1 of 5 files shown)


raphaellueckl:
Hi Markus

I did the same thing you do here for `input-text`, but with some slight changes that you could adapt:

If maxlength is 10, I allow only 9 characters and the 10th will lead to the "max number reached" validation error. Technically it is the same like before, but we do no longer "lie" to the user. If `maxlength` was set to 10 before, the user had a counter starting from 10. Like "10 characters left to insert..." which is wrong, since you are really only allowed to use 9 characters without triggering the validation message.

The reason for this were complaints from users, that copy&pasted a huge text into those fields and the part bigger than maxlength was just cut off and they did not notice (and apparently complained later). Therefore we need to inform the user that he probably used too may letters, instead of just silently cutting his content.

**tl;dr:** If max-length is 10, we should show the user "9 characters left to insert" instead of "10 characters left to insert" (but show a validation message if he really uses all 10).

PR #1331, opened 2019-09-13
Feature/enrich axa input

### LINKED ISSUE #1328: AXA Input, maxlength attribute should be added and behave the same as on textarea
AXA Input: maxLength, counter, counterMax attribute should be added and behave the same as on textarea.

Needed for AEM Forms.
Approved by Martin.

raphaellueckl:
resolves #1328


src/components/10-atoms/input-text/ui.test.js
```diff
@@ -86,3 +86,31 @@ test('should submit inputs correctly in form', async t => {
...
+  );
+  const $counterInfo = await $axaTag.find('.a-input-text__counter-info');
+  await t.expect($counterInfo.textContent).contains('Still 4 characters left');
+  const $checkMark = await $axaTag.find('.a-input-text__check');
+  await t.expect($checkMark.exists).ok();
+
+  const $input = await $axaTag.find(CLASS);
+  await t
+    .selectText($input)
+    .pressKey('delete')
```
markus-walther: If I understand this correctly, you want to ensure the input field is initially empty, right? This and the line before can be implemented more elegantly by using the `{replace: true}` option as the 2nd arg. of `typeText`.

markus-walther: See https://devexpress.github.io/testcafe/documentation/test-api/actions/action-options.html#typing-action-options

src/components/10-atoms/input-text/README.md
```diff
@@ -73,10 +74,27 @@ The string-valued `label` provides the label text as HTML.
...
 The string-valued `refId` sets the reference for label and input, if no `refId` is set a random id number will be created.
 
-### name\*
+### name (required within `<form>` tag)
 
 The string-valued `name` of the element for purposes of form submission.
 
+### maxLength
+
+Defines a maximum number of characters, that a user has available. Attention: If the user uses ALL available characters, a validation message will occur and the form should not be submittable. Therefore, the counter of available characters left displays always the value of `maxLength -1`. The reason for this strange behavior is, that another implementation is not possible without getting rid of the internal html `<textfield>` element, which would degrade the accessibility rating of the component.
```
markus-walther: Your wording is problematic as it stands: the form will be submittable anyway as currently implemented. i suggested rewriting this paragraph.

src/components/10-atoms/input-text/README.md
```diff
@@ -73,10 +74,27 @@ The string-valued `label` provides the label text as HTML.
...
+
+### counter
+
+Defines the text that the user will see, when he types into an input field that has the `maxLength`-attrbute set.
```
markus-walther: spelling: attribute

src/components/10-atoms/input-text/README.md
```diff
@@ -73,10 +74,27 @@ The string-valued `label` provides the label text as HTML.
...
+This can have several options:
+
+- Leave it empty, so only the numbers of characters left will be displayed.
+- Set it to "characters left" to get a message formatted like this: `9 characters left`.
+- Set the counter between the text: "Still ##counter## characters left" will result into `Still 9 characters left'`. Keep in mind that the user can also see the number "0", so adapt the text so this will not look weird. `Still 0 characters left` does not sound so good.
```
markus-walther: I think the explanation must go beyond examples to describe the general case, cf. axa-textarea README.

src/components/10-atoms/input-text/README.md
```diff
@@ -73,10 +74,27 @@ The string-valued `label` provides the label text as HTML.
...
+
+### counterMax
+
+Defines the validation message that the user will see, if he uses more characters than that are actually allowed.
```
markus-walther: Refer to `maxLength` to make explanation less vague?

src/components/10-atoms/input-text/index.js
```diff
@@ -54,14 +65,72 @@ class AXAInputText extends AXAPopupMixin(NoShadowDOM) {
...
+      !this.disabled
+    );
+  }
+
+  get showMessages() {
+    return this.showError || this.showCounter || this.showCounterMax;
+  }
+
+  get replaceCounterPlaceholder() {
+    return this.counter.replace(/##.*##/, this.charsLeft - 1);
```
markus-walther: Keep it DRY: I suggest `const userCharsLeft = this.charsLeft - 1` and then use it everywhere where you'd subtract 1 now.

src/components/10-atoms/input-text/index.js
```diff
@@ -104,15 +173,30 @@ class AXAInputText extends AXAPopupMixin(NoShadowDOM) {
       // yes, set UI from model state
       this.nativeInput.value = this.modelValue;
     }
+
+    if (this.maxLength) {
+      this.modelCounter = this.getCounterText;
+    }
   };
 
+  isNotAtMaxLength() {
```
markus-walther: I doubt this is needed. If maxLength could ever be of type `string`, just use `==` instead of `===`.

src/components/10-atoms/input-text/index.js
```diff
@@ -104,15 +173,30 @@ class AXAInputText extends AXAPopupMixin(NoShadowDOM) {
...
+    return this.charsLeft !== 0 && this.charsLeft !== '0';
+  }
+
   firstUpdated() {
-    const { defaultValue, isReact } = this;
+    const { defaultValue, isReact, value } = this;
     this.nativeInput = this.querySelector('input');
 
+    if (isReact) {
```
markus-walther: This smells DRY violation. I suggest `if (isReact) { this.nativeInput.value = defaultValue || value; }`

src/components/10-atoms/input-text/index.js
```diff
@@ -138,7 +223,8 @@ class AXAInputText extends AXAPopupMixin(NoShadowDOM) {
 
     const inputClasses = {
       'a-input-text__input': true,
-      'a-input-text__input--error': invalid && !disabled,
+      'a-input-text__input--error':
+        (invalid && !disabled) || this.showCounterMax,
       'a-input-text__input--check': checkMark && !disabled,
```
markus-walther: Why is checkMark not conditioned by non-maxlength-reached here?

src/components/10-atoms/input-text/index.js
```diff
@@ -165,13 +251,14 @@ class AXAInputText extends AXAPopupMixin(NoShadowDOM) {
...
             aria-required="${required}"
+            maxlength="${maxLength}"
             ?disabled="${disabled}"
             @input="${this.handleInput}"
             @focus="${this.handleFocus}"
             @blur="${this.handleBlur}"
           />
           ${
-            checkMark
+            checkMark && this.isNotAtMaxLength()
```
markus-walther: I suggest to wrap this composite condition in a single helper method `showCheckmark` and use that method both here and in line 228

src/components/10-atoms/input-text/index.react.js
```diff
@@ -16,6 +16,9 @@ export default createElement => ({
   required = false,
   disabled = false,
   checkMark = false,
+  maxLength = '',
```
markus-walther: This must be `maxLength = 0`, since you are assigning a property here (not a string-valued attribute) and the declared property is of type `Number` (which is semantically correct).

src/components/10-atoms/input-text/ui.test.js
```diff
@@ -86,3 +86,31 @@ test('should submit inputs correctly in form', async t => {
...
+    .typeText($input, 'Pattern Warriors')
+    .expect($input.value)
+    .eql('Patte');
+
+  await t.expect($checkMark.exists).notOk();
+  await t.expect($counterInfo.textContent).contains('Character limit reached!');
+  await t
+    .expect($counterInfo.getStyleProperty('color'))
+    .eql('rgb(201, 20, 50)');
+});
```
markus-walther: I expected also a test for `still 0 characters left` - could this be added?

src/components/10-atoms/input-text/ui.test.js
raphaellueckl: This is not my code, but I can fix that. I was confused too.

src/components/10-atoms/input-text/index.js
raphaellueckl: This also is not my code, so I better do not touch it within this PR. Not that I introduce potential other bugs.

src/components/10-atoms/input-text/index.react.js
raphaellueckl: According to w3c, the default value is `524288`. I will stick with that, if that is ok for you.

src/components/10-atoms/input-text/index.js
raphaellueckl: To implement your suggestion, I got rid of one helper method and inlined it. So I can make that a local variable instead of a class level variable, because that would have been confusing in my opinion. Having something like `charsLeft` and `charsLeftUntilMaxLength` on the class level would make a lot of developers scratch their heads.

src/components/10-atoms/input-text/index.js
raphaellueckl: Uh, are we allowing that? If yes, I will go for it.

raphaellueckl: Eslint forbids it.

raphaellueckl: I wanted to fix it, but `nativeValue` is sometimes `string` and sometimes `number`.

raphaellueckl:
Hi Markus

Please have another look, I addresses all your requests.

src/components/10-atoms/input-text/index.js
```diff
@@ -29,8 +29,20 @@ class AXAInputText extends AXAPopupMixin(NoShadowDOM) {
...
       isReact: { type: Boolean },
+
+      modelCounter: { type: String },
+      counter: { type: String },
+      counterMax: { type: String },
+      maxLength: {
+        type: Number,
+        converter: {
+          toAttribute(value) {
+            // 524288 is the default value according to W3C
```
raphaellueckl: Try removing converter.

src/components/10-atoms/input-text/index.js
```diff
@@ -165,13 +251,14 @@ class AXAInputText extends AXAPopupMixin(NoShadowDOM) {
...
             aria-required="${required}"
+            maxlength="${maxLength}"
             ?disabled="${disabled}"
             @input="${this.handleInput}"
             @focus="${this.handleFocus}"
             @blur="${this.handleBlur}"
           />
           ${
-            checkMark
+            this.showCheckMark()
```
raphaellueckl: remove brackets and add getter syntax above

PR #1352, opened 2019-09-19
Bugfix/testimonials

### LINKED ISSUE #1346: Testimonials: IE only: show-all-inline without margins
the author should have more space to the text above:

image

### LINKED ISSUE #1350: Testimonials: showallinline=false not working
if attribute is set to false, component shows all inline. Thats wrong. The if should test false/true

danyball:
fixes #1350 
fixes #1346


src/components/30-organisms/testimonials/index.js
```diff
@@ -55,7 +57,7 @@ class AXATestimonials extends InlineStyles {
       title,
       subtitle,
       showallinline,
-      autorotatetime,
+      autorotatetime = DEFAULT_AUTOROTATETIME,
```
raphaellueckl: Why a second time here?

src/components/10-atoms/carousel/README.md
```diff
@@ -33,7 +33,7 @@ export default AXACarouselReact;
 ```
 
 ```js
-<AXACarouselReact keysenabled="true">
+<AXACarouselReact keysenabled="">
```
raphaellueckl: keysenabled as an empty string? why?

danyball: true/false here is not correct. i am checking if the attribute is set, and if not, the result will be false. But keysenabled="false" had true as the result.

raphaellueckl: BBut if this only acts as a boolean, you could instead write just `keysenables` without an empty value attached to it.

src/components/10-atoms/carousel/README.md
```diff
@@ -53,7 +53,7 @@ Import the carousel-defining script and use a carousel like this:
     <title>Your awesome title</title>
   </head>
   <body>
-    <axa-carousel keysenabled="true">
+    <axa-carousel keysenabled="">
```
raphaellueckl, line 56: Why empty string here?

src/components/10-atoms/carousel/README.md
```diff
@@ -13,7 +13,7 @@ npm install @axa-ch/carousel
 ```js
 import 'src/components/10-atoms/carousel/lib/index';
 
-<axa-carousel keysenabled="true">
+<axa-carousel keysenabled="">
```
raphaellueckl, line 16: Why empty string here?

src/components/30-organisms/testimonials/README.md
```diff
@@ -14,12 +14,9 @@ npm install @axa-ch/testimonials
 import '@axa-ch/testimonials';
 
 <axa-testimonials
-  autorotatetime="1000"
-  autorotatedisabled="false"
-  keysenabled="true"
-  showallinline="false"
   title="The main title"
   subtitle="This is the subtitle"
+  keysenabled=""
```
raphaellueckl: Why empty string here?

src/components/30-organisms/testimonials/README.md
```diff
@@ -74,12 +68,9 @@ Import the testimonials-defining script and use a testimonials like this:
   </head>
   <body>
     <axa-testimonials
-      autorotatetime="1000"
-      autorotatedisabled="false"
-      keysenabled="true"
-      showallinline="false"
       title="The main title"
       subtitle="This is the subtitle"
+      keysenabled=""
```
raphaellueckl, line 53: Why empty string here?

danyball: This is just a variation. All these possibilities are correct:

> `Correct would be either <div hidden> or <div hidden=""> or <div hidden="hidden"> or case-insensitive and single quotes/unquoted variations of any of them.`

Luca said: for html examples set `=""`and for react examples just the attribute.

src/components/10-atoms/carousel/index.js
```diff
@@ -155,7 +155,8 @@ class AXACarousel extends InlineStyles {
   // AutoRotate:
 
   _startAutoRotate() {
-    if (!this.autorotatedisabled) {
+    // eslint-disable-next-line no-restricted-globals
+    if (!this.autorotatedisabled && !!this.autorotatetime === true && !isNaN(this.autorotatetime)) {
```
01231, line 177: if `this.autorotatetime = true` it would pass this conditional statement

01231: `!!true === true` && `!isNaN(true) === true`

src/components/30-organisms/testimonials/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@axa-ch/testimonials",
-  "version": "0.0.1-alpha.0",
+  "version": "1.0.0",
```
01231, line 3: shouldn't we change version number after a release?

src/components/10-atoms/carousel/README.md
```diff
@@ -13,7 +13,7 @@ npm install @axa-ch/carousel
 ```js
```
01231, line 13: some general comment: in file `index.react.d.ts` of testimonials you forgot to delete row 3 (`type Variant = 'foo' | 'bar';`)

src/components/10-atoms/carousel/index.js
danyball: yes, it should pass if autorotatetime is a number but NOT an emtpy string "" -> do you know a better solution?

src/components/30-organisms/testimonials/package.json
danyball: Because lerna manage the versions? I dont want to forget to set version to 1.0 at release time - so i manually set it. But youre right... not that nice here.

src/components/30-organisms/testimonials/README.md
```diff
@@ -44,17 +21,14 @@ export default AXATestimonialsReact;
...
-  showallinline="false"
   title="The main title"
   subtitle="This is the subtitle"
+  keysenabled
 >
   <span>first slide</span>
   <div>
     <span>second slide</span>
-    <span class="author">Andrew Jackson, Advocate</span>
+    <span className="author">Andrew Jackson, Advocate</span>
```
raphaellueckl: Use a prefix here. @LucaMele We also wanted to discuss what kind of classes we give child elements, if we even should rely on classes. AFAIR in child css we need to reference the tags. Dani can tell you abbout the discussion tomorrow.

src/components/30-organisms/testimonials/story.js
```diff
@@ -45,25 +45,28 @@ story.add('Testimonials', () => {
         >
         <span class="author">Andrew Jackson, Advocate</span>
       </div>
-      <div>
+      <div class="showallinline_ie_bugfix">
```
raphaellueckl: As discusses, this shouldn't have a name like this. @LucaMele  Maybe you can take a look at the bug dani is workarounding here.

src/components/30-organisms/testimonials/README.md
LucaMele: @danyball please use prefix and BEM always for classes. Therefore here something like: `o-testimonials__author` if is an organism or `m-testimonials__author`.

Other option is to set a attribute, which i dont like that much

src/components/30-organisms/testimonials/child.scss
```diff
@@ -1,7 +1,7 @@
 @import '../../10-atoms/carousel/child';
 
 axa-testimonials {
-  span.author{
+  .author{
```
LucaMele: dtto classname is too generic

src/components/30-organisms/testimonials/index.js
```diff
@@ -38,8 +38,8 @@ class AXATestimonials extends InlineStyles {
     super();
 
     // default values props
-    this.title = '';
-    this.subtitle = '';
+    this.title = null;
```
LucaMele: Why from type String to Type Object? I would revert this change. `title` and `subtitle` are set as strings in the props so please use as default value empty string and not an object.

src/components/30-organisms/testimonials/story.js
```diff
@@ -36,7 +36,7 @@ story.add('Testimonials', () => {
       title="${title}"
       subtitle="${subtitle}"
     >
-     <div>
+      <div class="showallinline_ie_bugfix">
```
LucaMele: dtto, name does not follow our css rules: `o-testimonials__showallinline__ie__bugfix`

src/components/30-organisms/testimonials/story.js
LucaMele: dtto, name does not follow our css rules: `o-testimonials__showallinline__ie__bugfix`

src/components/30-organisms/testimonials/react/story.jsx
```diff
@@ -0,0 +1,74 @@
...
+          sorted in 10 minutes after 3 weeks of phone calls. Very helpful once I
+          got through to the correct person but I was constantly being
+          transferred from person to person, in the end I called into my local
+          axa once in the city and got sorted in 10 minutes after 3 weeks of
+          phone calls. Very helpful once I got through to the correct person but
+          I was constantly being transferred from person to person, in the end I
+          called into my local axa once in the city and got sorted in 10 minutes
+          after 3 weeks of phone calls.
+        </span>
+        <span className="author">Andrew Müller, Advocate</span>
```
LucaMele: @danyball here still not BEM class

src/components/30-organisms/testimonials/README.md
raphaellueckl: Still old classname

PR #1381, opened 2019-09-24
[obsolete] endow axa-dropdown with autocomplete functionality

### LINKED ISSUE #1247: Dropdown: in NOT NATIVE MODE: once open, autocomplete should work same as in native mode
In NOT NATIVE MODE: once open, autocomplete should work same as in native mode.

Use case: you have a list with all countries and want to look for italy. Under native dropdown you can open it and start typing It.. and it scrolls down to where Italy is.

markus-walther:
Fixes #1247.

Needs careful review, and crossbrowser-testing.


src/components/20-molecules/dropdown/index.js
```diff
@@ -183,8 +205,46 @@ class AXADropdown extends NoShadowDOM {
     const { key, keyCode } = e;
     if (key === 'Escape' || key === 'Esc' || keyCode === 27) {
       e.preventDefault();
-      this.openDropdown(false);
+      return this.openDropdown(false);
     }
+    return this.handleAutosuggest(key, e);
+  }
+
+  handleAutosuggest(key) {
```
danyball, line 280: Please add a unit test for this method. Its the best documentation for a method of this size.

src/components/20-molecules/dropdown/index.js
```diff
@@ -15,6 +15,9 @@ const ARROW_ICON = svg([ExpandSvg]);
 const DEBOUNCE_DELAY = 250; // milliseconds
 const DROPDOWN_UL_MAXHEIGHT = '200px';
 const EMPTY_FUNCTION = () => {};
+const WORD_END = '\x00';
+const INDEX_END = '\x01';
+const AUTOSUGGEST_INACTIVITY_DELAY = 200; // milliseconds
```
raphaellueckl: Is this the default limit? I fell like this is a bit short. When testing it out, if often did not work, maybe because i was not fast enough.

src/components/20-molecules/dropdown/index.js
```diff
@@ -387,11 +448,20 @@ class AXADropdown extends NoShadowDOM {
...
-    const { select, defaultTitle } = this;
+  updated(changedProperties) {
+    const { select, defaultTitle, items } = this;
     // adjust native <select>
     select.selectedIndex =
       this.findByValue(null, true) + (defaultTitle ? 1 : 0);
+
+    if (changedProperties.has('items')) {
+      // rebuild keyboard-autosuggestion data structures
+      this._autosuggestLetters = new Set();
```
raphaellueckl: Why is this Set created outside of the `buildAutosuggestDictionary`, if that method is the only one making write operations to this set? This makes the code more unreadable, since within that method, you have to check what is actually coming, which is always an empty set. This violates the "single responsibility" principle imho.

src/components/20-molecules/dropdown/index.js
```diff
@@ -183,8 +205,46 @@ class AXADropdown extends NoShadowDOM {
...
+          _autosuggestDictionary.slice(matchIndexPos),
+          10
+        );
+        // ... and its value
+        const { value } = items[firstSelectedItem];
+        // re-render to visually select matched item
+        this.updateItems(value);
+      }
+      // next characters start a *new* autosuggestion
+      this._typedSofar = '';
```
raphaellueckl: typo (camelCase): typedSoFar

src/utils/debounce.js
```diff
@@ -18,7 +18,7 @@ const debounce = function debounce(
    * @function
    */
   function debounced(...rest) {
-    // cash stuff in closure
```
raphaellueckl, line 21: LOOOOOL

src/components/20-molecules/dropdown/index.js
danyball: Yes, i wasnt fast enough many times, too. This feature is a bit tricky. What do you think about a timer on every keystroke? So every stroke has a new timeout and the user is able to type long words without a "global timeout". Or isnt that the default behaviour for dropdowns?

src/components/20-molecules/dropdown/index.js
```diff
@@ -387,11 +448,20 @@ class AXADropdown extends NoShadowDOM {
     window.addEventListener('axa-dropdown-close', this.handleWindowClick);
   }
 
-  updated() {
-    const { select, defaultTitle } = this;
+  updated(changedProperties) {
```
danyball, line 536: I like UnitTests - this method would also be appropriate.

danyball:
If i open the dropdown and select an entry by typing - the dropdown-popup do not change. Is that the expected behaviour? I would have expected only the relevant entries to be displayed.

src/components/20-molecules/dropdown/index.js
markus-walther: This has been modelled after the timing behaviour of native &lt;select&gt; which indeed requires fast typing.

src/components/20-molecules/dropdown/index.js
markus-walther: Done.

markus-walther:
> If i open the dropdown and select an entry by typing - the dropdown-popup do not change. Is that the expected behaviour? I would have expected only the relevant entries to be displayed.

I cannot reproduce this. When I open the styled, non-native dropdown and type fast, I can select entries just the same as when it is closed and focussed. What exactly did you do?

danyball:
> > If i open the dropdown and select an entry by typing - the dropdown-popup do not change. Is that the expected behaviour? I would have expected only the relevant entries to be displayed.
> 
> I cannot reproduce this. When I open the styled, non-native dropdown and type fast, I can select entries just the same as when it is closed and focussed. What exactly did you do?

image

With `popup` i mean the list with all the entries. If i start typing i expect that the entries that do not match are hidden. But not sure if that is the default behaviour.

src/components/20-molecules/dropdown/index.js
markus-walther: Fair point. Was a result of incremental development (I think). Converted `buildAutosuggestDictionary` into a proper method. Done.

src/components/20-molecules/dropdown/index.js
markus-walther: I increased to 300ms. Experiments with small screen widths - where native select is active - on Storybook show that native UX is 'tricky' there as well.

markus-walther: And I don't understand the timer-on-every-keystroke comment: isn't that exactly what happens in handleAutosuggest?

src/components/20-molecules/dropdown/index.js
markus-walther: No. Unit test and documentation are different things, and this method is well-documented.
I'm open to logic/design errors being pointed out in this method, though.

src/components/20-molecules/dropdown/index.js
markus-walther: No, for similar reasons as above.

markus-walther:
> > > If i open the dropdown and select an entry by typing - the dropdown-popup do not change. Is that the expected behaviour? I would have expected only the relevant entries to be displayed.
> > 
> > 
> > I cannot reproduce this. When I open the styled, non-native dropdown and type fast, I can select entries just the same as when it is closed and focussed. What exactly did you do?
> 
> image
> 
> With `popup` i mean the list with all the entries. If i start typing i expect that the entries that do not match are hidden. But not sure if that is the default behaviour.

But you can test the default behaviour in the native &lt;select&gt;, and it is exposed on StoryBook  in small screen width. There no such 'filtering' of unmatched entries happens on typing (presumably because it would be bad UX).

src/components/20-molecules/dropdown/index.js
```diff
@@ -15,6 +15,9 @@ const ARROW_ICON = svg([ExpandSvg]);
 const DEBOUNCE_DELAY = 250; // milliseconds
 const DROPDOWN_UL_MAXHEIGHT = '200px';
 const EMPTY_FUNCTION = () => {};
+const WORD_END = '\x00';
```
LucaMele: Maybe a comment here would be nice

src/components/20-molecules/dropdown/index.js
```diff
@@ -183,15 +187,82 @@ class AXADropdown extends NoShadowDOM {
...
+      this._typedSoFar = '';
+    }, AUTOSUGGEST_INACTIVITY_DELAY);
   }
 
   handleDropdownClick(e) {
     e.preventDefault();
     e.stopPropagation();
     // toggle dropdown
-    this.openDropdown(!this.open);
+    const open = !this.open;
```
LucaMele, line 316: I would rename this to `isOpen` as `open` for me is always a positive value. Just an idea if you see it differently leave it

src/components/20-molecules/dropdown/index.js
danyball: Have no errors found - it works. But we should test all methods with UnitTests if it is possible. Its so much easier to fix things if we have specific tests. But we can discuss that in our next after daily. I look forward to your arguments.

PR #1387, opened 2019-09-27
Add on row click callbacks for table sortable

LucaMele:
fixes #1363

PR #1388, opened 2019-09-27
Feature/image icon search

### LINKED ISSUE #1301: Customer friendly search for images and icons
We need a little search for images and icons, because we have 300+ SVGs. It would be cool to have keywords to each Svg.

01231:
fixes #1301

### DIFF
--- a/src/demo/native/story.js
+++ b/src/demo/native/story.js
@@ -143,13 +143,13 @@ storiesOf('Demos', module)
         }
       </style>
       <h2>
-        Note: There are green border around each SVG to better see dimensions.
+        Note: The green borders reveal the dimensions of the SVGs.
       </h2>
       <div>
         <input
           class="input-field"
           type="text"
-          placeholder="Find"
+          placeholder="Find icon/imag"
           oninput="onCallbackInput(arguments[0])"
         />
       </div>
--- a/src/demo/native/story.js
+++ b/src/demo/native/story.js
@@ -56,8 +56,8 @@ storiesOf('Demos', module)
 
       const renderAreaIcons = document.querySelector('.icons');
       const renderAreaImages = document.querySelector('.images');
-      const iconHeader = document.querySelector('.iconHeader');
-      const imageHeader = document.querySelector('.imageHeader');
+      const iconHeader = document.querySelector('.icon-header');
+      const imageHeader = document.querySelector('.image-header');
 
       filteredIcons = filteredIcons.filter(icon => {
         const { length } = icon.path.split('.svg.js')[0];
@@ -154,7 +154,7 @@ storiesOf('Demos', module)
         />
       </div>
 
-      <h3 class="iconHeader">${icons.length} Icons:</h3>
+      <h3 class="icon-header">${icons.length} Icons:</h3>
       <div class="icons">
         ${svg(
           icons.map(
@@ -166,7 +166,7 @@ storiesOf('Demos', module)
         )}
       </div>
 
-      <h3 class="imageHeader">${images.length} Images:</h3>
+      <h3 class="image-header">${images.length} Images:</h3>
       <div class="images">
         ${svg(
           images.map(
--- a/src/demo/native/story.js
+++ b/src/demo/native/story.js
@@ -17,7 +17,7 @@ const reqSvgsImages = require.context(
 );
 const filepathsImages = reqSvgsImages.keys();
 
-const icons = filepathsIcons.map(path => {
+let icons = filepathsIcons.map(path => {
   return {
     svgstring: reqSvgsIcons(path).default,
     path,
@@ -46,6 +46,31 @@ storiesOf('Demos', module)
 
     const colors = select('color', ['red', 'blue', 'white', 'black'], 'black');
 
+    window.onCallbackInput = ev => {
+      const { value } = ev.target;
+
+      icons = icons.filter(icon => {
+        const { length } = icon.path.split('.svg.js')[0];
+        const lengthOfPrefix = 2;
+        const iconName = icon.path.substr(
+          lengthOfPrefix,
+          length - lengthOfPrefix
+        );
+        const foundSearchTerm = iconName.search(value) > -1;
+        console.log(iconName, foundSearchTerm);
+        return foundSearchTerm ? icon : '';
+      });
+      console.log(icons);
+      console.log('---------------------');
+      const renderArea = document.querySelector('.icons');
+
+      renderArea.innerHTML = icons.map(
+        i =>
+          `<div>${i.svgstring}<span style="padding-left: 10px;">${
+            i.path
+          }</span></div>`
+      );
+    };
     const template = html`
       <style>
         body {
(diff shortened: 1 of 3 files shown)


src/demo/native/story.js
```diff
@@ -59,21 +121,52 @@ storiesOf('Demos', module)
...
-            `<div>${i.svgstring}<span style="padding-left: 10px;">${
-              i.path
-            }</span></div>`
-        )
-      )}
+      <div>
+        <input
+          class="input-field"
+          type="text"
+          placeholder="Find"
```
raphaellueckl: "Find icon"?

src/demo/native/story.js
```diff
@@ -59,21 +121,52 @@ storiesOf('Demos', module)
           width: 40px;
           height: 40px;
         }
+        .input-field {
```
raphaellueckl, line 124: Why not "input-text"?

raphaellueckl (changes requested):

src/demo/native/story.js
01231: "Find icon/image"?

src/demo/native/story.js
01231: Do you mean the name of the class, or why I'm not using AXAInput?

src/demo/native/story.js
```diff
@@ -59,21 +121,52 @@ storiesOf('Demos', module)
...
+          border: 1px solid #00008f;
+        }
+        .input-field:focus {
+          outline: 2px solid #00008f;
+          outline-offset: -2px;
+          color: #000;
+        }
       </style>
       <h2>
         Note: There are green border around each SVG to better see dimensions.
```
raphaellueckl: This sounds a bit weird.

Imho better:
`Note: The green borders reveal the dimensions of the SVGs.`

raphaellueckl (changes requested):

src/demo/native/story.js
raphaellueckl: axainput

src/demo/native/story.js
raphaellueckl: Even better (y)

src/demo/native/story.js
```diff
@@ -46,6 +46,68 @@ storiesOf('Demos', module)
 
     const colors = select('color', ['red', 'blue', 'white', 'black'], 'black');
 
+    window.onCallbackInput = ev => {
```
danyball, line 49: Thats not a good idea. Do not set functions on `window` to not have collisions with other frameworks. Its a good practice to not touch the global window object.

src/demo/native/story.js
```diff
@@ -46,6 +46,68 @@ storiesOf('Demos', module)
```
danyball, line 46: General Comment: What do you think about to generate a small React app for this? Like here `/src/components/10-atoms/button/react/demo.jsx`.

src/demo/native/story.js
```diff
@@ -46,6 +46,68 @@ storiesOf('Demos', module)
...
+        const { length } = image.path.split('.svg.js')[0];
+        const iconName = image.path.substr(
+          lengthOfPrefix,
+          length - lengthOfPrefix
+        );
+        const foundSearchTerm = iconName.search(value.trim()) > -1;
+        return foundSearchTerm ? image : '';
+      });
+
+      renderAreaIcons.innerHTML = filteredIcons
```
danyball, line 82: Can you use variables for SVG-names instead of change the whole innerHTML?

danyball (dismissed):
It works great :) For version 1 of this feature it is quite nice and ready to approve (but please remove `window`-object usage).

Think about to create a react-app to avoid usage of `window` object and to use react render features instead of `innerHTML` manipulating.

src/demo/native/story.js
danyball: That was my ugly english text :D Feel free to change! :)

01231: I will @danyball xD

src/demo/native/story.js
01231: I don't know exactly what you mean

src/demo/native/story.js
01231: Hmm I used this a template -> https://github.com/axa-ch/patterns-library/pull/1387/files#diff-ad77956675a7781a5892ae6169b513a6R75
How is this going to affect other frameworks. It is only a story. So it doesn't pollutes the components core...(as far I know)

src/demo/native/story.js
NoahLiechti: Yeah why not. I'll do it now.

src/demo/native/story.js
```diff
@@ -59,21 +121,52 @@ storiesOf('Demos', module)
...
-            `<div>${i.svgstring}<span style="padding-left: 10px;">${
-              i.path
-            }</span></div>`
-        )
-      )}
+      <div>
+        <input
+          class="input-field"
+          type="text"
+          placeholder="Find icon/imag"
```
raphaellueckl: `Find Icon / Image`

raphaellueckl (changes requested):

src/demo/native/story.js
01231: it would be a hard coupling to a holistic component

raphaellueckl: Why? This is a normal story, like all the other stories are.

01231: Yes, but there are no stories where other  axa components are being used.

LucaMele: I agree with @NoahLiechti. In stories where we dont have lerna support coupling hardly via import `import from axa-input-text` is not good. Is only for demo purposes so its ok as it is IMHO

src/demo/native/story.js
LucaMele: agree generally with @danyball. Here however we have a scoped window object: is inside a story that will never go onto our clients Browser and is inside a `iframe` in our storybook, therefore i would leave it as @NoahLiechti did it.

native `onclick` or `onchange` always require a window function. If you really want to avoid it you have to add a listener. I personally think is a good idea to test here in the story the native `onclick` handler

LucaMele (dismissed):

src/demo/native/story.js
```diff
@@ -59,21 +121,52 @@ storiesOf('Demos', module)
...
+            i =>
+              `<div>${i.svgstring}<span style="padding-left: 10px;">${
+                i.path
+              }</span></div>`
+          )
+        )}
+      </div>
 
-      <h3>${images.length} Images:</h3>
+      <h3 class="imageHeader">${images.length} Images:</h3>
```
raphaellueckl: no camel case for classnames

https://www.freecodecamp.org/news/css-naming-conventions-that-will-save-you-hours-of-debugging-35cea737d849/

raphaellueckl (changes requested):

raphaellueckl (approved):

PR #1394, opened 2019-09-30
Bugfix/datepicker minor bugs

### LINKED ISSUE #1384: Bugfix/datepicker error message remains after valid date is selected
Can be reproduced with any invalid date/string

Bildschirmvideo aufnehmen 2019-09-26 um 16 02 46

### LINKED ISSUE #1386: Bugfix/datepicker size changes when checkmark is set
The size of the input-field increases when setting the checkmark attribute.

Bildschirmfoto 2019-09-27 um 10 00 22
Bildschirmfoto 2019-09-27 um 10 00 15

### LINKED ISSUE #1464: axa-datepicker: cancel/ok button text should not wrap
With long texts, the calendar button texts for Cancel and OK might wrap around.

A case in point would be German 'Abbrechen', and 'Zustimmen', respectively.

Fix this so that texts do not wrap (accepting the risk that overlong texts would introduce new UI glitches).

markus-walther:
Fixes #1384.
Fixes #1386.
Fixes #1464.

Note that Source Sans Pro font must be used.

### DIFF
--- a/src/components/20-molecules/datepicker/index.js
+++ b/src/components/20-molecules/datepicker/index.js
@@ -667,6 +667,9 @@ class AXADatepicker extends NoShadowDOM {
       this.toggleDatepicker();
       input.value = isControlled ? stateValue : value;
     }
+    // since the calendar UI only allows legal dates to be picked,
+    // any preexisting error should be cleared
+    this.error = null;
   }
 
   fireEvents(validDate) {
--- a/src/components/20-molecules/datepicker/index.scss
+++ b/src/components/20-molecules/datepicker/index.scss
@@ -84,6 +84,7 @@ axa-datepicker {
     position: relative;
     display: flex;
     width: 197px;
+    max-width: 226px;
     align-items: center;
   }
 
--- a/src/components/20-molecules/datepicker/index.scss
+++ b/src/components/20-molecules/datepicker/index.scss
@@ -304,6 +304,7 @@ axa-datepicker {
 
   .m-datepicker__button {
     margin: 0;
+    white-space: nowrap;
   }
 
   .m-datepicker__input-icon {


src/components/20-molecules/datepicker/index.scss
```diff
@@ -84,6 +84,7 @@ axa-datepicker {
     position: relative;
     display: flex;
     width: 197px;
+    max-width: 226px;
```
01231, line 87: the input-field still has an other size

src/components/20-molecules/datepicker/index.js
```diff
@@ -667,6 +667,9 @@ class AXADatepicker extends NoShadowDOM {
       this.toggleDatepicker();
       input.value = isControlled ? stateValue : value;
     }
+    // since the calendar UI only allows legal dates to be picked,
```
01231, line 670: works 👍

01231 (changes requested):

src/components/20-molecules/datepicker/index.scss
markus-walther: Really? The button is moving slightly which is hard to fix, but the outer boundary is no longer moving, which your screenshots showed. Please demonstrate!

01231: Input-Field + Button = 217px; datepicker = 217px
Bildschirmfoto 2019-10-02 um 14 19 05
Input-Field + Button = 200px; datepicker = 226px
Bildschirmfoto 2019-10-02 um 14 19 23

Demonstration
Bildschirmvideo aufnehmen 2019-10-02 um 14 13 22

In my opinion input-file + button should be the same size in both examples.

markus-walther: OK, the difference in observation is caused by feature-branch pages *not* installing Source Sans Pro fonts (your case) vs my local Storybook having it installed (as required by designers for this component). Not sure whether component must be made robust against a missing required font.

01231 (approved):
Nice!

PR #1395, opened 2019-10-01
Let table head be black if is not sortable

### LINKED ISSUE #1392: Row headers default in black if not sortable
Row headers default in black if not sortable. 

Links and titles with a functional action should be colored in blue and those which have no action in plain axa black (#333333).

LucaMele:
fixes #1392


01231 (approved):
noice!

PR #1399, opened 2019-10-02
Feature/add knobs cookie disclaimer

### LINKED ISSUE #1398: Knobs for cookie-disclaimer
For a consistent appearance of the pattern-library, the user should have the opportunity so play around with knobs.

01231:
Fixes #1398

### DIFF
--- a/src/components/20-molecules/cookie-disclaimer/story.js
+++ b/src/components/20-molecules/cookie-disclaimer/story.js
@@ -1,31 +1,54 @@
 /* global document */
 import { storiesOf } from '@storybook/html';
+import { text, withKnobs } from '@storybook/addon-knobs';
+import { html, render } from 'lit-html';
 import './index';
 import Readme from './README.md';
 import withNoBorder from '../../../../.storybook/addons/no-border';
 
 storiesOf('Molecules/Cookie disclaimer', module)
   .addDecorator(withNoBorder)
+  .addDecorator(withKnobs)
   .addParameters({
     readme: {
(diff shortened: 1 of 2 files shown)


src/components/20-molecules/cookie-disclaimer/story.js
```diff
@@ -1,31 +1,54 @@
...
-        <p>Any Description for the cookie disclaimer</p>
-        <axa-link variant="arrowright-animated-white" href="https://axa.ch/de/informationen/datenschutz.html">
-          Data protection
-        </axa-link>
-        </axa-cookie-disclaimer>
+  .add('Cookie disclaimer - Default', () => {
+    const buttonname = text('buttonname', 'Akzeptieren');
+    const title = text('text', 'This website uses cookies');
+    const description = text(
+      'description (this is not a direct attribute of cookie disclaimer)',
```
raphaellueckl: I would remove that explanation in the brackets everywhere.

raphaellueckl: I would probably also write `Description` with uppercase since it is not an attribute. But not sure if that is good, up to you.

raphaellueckl (dismissed):

src/components/20-molecules/cookie-disclaimer/story.js
01231: This with uppercase makes sense to me. And I shortened the text in brackets.

raphaellueckl (approved):

PR #1400, opened 2019-10-03
[small]Change carousel text

raphaellueckl:
We should care more about the fact that our work is publicly accessible. I changed a randomly repeatet text into some lorem upsum text.


01231 (commented):
I liked it the other way.
[why we shouldn't use lorem ipsum](https://blog.prototypr.io/why-testing-with-real-content-is-better-than-lorem-ipsum-c7c79586ee72)

raphaellueckl:
@NoahLiechti done

01231 (approved):

PR #1402, opened 2019-10-03
Feature/add knobs input file

### LINKED ISSUE #1401: Knobs for input-file
For a consistent appearance of the pattern-library, the user should have the opportunity so play around with knobs.

01231:
Fixes #1401

### DIFF
--- a/src/components/10-atoms/input-file/react/story.jsx
+++ b/src/components/10-atoms/input-file/react/story.jsx
@@ -6,7 +6,7 @@ import ReactDOM from 'react-dom';
 import AXAInputFileReact from './AXAInputFileReact';
 import Readme from '../README.md';
 
-export const iconOptions = {
+const iconOptions = {
   none: '',
   'arrow-right': 'arrow-right',
   collapse: 'collapse',
--- a/src/components/10-atoms/input-file/react/story.jsx
+++ b/src/components/10-atoms/input-file/react/story.jsx
@@ -1,115 +1,61 @@
 /* global document */
 import { storiesOf } from '@storybook/html';
+import { text, select, boolean, withKnobs } from '@storybook/addon-knobs';
 import React from 'react';
 import ReactDOM from 'react-dom';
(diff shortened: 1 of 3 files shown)


src/components/10-atoms/input-file/react/story.jsx
```diff
@@ -1,116 +1,62 @@
 /* global document */
 import { storiesOf } from '@storybook/html';
+import { text, select, boolean, withKnobs } from '@storybook/addon-knobs';
 import React from 'react';
 import ReactDOM from 'react-dom';
 import AXAInputFileReact from './AXAInputFileReact';
 import Readme from '../README.md';
 
+export const iconOptions = {
```
raphaellueckl: Why is this variable exported? That makes no sense imho.

raphaellueckl (changes requested):

src/components/10-atoms/input-file/story.js
```diff
@@ -1,48 +1,55 @@
 /* global document */
 import { storiesOf } from '@storybook/html';
+import { text, select, boolean, withKnobs } from '@storybook/addon-knobs';
+import { html, render } from 'lit-html';
 import './index';
 import Readme from './README.md';
 
+export const iconOptions = {
```
raphaellueckl: Why is this variable exported? That makes no sense imho.

raphaellueckl (changes requested):

src/components/10-atoms/input-file/react/story.jsx
raphaellueckl: I know now why and we should not do this, because it makes the component a "folder".

src/components/10-atoms/input-file/story.js
raphaellueckl: I know now why and we should not do this, because it makes the component a "folder".

raphaellueckl (approved):
top

PR #1412, opened 2019-10-07
If number is selected, also compare types

LucaMele:
If a model is passed containg values with type number, dropdown crashes


DonaldDuck1337 (approved):

01231 (approved):
Nice!

PR #1428, opened 2019-10-11
Fix footer children for AEM

raphaellueckl:
AEM does not just insert raw children that are being used as slots. AEM puts in a lot of tags and DOM nodes and the children are somewhere within there. We need to unwrap them.


src/components/30-organisms/footer/index.js
```diff
@@ -68,35 +62,59 @@ class AXAFooter extends InlineStyles {
...
-      const { nodeName } = child;
-      return (
-        filter(criteria)(child) && !HEADINGS.includes(nodeName.toLowerCase())
-      );
-    };
+    const noHeaderFilter = criteria => child =>
+      filter(criteria)(child) &&
+      !HEADINGS.includes(child.nodeName.toLowerCase());
 
     const onlyColumns = childrenArr.filter(
```
01231: use `allChildrenWithSlotAttribute`instead of `childrenArr `and delete row 110

01231 (changes requested):

src/components/30-organisms/footer/index.js
raphaellueckl: ah right, I forgot. this was some leftover from the previous version

01231 (approved):

PR #1437, opened 2019-10-14
Feature/cleanup tbd

### LINKED ISSUE #1421: Cleanup 'tbd's in the package.json files
The description should never be 'tbd'. This looks extremely cheap when you search for the component on npm.

raphaellueckl:
Fixes #1421

- We have no longer a "todo" in the package.json scaffolded file in the description, which was checked in like this or with a "tbd" mark.
- Now all descriptions have the same structure.
- The polyfill still has the hint, that it is for IE support.

### DIFF
--- a/scripts/create-a-m-o/functions.js
+++ b/scripts/create-a-m-o/functions.js
@@ -67,6 +67,7 @@ const createFiles = (store, a, m, o, done) => () => {
   templateJson.name = `@axa-ch/${fileName}`;
   templateJson.version = '0.0.0-beta.0';
   templateJson.homepage = `https://github.com/axa-ch/patterns-library/tree/develop/src/components/${folderMap[type]}/${fileName}#readme`;
+  templateJson.description = `The ${fileName} component for the AXA Pattern Library`;
 
   fs.writeFileSync(
     `${BASE_FOLDER}/package.json`,
--- a/scripts/create-a-m-o/templates/template-package.json
+++ b/scripts/create-a-m-o/templates/template-package.json
@@ -1,7 +1,7 @@
 {
   "name": "__",
   "version": "__",
-  "description": "> TODO: description",
+  "description": "The __ component for the AXA Pattern Library",
   "author": "Pattern Warriors",
   "homepage": "__",
   "license": "Copyright 2019 AXA Versicherungen AG",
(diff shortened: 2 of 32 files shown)


scripts/create-a-m-o/functions.js
```diff
@@ -67,6 +67,7 @@ const createFiles = (store, a, m, o, done) => () => {
   templateJson.name = `@axa-ch/${fileName}`;
   templateJson.version = '0.0.0-beta.0';
   templateJson.homepage = `https://github.com/axa-ch/patterns-library/tree/develop/src/components/${folderMap[type]}/${fileName}#readme`;
+  templateJson.description = `The ${fileName} component for the AXA Pattern Library`;
```
01231, line 70: You could also remove the default description of `> TODO: description` in `/patterns-library/scripts/create-a-m-o/templates/template-package.json`

scripts/create-a-m-o/functions.js
```diff
@@ -67,6 +67,7 @@ const createFiles = (store, a, m, o, done) => () => {
   templateJson.name = `@axa-ch/${fileName}`;
```
01231, line 67: General question: Isn't the new description too obvious? I mean you can guess your description from its name. In my opinion the description was pretty okay.

01231 (changes requested):

scripts/create-a-m-o/functions.js
raphaellueckl: hä? das hani jo gmacht

scripts/create-a-m-o/functions.js
raphaellueckl: I noticed a lot of time, that when i searched for a component on npm, it was found with a "subtitle" of "TBD" or "TODO: Add descritpion". I thought it would be best to keep it as simple as obvious possible. A description how to use the component does not belong there, which sometimes also was the case.

In the end, we can still adapt it imho. But having something else than "todo" after scaffolding is important imho.

scripts/create-a-m-o/functions.js
01231: sorry, hani überseh

01231 (approved):

PR #1447, opened 2019-10-16
Feature/fix prod not loading fonts and manifest

raphaellueckl:
On Azure, we need to allow certain MIME types to be able to access them.


01231 (approved):
noice!

PR #1458, opened 2019-10-23
Fix responsive behaviour of axa-fieldset

### LINKED ISSUE #1457: axa-fieldset: responsive behaviour broken for wrapped components
For legitimate technical reasons, applicable components (axa-radio, axa-checkbox) as children of axa-fieldset may be wrapped in one intervening `div` (e.g. in an AEM-forms context).

In this case, at screenwidths below the most narrow breakpoint, components should assume 100 % screen width, whether wrapped or not.

Currently, they don't (in the wrapped case).

Fix this.

markus-walther:
Fixes #1457.


01231 (approved):
Nice!

PR #1466, opened 2019-10-28
Set active link on each rerender

raphaellueckl:
If active was set and the element rerenders, it gets lost. This PR fixes that.

### DIFF
--- a/src/components/20-molecules/footer-small/index.js
+++ b/src/components/20-molecules/footer-small/index.js
@@ -1,6 +1,7 @@
+/* eslint-disable import/no-extraneous-dependencies */
 import { html, css, unsafeCSS } from 'lit-element';
 import { repeat } from 'lit-html/directives/repeat';
-/* eslint-disable import/no-extraneous-dependencies */
+import { classMap } from 'lit-html/directives/class-map';
 import '@axa-ch/container';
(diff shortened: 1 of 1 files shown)


src/components/20-molecules/footer-small/index.js
```diff
@@ -100,16 +91,24 @@ class AXAFooterSmall extends InlineStyles {
...
-                      class="m-footer-small__link--bold"
-                      @click=${ev => this.handleLanguageClick(ev, index)}
-                      >${languageItem.textContent}</a
-                    >
-                  </li>
-                `
+                (languageItem, index) => {
+                  const classes = {
+                    'm-footer-small__link--bold': true,
+                    'm-footer-small__link--active': languageItem.classList.contains(
```
markus-walther, line 97: This line initially  leaves one puzzled, until we understand you are copying a lightDOM element to inside the ShadowDOM, and want to copy relevant classes, too. Comment might be helpful...

markus-walther (approved):
Comment is nice, but not a must.

src/components/20-molecules/footer-small/index.js
raphaellueckl: I will add the comment to the release branch. Because I will lose the approval when I put another commit.

PR #1475, opened 2019-10-31
bugfix/button onclick typo

01231:
Fixes bug, where no forms can be submitted in Edge.

# Done is when (DoD):
- [ ] I have added UI/Unit tests for my changes
- [ ] Design has been reviewed with (here name of reviewer)
- [x] Old tests and eslint rule are not broken
- [ ] My changes are well documented in the readme and showcased in the stories 

# Checklist:
- [ ] I have commented my code, particularly in hard-to-understand areas (focus on **WHY**)

### DIFF
--- a/src/components/10-atoms/button/index.js
+++ b/src/components/10-atoms/button/index.js
@@ -60,7 +60,7 @@ class AXAButton extends LitElement {
     const { style } = this;
     const isNativeShadowDOM = /\{\s*\[native code\]\s*\}/.test(
       // eslint-disable-next-line no-undef
-      ShadowRoot.toString.trim()
+      ShadowRoot.toString().trim()
     );
 
     // shadow dom submit btn workaround
--- a/src/components/10-atoms/button/index.js
+++ b/src/components/10-atoms/button/index.js
@@ -41,7 +41,7 @@ class AXAButton extends LitElement {
     this.icon = '';
     this.motionOff = false;
     this.disabled = false;
-    this.onClick = () => { };
+    this.onClick = () => {};
   }
 
   get isTypeSubmitOrReset() {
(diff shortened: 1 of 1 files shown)


raphaellueckl:
hombre da isch ja alles kabutt, was sölli denn do reviewe?

src/components/10-atoms/button/index.js
```diff
@@ -127,7 +127,7 @@ class AXAButton extends LitElement {
   disconnectedCallback() {
     super.disconnectedCallback();
 
-    if (this.isTypeSubmitOrReset) this.onclick = null;
+    if (this.isTypeSubmitOrReset) this.onClick = null;
```
danyball: a `if` without `{}` is bad practise. Pls add them.

01231: I will🥶

danyball (commented):
Generally: i dont think that this is our problem at AEM forms. It is definitely a typo, but Line 67 is running anyway, isnt it?? Its worth trying.

01231:
So I found out that `onclick` is the event. When the `axa-button` gets clicked it clicks the fake button instead. `onClick` refers to the function-valued attribute. They have nothing in common. It wouldn't make sense that this is a typo, because then the fake-button stuff would only apply to react and other frameworks.

src/components/10-atoms/button/index.js
```diff
@@ -58,13 +58,23 @@ class AXAButton extends LitElement {
 
   firstUpdated() {
     const { style } = this;
+    const isNativeShadowDOM = /\{\s*\[native code\]\s*\}/.test(
```
markus-walther: Too complicated: how about ``${ShadowRoot}`.indexOf('native code') > -1`?

markus-walther (changes requested):

markus-walther (approved):
LGTM

PR #1493, opened 2019-11-08
Fix cursor problem in safari

raphaellueckl:
Hi Markus

This fixes the bug on safari, that when you write into a text field, and move the cursor back to insert an additional word somewhere in between, the cursor would jump to the end again instead of staying at the location.

Reproduce:
- Open Safari
- Open an AEM form (for instance: https://acc.axa.ch/en/formulare/ikrk.html)
- Write `Hello, my name is Robert Robinho`
- Move the cursor to the left and correct the sentence to `Hello, my name is NOT Robert Robinho`

Since you introduced this (which makes sense) to fix another bug and nobody here remembers which one that was, I will await your opinion on this.

### DIFF
--- a/src/components/10-atoms/input-text/index.js
+++ b/src/components/10-atoms/input-text/index.js
@@ -194,15 +194,10 @@ class AXAInputText extends AXAPopupMixin(NoShadowDOM) {
   };
 
   _isSafari() {
-    if (navigator && navigator.vendor && navigator.vendor.includes) {
-      console.log(
-        'navigator vendor: ',
-        navigator.vendor.includes('Apple Computer')
-      );
-    }
     return (
       navigator &&
       navigator.vendor &&
+      navigator.vendor.includes &&
       navigator.vendor.includes('Apple Computer')
     );
   }
--- a/src/components/10-atoms/input-text/index.js
+++ b/src/components/10-atoms/input-text/index.js
@@ -252,8 +252,10 @@ class AXAInputText extends AXAPopupMixin(NoShadowDOM) {
       <div class="a-input-text__input-wrapper">
         <div class="a-input-text__input-elements">
         ${
-          // Safari has a jumping cursor bug, which is fixed with this approach.
-          // Downside: After a user manipulated an input field, it cannot be visibly changed again by javascript.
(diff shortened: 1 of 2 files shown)


src/components/10-atoms/input-text/index.js
```diff
@@ -22,7 +22,7 @@ class AXAInputText extends AXAPopupMixin(NoShadowDOM) {
       label: { type: String },
       required: { type: Boolean },
       placeholder: { type: String },
-      value: { type: String, defaultValue: undefined }, // proper default for controlled-mode under React
+      value: { type: String, reflect: true, defaultValue: undefined }, // proper default for controlled-mode under React
```
LucaMele: Why does it have to reflect? this can be seen as major change. Also this needs to be tested deeply in react

raphaellueckl: My bad, this is not needed

src/components/10-atoms/input-text/index.js
```diff
@@ -251,21 +251,47 @@ class AXAInputText extends AXAPopupMixin(NoShadowDOM) {
...
-            @input="${this.handleInput}"
-            @focus="${this.handleFocus}"
-            @blur="${this.handleBlur}"
-          />
+        ${
+          // On Safari, the caret (cursor) jumps to the end of the value, which
+          // is fixed with this approach.
+          // Downside: After a user manipulated the value manually, it cannot
+          // be updated anymore by javascript (safari only).
+          window.safari
```
LucaMele: please use this instead of window: `(window || global).safari`

LucaMele (approved):
LGTM

LucaMele (changes requested):
Wait, please add `navigator` to the `eslintrc` instead of disable that rule all over the file

src/components/10-atoms/input-text/index.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-undef */
```
LucaMele: @raphaellueckl add `navigator` to `eslintrc` as global instead of disabling a rule for the entire file!

LucaMele (approved):

PR #1496, opened 2019-11-11
[Needs designer approval] Improve breakpoints to fix mobile design

### LINKED ISSUE #1060: axa-container max-width collides with breakpoint
The max width imported into axa-container (from layout.scss) collides with the mobile breakpoint.

max-width: 540px
mobile breakpoints: 576px

This leads to a weird look&feel within that delta. This can be resolved by just removing the import of the max-width in axa-container. It would be best to directly fix it in the layout.scss file.

Pls check for sideeffects when making a change there.

Screenshot 2019-06-04 at 10 46 07

Content should also stretch to the horizontal edges.

raphaellueckl:
Fixes #1060

Please have a look at the issue I wrote. I fixed that by keeping max-height at 100% on the mobile view, but decreased it back to 540px for the first breakpoint after "mobile" (small). Check the footer and adapt the sizes of it to check it. As soon as the feature branch deployment happened, I will add the link.

This should get at least one designer-approval.

This is also what the footer in the old pattern library is doing, as you can see here: https://axa-ch.github.io/patterns-library/components/o-footer/index.html

I tested all the components that we use. The design isn't broken anywhere.

### DIFF
--- a/src/components/00-materials/styles/30-layout.scss
+++ b/src/components/00-materials/styles/30-layout.scss
@@ -18,8 +18,12 @@ $mediaquery-lg: $breakpoint-large ($breakpoint-xlarge - 1px); // (min-width: 992
 $mediaquery-xlg-up: $breakpoint-xlarge; // (min-width: 1200px)
 
 @mixin max-width {
-  max-width: 540px;
+  max-width: 576px;
   margin: auto;
+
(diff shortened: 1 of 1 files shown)


kenaxa:
Design wise it is being displayed correctly ✔️

01231 (approved):

PR #1503, opened 2019-11-12
some browsers don't have the uk locale. defensive is to use just en

LucaMele:
Screenshot 2019-11-12 at 11 16 36

https://en.wikipedia.org/wiki/Locale_(computer_software)


markus-walther (approved):
LGTM

PR #1515, opened 2019-11-15
Bugfix/cursor selection works except double letters

raphaellueckl:
In safari, the cursor was broken.


danyball:
Did you find something about the bug on google?

PR #1524, opened 2019-11-20
Add prefixing with postcss

raphaellueckl:
## Versioning

For us to being able to have multiple versions of components on a webpage, we version the CSS classes. In production, all classes will have a prefix. This prefix makes sure that CSS is targeting only components of the same version and prevent CSS from leaking into other components.

Prefix structure: `nva${COMPONENT_VERSION_BEFORE_PUBLISH}`

**What does that mean?**
`nva` stands for "next version after...". At the time of building, the package.json consists still the version of the previous release. Only upon releasing, our monorepository framework `lerna` will increase the number and publish it. This means, that whenever you see `nva1-2-3_button-link` you have to keep in mind that the version in productions is the version that was released after the number that this prefix is indicating.

**How does the workflow look like with this?**
The developer works on a component as usual. This whole process will not bother during development. As soon as the components gets built, the artifact will contain the prefixes.

**Is there a way to verify that it works before a publish?**
Yes. After the components is built, `npm link` can be run from the components root folder (no the project root folder). After that, anywhere on the computer, this artifact can be used by running `npm link @axa-ch/COMPONENT_NAME`. This will create a symlink directly from your artifact to the destination.

**Currently Remaining Issues:**
- [ ] axa-dropdown is imported from node modules. This breaks its tests now with prefixing, but was a struggle since long ago.
- [ ] axa-fieldset has CSS that relies on css from our components. This should not be possible. This also does not work with the prefixing approach.


scripts/build/rollup-plugin-javascript-prefixer.js
```diff
@@ -0,0 +1,91 @@
...
+const _getComponentInfo = () => {
+  const elementMapping = new Map();
+  elementMapping.set('10-atoms', 'a-');
+  elementMapping.set('20-molecules', 'm-');
+  elementMapping.set('30-organisms', 'o-');
+
+  const cwdAsStringArray = process.cwd().split('/');
+  const customElementName = componentPackageJson.name.replace('@axa-ch/', '');
+  const customElementTypePrefix = elementMapping.get(
+    cwdAsStringArray[cwdAsStringArray.length - 2]
```
danyball: `2` is a "magic number". Set it as a const or better: calculate it.

scripts/build/rollup-plugin-javascript-prefixer.js
```diff
@@ -0,0 +1,91 @@
...
+  };
+};
+
+export const componentInfo = _getComponentInfo();
+
+/**
+ * Look for all classes and replace all that start with the BEM prefix and add
+ * the versioned prefix.
+ *
+ * @param {string} file: Filename, e.g. './index.js'
```
danyball: You dont neet that ":" after file

scripts/build/rollup-plugin-javascript-prefixer.js
```diff
@@ -0,0 +1,91 @@
...
+ */
+const _prefixJS = file => {
+  fs.copyFileSync(file, `${file}.bak`);
```
danyball: What happens to file.bak after running `_prefixJS`? Do you need it?

scripts/build/rollup-plugin-javascript-prefixer.js
```diff
@@ -0,0 +1,91 @@
...
+ * file once, since the result would anyway stay the same.
+ *
+ * @param {number} iterations: Number of iterations
+ */
+export default function jsPrefixer(iterations) {
+  let counter;
+  return {
+    name: 'js-prefixer',
+    buildStart: () => {
+      if (!counter) counter = iterations;
```
danyball: Put {} around your if to better maintenance the code later.

scripts/build/rollup-plugin-javascript-prefixer.js
```diff
@@ -0,0 +1,91 @@
...
+  ); // atom (a-) / molecule (m-) / organism (o-)
+  const newPrefix = `nva${componentPackageJson.version.replace(/\./g, '-')}`;
+  return {
+    prefix: newPrefix, // 'nva1-2-3'
+    standardComponentClassPrefix: customElementTypePrefix + customElementName, // a-button-link
+    cssPrefix: `${newPrefix}_${customElementTypePrefix}${customElementName}`, // .nva1-2-3_a-button-link
+  };
+};
+
+export const componentInfo = _getComponentInfo();
```
danyball: I like to have all the exports at the end of the file to see a overview of all exported methods. But thats minor.

scripts/build/sass-prefixer.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * This is a helper function that prefixes css classes in javascript.
+ *
+ * @param {string} selector: Selector being passed, e.g. '.m-datepicker .m-datepicker__check:after'
+ * @param {object} componentInfo: Object containing componentName and prefixes
+ */
+export default function prefixSass(selector, componentInfo) {
```
danyball: I like to name helper functions like the filename. I was a little bit confused about line 53 in rollup.config, that uses `sassPrefixer()`.

scripts/build/sass-prefixer.js
```diff
@@ -0,0 +1,32 @@
...
+  const splitSelectors = selector.split('.');
```
danyball, line 12: Only classes are relevant? Or what about other css selectors like `#`?

scripts/build/rollup-plugin-javascript-prefixer.js
```diff
@@ -0,0 +1,91 @@
...
+ * @param {number} iterations: Number of iterations
+ */
+export default function jsPrefixer(iterations) {
+  let counter;
+  return {
+    name: 'js-prefixer',
+    buildStart: () => {
+      if (!counter) counter = iterations;
+      if (counter === iterations) {
+        jsFilesToPrefix.forEach(f => {
```
danyball: I like to name such const variables in uppercase: `jsFilesToPrefix`. I was shortly confused where this var is comming from :)

scripts/build/rollup-plugin-javascript-prefixer.js
```diff
@@ -0,0 +1,91 @@
...
+const _prefixJS = file => {
+  fs.copyFileSync(file, `${file}.bak`);
+  const content = fs.readFileSync(file, 'utf8');
+  const afterReplace = content
+    .split(`"${componentInfo.standardComponentClassPrefix}`)
+    .join(`"${componentInfo.cssPrefix}`)
+    .split(`'${componentInfo.standardComponentClassPrefix}`)
+    .join(`'${componentInfo.cssPrefix}`)
+    .split(` ${componentInfo.standardComponentClassPrefix}`)
+    .join(` ${componentInfo.cssPrefix}`);
```
danyball: Is that another way to replace strings? If you want to replace strings please use .replace to better see what you want to do.

scripts/build/rollup-plugin-javascript-prefixer.js
```diff
@@ -0,0 +1,91 @@
...
+  fs.writeFileSync(file, afterReplace);
```
danyball: Is it possible to use more performant (and "better" like node documentation says) async-functions?

danyball (changes requested):

CONTRIBUTION.md
```diff
@@ -157,3 +157,18 @@ storiesOf('Molecules/Top content bar', module)
 - Lerna will update the component `package.json` with new version numbers and auto-commit/push those.
 - Manually undo the removal of all not-to-be-released components in lerna.json, run `npm install` and commit the generated package-lock files to the release branch.
 - Merge the release branch to the `develop` branch in order to preserve the new version numbers.
+
+## Versioning
+
+For us to being able to have multiple versions of components on a webpage, we version the CSS classes. In production, all classes will have a prefix. This prefix makes sure that CSS is targeting only components of the same version and prevent CSS from leaking into other components.
```
01231: For us ~~to~~ being able to

scripts/build/rollup-plugin-javascript-prefixer.js
raphaellueckl: It gets removed. We keep it until all the build configs are done, then we remove it in the `buildEnd` lifecycle of this rollup plugin.

scripts/build/rollup-plugin-javascript-prefixer.js
raphaellueckl: `.replace` is IMHO a bad idea, because there is no "replaceAll" method. You would need 3 while-loops for achieving the same functionality.

scripts/build/rollup-plugin-javascript-prefixer.js
raphaellueckl: Yes, I wanted to do that. I might switch to that in the end. I was not able to use that in the rollup config.

01231 (changes requested):

CONTRIBUTION.md
01231: or `For us to be able to`

scripts/build/rollup-plugin-javascript-prefixer.js
raphaellueckl: I once did that too, since in java you do that for static final variables. But imho it doesn't make sense to uppercase all const variables, this looks so weird, like if it is an environment variable or something like that.

scripts/build/sass-prefixer.js
raphaellueckl: I guess there is no need to prefix IDs. ID's should not be used.

raphaellueckl:
Comments with 👍 will be implemented. Comments without it won't or have an open discussion.
As soon as one is implemented, I will put a 🚀 .

scripts/build/rollup-plugin-javascript-prefixer.js
raphaellueckl: While I understand this wish, I'm not sure about this. If I move it down, I have to declare the const and only at the bottom export it. When you export it where you declare it, it adds to readability. Otherwise you see an export that you need to jump into first to get the context.

scripts/build/rollup-plugin-javascript-prefixer.js
raphaellueckl: Ah sorry, I misunderstood. I agree.

scripts/build/rollup-plugin-javascript-prefixer.js
danyball: Ich bin da wohl etwas schnell drüber gegangen. Jetzt sehe ich, dass die function genau drüber steht. Ja stimmt so macht das Sinn. Aber noch ein anderer Vorschlag dazu:

Du machst eine Methode öffentlich, die `_.....`heisst, also intern sein soll. Warum nicht `_`weglassen und direkt exporten? Also so:
`export const getComponentInfo`

scripts/build/rollup-plugin-javascript-prefixer.js
danyball: Ah and split-join is a workaround for that. Nice. Another way to replaceAll is a replace with a RegExp. But i think thats not a good idea if you have variables.

scripts/build/rollup-plugin-javascript-prefixer.js
danyball: `  const componentDirectoryName = cwdAsStringArray[cwdAsStringArray.length - 2];
`
This part of your code is now better readable. But i mean, why the number "2". Where is this number comming from? Set a comment why a 2 and not a 3 or 4 (do you know what i mean?!). Or better set a const, f.a. `const dasVorletzteElementimArray = 2;`

scripts/build/rollup-plugin-javascript-prefixer.js
danyball: Ah now i see. And why do you need that bak file? Cant you change the original files directly?

scripts/build/rollup-plugin-javascript-prefixer.js
raphaellueckl: I usually use variable names to declare what is the value contained. From the variable name you see that we extract the componentDir from the whole path. Do you think I should be more explicit? Because "2" is just the second last element in the array, which will match to the folder containing the component. Naming the "2", "indexForComponentDir" would probably more confusing, but that is just my opinion. Would love to hear what the others think.

scripts/build/rollup-plugin-javascript-prefixer.js
raphaellueckl: Then I would need an algorithm that reverts it somehow into its original state. That's not the way to do it. But what would be better is having no extra file or use the extra file in the other build steps without manipulating the original one, but I didn't find a way to do that so far. Maybe there is one.

scripts/build/rollup-plugin-javascript-prefixer.js
raphaellueckl: Because the method would then execute all the time, instead of executing once and then always return that result (result is stored in the value).

LucaMele:
@raphaellueckl during the build: why are components changed and reverted back in the `src` folder? this is quite dangerous. If we need to change something in `src`, we have to move them first to a `.tmp` folder. If build brakes unexpectedly it can break our components

LucaMele (changes requested):
As you already noticed, with-react is not working while is the same as index.js

LucaMele:
@raphaellueckl i would welcome to have this feature also in DEV, ergo on the storybook. If not, the risk is that our DEV env drifts too much away from our PROD one, which already does today. Increasing specificity can be a game changer and having a PROD built with higher Specificity then DEV can lead to different behaviour (especially when we start the implementation if pages)

CONTRIBUTION.md
```diff
@@ -18,14 +18,13 @@ Open your code editor of choice and edit your desired component/package in
 
 ## Rules and Lintings
```
LucaMele, line 20: @raphaellueckl why did u remove this? there was no team meeting where we agreed to change our rules

LucaMele: this sentence is important to stay for those people that expect a Bootstrap-like framework.

LucaMele (changes requested):

package.json
```diff
@@ -41,6 +41,7 @@
     "@babel/preset-env": "^7.3.1",
     "@babel/preset-react": "^7.0.0",
     "@babel/runtime": "^7.4.4",
+    "@rollup/plugin-replace": "2.2.1",
```
LucaMele: this doesnt work officially with the rollup version we have ATM:

https://github.com/rollup/plugins/tree/master/packages/replace#requirements

our version is: `rollup@1.16.3`. Please use old replace plugin ` "rollup-plugin-replace": "^2.1.0",` or update rollup and all plugins

rollup.config.js
```diff
@@ -1,13 +1,20 @@
...
+import jsPrefixer, {
+  componentInfo,
+} from './scripts/build/rollup-plugin-javascript-prefixer';
+import sassPrefixer from './scripts/build/sass-prefixer';
+
+const nodeResolve = require('rollup-plugin-node-resolve');
 const commonjs = require('rollup-plugin-commonjs');
 const babel = require('rollup-plugin-babel');
-const replace = require('rollup-plugin-replace'); // use to setup project enviroment variables
+const replace = require('@rollup/plugin-replace'); // use to setup project enviroment variables
```
LucaMele: dtto. officially not compatible with our current rollup version. if it works, is just good luck and can have side effects.

rollup.config.js
```diff
@@ -18,23 +25,34 @@ const globalSassImports = require('./config/globals.js')
   })
   .join('\n');
 
+const shouldRunBuildForReact = fs.existsSync('./index.react.js');
+const amountOfSteps = shouldRunBuildForReact ? 3 : 2;
```
LucaMele: @raphaellueckl why 3 steps for react?

rollup.config.js
```diff
@@ -131,25 +147,22 @@ const dist = {
...
 
-// Sometimes there is no react version to build. I.e icons
-const rollupConfig = [lib];
-if (fs.existsSync('./index.react.js')) {
-  rollupConfig.push(libReact);
-}
-
+rollupConfig.push(lib);
 rollupConfig.push(dist);
+if (shouldRunBuildForReact) rollupConfig.push(libReact);
```
LucaMele: @raphaellueckl respect best practices please and use brackets

scripts/build/rollup-plugin-javascript-prefixer.js
```diff
@@ -0,0 +1,97 @@
...
+  })
+  .map(f => `./${f}`);
+
+const _getComponentInfo = () => {
+  const elementMapping = new Map();
+  elementMapping.set('10-atoms', 'a-');
+  elementMapping.set('20-molecules', 'm-');
+  elementMapping.set('30-organisms', 'o-');
+
+  const cwdAsStringArray = process.cwd().split('/');
```
LucaMele: @raphaellueckl use OS dependent var `path.sep` for PATH splits:  https://nodejs.org/api/path.html#path_path_sep

scripts/build/rollup-plugin-javascript-prefixer.js
```diff
@@ -0,0 +1,97 @@
...
+  const componentDirectoryName = cwdAsStringArray[cwdAsStringArray.length - 2];
```
LucaMele: Looks very hacky. Please use better Node Functions like: https://nodejs.org/api/path.html#path_path_dirname_path

scripts/build/rollup-plugin-javascript-prefixer.js
```diff
@@ -0,0 +1,97 @@
+/* eslint-disable import/no-dynamic-require */
+/* eslint-disable import/no-extraneous-dependencies */
+const glob = require('glob');
+const fs = require('fs').promises;
+
+const componentPackageJson = require(`${process.cwd()}/package.json`);
```
LucaMele: please use OS path agnostic selectors: `path.sep` for concatenation

scripts/build/rollup-plugin-javascript-prefixer.js
```diff
@@ -0,0 +1,97 @@
...
+
+const _getComponentInfo = () => {
+  const elementMapping = new Map();
+  elementMapping.set('10-atoms', 'a-');
+  elementMapping.set('20-molecules', 'm-');
+  elementMapping.set('30-organisms', 'o-');
+
+  const cwdAsStringArray = process.cwd().split('/');
+  const componentDirectoryName = cwdAsStringArray[cwdAsStringArray.length - 2];
+  const customElementName = componentPackageJson.name.replace('@axa-ch/', '');
```
LucaMele: @raphaellueckl why are you taking dependent components from the package json? please document it

scripts/build/rollup-plugin-javascript-prefixer.js
LucaMele: @raphaellueckl why do you make first a private function and then export it? doesn't make sense to me. Please export it directly

Why is it called more then once?

scripts/build/rollup-plugin-javascript-prefixer.js
LucaMele: @raphaellueckl this pieace is not readable at all. Why do some joins have `"` others `'` and why split join split join? This should be refactored

scripts/build/rollup-plugin-javascript-prefixer.js
```diff
@@ -0,0 +1,97 @@
...
+export const componentInfo = _getComponentInfo();
+
+/**
+ * Look for all classes and replace all that start with the BEM prefix and add
+ * the versioned prefix.
+ *
+ * @param {string} file Filename, e.g. './index.js'
+ */
+const _prefixJS = async file => {
+  await fs.copyFile(file, `${file}.bak`);
```
LucaMele: @raphaellueckl if you have to wait anyway, better use SYNC functions of NODE. Is safer

scripts/build/rollup-plugin-javascript-prefixer.js
```diff
@@ -0,0 +1,97 @@
...
+  const content = await fs.readFile(file, 'utf8');
```
LucaMele: dtto. If u need to wait, use sync functions

scripts/build/rollup-plugin-javascript-prefixer.js
```diff
@@ -0,0 +1,97 @@
...
+  const afterReplace = content
+    .split(`"${componentInfo.standardComponentClassPrefix}`)
+    .join(`"${componentInfo.cssPrefix}`)
+    .split(`'${componentInfo.standardComponentClassPrefix}`)
+    .join(`'${componentInfo.cssPrefix}`)
+    .split(` ${componentInfo.standardComponentClassPrefix}`)
+    .join(` ${componentInfo.cssPrefix}`);
+  await fs.writeFile(file, afterReplace);
```
LucaMele: dtto

scripts/build/rollup-plugin-javascript-prefixer.js
```diff
@@ -0,0 +1,97 @@
...
+};
+
+/**
+ * Every config will make the prefixer run again. We only want to parse the
```
LucaMele: @raphaellueckl what do you mean with: Every config? which config?

scripts/build/sass-prefixer.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * This is a helper function that prefixes css classes in javascript.
+ *
+ * @param {string} selector Selector being passed, e.g. '.m-datepicker .m-datepicker__check:after'
+ * @param {object} componentInfo Object containing componentName and prefixes
+ */
+export default function sassPrefixer(selector, componentInfo) {
```
LucaMele: I personally dont like to have magic objects containing everything and nothing and is obfuscated. Better to clearly pass needed params as function paramters so that the function contract is defined clearly

scripts/build/sass-prefixer.js
```diff
@@ -0,0 +1,32 @@
...
+      splitSelectors[i].startsWith(componentInfo.standardComponentClassPrefix)
+    ) {
+      prefixedSelectors.push(`.${componentInfo.prefix}_${splitSelectors[i]}`);
+    }
+    // ".a-button".split('.') == ["", "a-button"]
+    // "a-button" was initially a class because there is an empty string in the
+    // previous place of the array. So if 'i' would be 0, the empty string
+    // would not be there.
+    else if (i !== 0) {
+      prefixedSelectors.push(`.${splitSelectors[i]}`);
```
LucaMele, line 23: does join not do the same? join only adds in between

LucaMele (changes requested):

scripts/build/rollup-plugin-javascript-prefixer.js
raphaellueckl: @LucaMele This is rollup behavior. We run multiple configs. I can show you p2p.

scripts/build/rollup-plugin-javascript-prefixer.js
raphaellueckl: @LucaMele All have the ES6 template string indicator, but I use `"` and `'` as a search pattern.

CONTRIBUTION.md
raphaellueckl: Oh, I forgot about this. This is a statement that is not true. The materials we export are not a webcomponent. Neither is the polyfill. This sentence is imho just confusing and does not really help. But we can discuss in the team.

package.json
raphaellueckl: Good find, thanks!

rollup.config.js
raphaellueckl: I will add a comment here.

scripts/build/rollup-plugin-javascript-prefixer.js
raphaellueckl: I took this idea from sebastian, who did that for midgard.

scripts/build/rollup-plugin-javascript-prefixer.js
raphaellueckl: This method is obsolete now.

scripts/build/rollup-plugin-javascript-prefixer.js
raphaellueckl: This method is obsolete now.

scripts/build/rollup-plugin-javascript-prefixer.js
raphaellueckl: this method is obsolete now.

scripts/build/rollup-plugin-javascript-prefixer.js
raphaellueckl: rollup config, good point!

scripts/build/sass-prefixer.js
raphaellueckl: I don't understand this comment.

rollup.config.js
raphaellueckl: obsolete now

scripts/build/rollup-plugin-javascript-prefixer.js
raphaellueckl: obsolete now

CONTRIBUTION.md
```diff
@@ -181,3 +181,18 @@ storiesOf('Molecules/Top content bar', module)
 - Lerna will update the component `package.json` with new version numbers and auto-commit/push those.
 - Manually undo the removal of all not-to-be-released components in lerna.json, run `npm install` and commit the generated package-lock files to the release branch.
 - Merge the release branch to the `develop` branch in order to preserve the new version numbers.
+
+## Versioning
+
+For us to be able to have multiple versions of components on a webpage, we version the CSS classes. In production, all classes will have a prefix. This prefix makes sure that CSS is targeting only components of the same version and prevent CSS from leaking into other components.
```
markus-walther: spelling: prevents

CONTRIBUTION.md
```diff
@@ -181,3 +181,18 @@ storiesOf('Molecules/Top content bar', module)
...
+
+Prefix structure: `nva${COMPONENT_VERSION_BEFORE_PUBLISH}`
+
+**What does that mean?**
+`nva` stands for "next version after...". At the time of building, the package.json consists still the version of the previous release. Only upon releasing, our monorepository framework `lerna` will increase the number and publish it. This means, that whenever you see `nva1-2-3_button-link` you have to keep in mind that the version in productions is the version that was released after the number that this prefix is indicating.
```
markus-walther: wording: At build time, package.json still contains the version number of the *previous* release.
spelling: productions -> production
emphasis: *after*

CONTRIBUTION.md
```diff
@@ -181,3 +181,18 @@ storiesOf('Molecules/Top content bar', module)
...
+
+**How does the workflow look like with this?**
```
markus-walther: wording: How does the workflow change?

CONTRIBUTION.md
```diff
@@ -181,3 +181,18 @@ storiesOf('Molecules/Top content bar', module)
 - Lerna will update the component `package.json` with new version numbers and auto-commit/push those.
 - Manually undo the removal of all not-to-be-released components in lerna.json, run `npm install` and commit the generated package-lock files to the release branch.
 - Merge the release branch to the `develop` branch in order to preserve the new version numbers.
+
+## Versioning
```
markus-walther: Wording: CSS Class Versioning

CONTRIBUTION.md
```diff
@@ -181,3 +181,18 @@ storiesOf('Molecules/Top content bar', module)
...
+
+For us to be able to have multiple versions of components on a webpage, we version the CSS classes. In production, all classes will have a prefix. This prefix makes sure that CSS is targeting only components of the same version and prevent CSS from leaking into other components.
+
+Prefix structure: `nva${COMPONENT_VERSION_BEFORE_PUBLISH}`
+
+**What does that mean?**
+`nva` stands for "next version after...". At the time of building, the package.json consists still the version of the previous release. Only upon releasing, our monorepository framework `lerna` will increase the number and publish it. This means, that whenever you see `nva1-2-3_button-link` you have to keep in mind that the version in productions is the version that was released after the number that this prefix is indicating.
+
+**How does the workflow look like with this?**
+The developer works on a component as usual. This whole process will not bother during development. As soon as the components gets built, the artifact will contain the prefixes.
```
markus-walther: wording: This whole process -> CSS class versioning only affects the build process. 

the artifact => build artifacts

CONTRIBUTION.md
```diff
@@ -181,3 +181,18 @@ storiesOf('Molecules/Top content bar', module)
...
+
+**Is there a way to verify that it works before a publish?**
```
markus-walther, line 199: wording: ... before publishing?

CONTRIBUTION.md
```diff
@@ -181,3 +181,18 @@ storiesOf('Molecules/Top content bar', module)
...
+Yes. After the components is built, `npm link` can be run from the components root folder (no the project root folder). After that, anywhere on the computer, this artifact can be used by running `npm link @axa-ch/COMPONENT_NAME`. This will create a symlink directly from your artifact to the destination.
```
markus-walther, line 200: wording: this artifact => this component's latest build artifacts

This will create a symlink ...  See [npm link](https://docs.npmjs.com/cli/link.html) for details.

rollup.config.js
```diff
@@ -70,16 +88,14 @@ const lib = {
       exclude: ['node_modules/**'],
       runtimeHelpers: false,
     }),
-    resolve({
+    nodeResolve({
       mainFields: ['module', 'main'],
       only: [/^\.{0,2}\/|\.scss$/i], // threat all node_modules as external apart od .scss files
```
markus-walther: spelling: threat -> treat, apart od -> apart from

scripts/build/rollup-plugin-javascript-prefixer.js
```diff
@@ -0,0 +1,39 @@
+const path = require('path');
```
markus-walther, line 9: file name is too generic, doesn't mention CSS-class-rewriting purpose. If file name stays the same despite this, then I'd expect a multi-line component explaining what this file is doing.

scripts/build/sass-prefixer.js
markus-walther: I tend to agree with @LucaMele here: a functional style with split-map-join would make things much cleared and explain away the edge case of i === 0 in your code. Something like this:

`const prefixedSelectors = splitSelectors.map( selector => `${selector.startsWith(standardComponentClassPrefix) ? prefix + '_' : ''}${selector}`).join('.');`

markus-walther: Btw, `startsWith` is not supported in IE, might consider replacing with `.indexOf(standardComponentClassPrefix) === 0`.

scripts/build/rollup-plugin-javascript-prefixer.js
```diff
@@ -0,0 +1,39 @@
...
+/* eslint-disable import/no-dynamic-require */
+const componentPackageJson = require(`${process.cwd()}${path.sep}package.json`);
+const elementMapping = new Map();
+elementMapping.set('10-atoms', 'a-');
+elementMapping.set('20-molecules', 'm-');
+elementMapping.set('30-organisms', 'o-');
+
+const cwdAsStringArray = process.cwd().split(path.sep);
+const parentDirIndex = cwdAsStringArray.length - 2;
```
markus-walther: Why `-2`? Needs explanation and/or appropriately preferrably symbolic constant. Magic numbers are generally to be replaced by all-caps constants in my opinion.

markus-walther (changes requested):
Good, minor changes in documentation and code notwithstanding.

I'm still a bit uneasy about string rewriting of JS source code (instead of proper parsing), but time will tell whether that's an issue in practice for our use case.

rollup.config.js
raphaellueckl: Not my code, but I changed it to:

`treat all node_modules as external, except .scss files`

scripts/build/rollup-plugin-javascript-prefixer.js
raphaellueckl: This is just the `path`-module from nodejs, nothing more.

raphaellueckl:
Since we decided to not implement that feature (too little advantages over the complexity of string-parsing), I will close this PR and leave the branch open for future reference. The approach will be to version the webcomponent-tags somehow (instead of the css classes), probably in a similar fashion like here.

PR #1536, opened 2019-11-26
Implement universal React wrapper and adapt components to it

markus-walther:
This needs a thorough review.

Code has been simplified and cleaned up now.
React `displayName` works + got a test.

No defaults for properties are applied anymore, since that doesn't seem necessary.


src/utils/with-react.js
```diff
@@ -73,4 +79,61 @@ const applyDefaults = ceInst => {
...
+  let map;
+  // iterate over all properties
+  Object.keys(properties).forEach(name => {
+    let value = properties[name];
+    // classify property by type to select correct map object
+    // (note that unregistered properties are classified as attr(ibute)s via their undefined .type)
+    const event = name.indexOf('on') === 0 && Function; // Function if onXXX event name, false otherwise
+    const className = name === 'className' && name;
+    const defaultType = { type: event };
+    const declaredType = (componentClass.properties[name] || defaultType).type;
```
LucaMele: @markus-walther please do not use `( ... ).type` as its not very readable unless you know how the JS Engine works on prioritisation and order of execution. IMHO not everyone knows it deeply

src/utils/with-react.js
```diff
@@ -73,4 +79,61 @@ const applyDefaults = ceInst => {
...
+    const type = event || className || declaredType;
```
LucaMele: @markus-walther please rewrite this in a more readable way. Too much logic on one line and too much underneath JS know how is required. Also relying so hard on falsy / truly can be risky and hard to debug if written like this.

src/utils/with-react.js
LucaMele: I honestly think is sexy like this, but im pretty sure that is not very readable for everyone. I will ask for other peoples opinion

src/utils/with-react.js
LucaMele: I honestly think is sexy like this, but im pretty sure that is not very readable for everyone. I will ask for other peoples opinion

src/utils/with-react.js
```diff
@@ -73,4 +79,61 @@ const applyDefaults = ceInst => {
 
 export { applyDefaults };
 
-export default createElement => val(createElement);
+const distributeProperties = ({ children, ...properties }, componentClass) => {
```
LucaMele: @markus-walther you are deconstructioning children and as return you are creating a new object with it. Why not deconstructing it directly here: 

IS: 
```
const reactStatelessComponent = properties => {
    const { attrs, props, children } = distributeProperties(
      properties,
      componentClass
    );

    return val(createElement)(
      tagName,
      { isReact: true, attrs, ...props },
      children
    );
  };
```

SUGGESTED: 
```
const reactStatelessComponent = ({children, ...properties}) => {
    const { attrs, props } = distributeProperties(
      properties,
      componentClass
    );

    return val(createElement)(
      tagName,
      { isReact: true, attrs, ...props },
      children
    );
  };
```

LucaMele (changes requested):

src/utils/with-react.js
markus-walther: Arguably the complexity is the same, no lines have been saved. But in the original solution the name distributeProperties is honoured more, since incoming children is a property that only via that function gets a special role.

src/utils/with-react.js
```diff
@@ -73,4 +79,61 @@ const applyDefaults = ceInst => {
...
+  const attrs = {};
+  const props = {};
+  let map;
+  // iterate over all properties
+  Object.keys(properties).forEach(name => {
+    let value = properties[name];
+    // classify property by type to select correct map object
+    // (note that unregistered properties are classified as attr(ibute)s via their undefined .type)
+    const event = name.indexOf('on') === 0 && Function; // Function if onXXX event name, false otherwise
+    const className = name === 'className' && name;
```
danyball: For me the `&&` in these 2 lines are not very readable. I like classic `if()` statements.

src/utils/with-react.js
danyball: For me that is readable. If there is a `).type` i am checking the paranthes. And the logic there is not very compley. So its ok for me.

src/utils/with-react.js
danyball: Spontan würde man alle 3 Werte gleichberechtigt sehen. Was sie aber nicht sind. Sehe das auch als gefährlich zum Warten. Evtl. einfach die ( ) Klammern dort setzen wo es auch die Engine machen würde.

src/utils/with-react.js
```diff
@@ -3,6 +3,12 @@ import val from '@skatejs/val';
 // defined values are different from undefined (for properties) or null (for attributes)
 const isDefined = value => !(value === undefined || value === null);
 
+const pascalCase = hyphenatedName =>
+  hyphenatedName
+    .split('-')
+    .map(part => part.charAt(0).toUpperCase() + part.slice(1))
+    .join('');
```
raphaellueckl, line 10: Nice!

src/utils/with-react.js
```diff
@@ -73,4 +79,61 @@ const applyDefaults = ceInst => {
...
+  // initialize
+  const attrs = {};
+  const props = {};
+  let map;
+  // iterate over all properties
+  Object.keys(properties).forEach(name => {
+    let value = properties[name];
+    // classify property by type to select correct map object
+    // (note that unregistered properties are classified as attr(ibute)s via their undefined .type)
+    const event = name.indexOf('on') === 0 && Function; // Function if onXXX event name, false otherwise
```
raphaellueckl: Clever approach! Although I wonder if it could not be a trap if something else starts with 'on'. I could be wrong though.

raphaellueckl (commented):
The code itself looks readable to me. I just don't know all the internals that react needs for this feature to work.

src/utils/with-react.js
markus-walther: done

src/utils/with-react.js
markus-walther: done via if's, satisfies the concern in another way

src/utils/with-react.js
markus-walther: Thanks

src/utils/with-react.js
markus-walther: Tightened it via checking it continues with uppercase letter, making the 'trap' more unlikely as per React onXXX conventions.

src/utils/with-react.js
markus-walther: done, rewritten via if's

src/utils/with-react.js
markus-walther: done

LucaMele:
This demo is not working anymore:

http://localhost:6006/?path=/story/demos--controlled-inputs-react
http://localhost:6006/?path=/story/demos--dynamic-children-react

Screenshot 2019-12-03 at 15 58 16

LucaMele:
Im not sure this is an error: Why 3 times AxaTestimonials?

Screenshot 2019-12-03 at 16 10 25

LucaMele:
Freeze seems not working anymore (dropdown option changes regardless of freeze):

Screenshot 2019-12-03 at 16 14 27

src/utils/with-react.js
LucaMele: @markus-walther: Is not about complexity rather de-coupling. The advantage of my suggestion is that distributeProperties does not get polluted with props it doesn't need. In your case you add a bidirectional dependency where is not needed (distributeProperties expects children in the interface and caller expect children as return key). I would like that you refactor to my suggestion

markus-walther:
> This demo is not working anymore:
> 
> http://localhost:6006/?path=/story/demos--controlled-inputs-react
> http://localhost:6006/?path=/story/demos--dynamic-children-react
> 
> Screenshot 2019-12-03 at 15 58 16

Fixed via commit 3f6ea305e4d44a24c2c9e5751828798cce9a1a3d.

src/utils/with-react.js
LucaMele: And please don't mark as resolved when its not resolved.

LucaMele: If you disagree with my suggestion, lets ask the team

markus-walther:
> Im not sure this is an error: Why 3 times AxaTestimonials?
> 
> Screenshot 2019-12-03 at 16 10 25

No idea. DOM doesn't have the duplication. Possibly a bug in the Chrome DevTools plugin or an interop issue when custom elements are involved?

LucaMele:
> > Im not sure this is an error: Why 3 times AxaTestimonials?
> > Screenshot 2019-12-03 at 16 10 25
> 
> No idea. DOM doesn't have the duplication. Possibly a bug in the Chrome DevTools plugin or an interop issue when custom elements are involved?

All right ok.. lets hope is that

LucaMele:
@markus-walther also we cannot forget to add README's migration notes. I would prefer to do it now, but also can be done after merge and before release. Very nice job with this PR! Thank you

LucaMele (approved):
YAAAAY.. lets alpha release it first but looks awesome

PR #1557, opened 2019-12-12
Bugfix/button does not trigger

LucaMele:
Fixes some issues after with-react merge

### DIFF
--- a/src/components/10-atoms/button/index.js
+++ b/src/components/10-atoms/button/index.js
@@ -100,8 +100,12 @@ class AXAButton extends LitElement {
       this.fakeButton.click();
     }
 
+    // If we are under react, onClick will be camel Case onClick. If so, use it
+    // otherwise if a consumer defined a onclick="fn", call that instead
     const onclick = this.onClick || this.originalOnclick;
 
+    // if click event is fired manually via javascript, the this.onclick = e => { function
+    // will be called and therefore make sure to not trigger it again.
     if (!ignoreOnClick && typeof onclick === 'function') {
       onclick(e);
     }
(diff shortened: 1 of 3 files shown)


src/components/10-atoms/button/index.js
```diff
@@ -81,11 +81,11 @@ class AXAButton extends LitElement {
     // If someone fires a click on the button and its type is submit then trigger fake button
     // press
     this.onclick = e => {
-      this.handleClick(e);
+      this.handleClick(e, true);
     };
   }
 
-  handleClick = e => {
+  handleClick = (e, ignoreOnClick = false) => {
```
danyball: I was a bit confused about that attribute `ignoreOnClick`. A function "handleClick" with an attribute "ignoreOnClick" is worthless?! Perhaps a more accurate name would be better to read.

src/components/10-atoms/button/index.js
```diff
@@ -95,6 +95,10 @@ class AXAButton extends LitElement {
       e.stopPropagation();
       this.fakeButton.click();
     }
+
+    if (!ignoreOnClick && typeof this.onClick === 'function') {
+      this.onClick(e);
```
danyball: You are calling `onClick` - is this the same method like in line 83 `onclick`? If so, a endless circuit could happen: Line 84 calls handleclick -> handleclick calls onclick on line 100 -> .......

danyball (changes requested):

src/components/10-atoms/button/index.js
LucaMele: @danyball unfortunately is necessary. the comment:    `// If someone fires a click on the button and its type is submit then trigger fake button` should explain why

danyball (approved):
I run that code with storybook > react example. button works. I dont know the bug, but your result works. So its fine for me ;)

LucaMele:
thank you very much

PR #1575, opened 2019-12-20
Feature/axa heading

### LINKED ISSUE #1088: Refactor Title & Text
Title isn a `article` atm. Should be a `h*` instead

raphaellueckl:
fixes #1088 

- [x] Implement axa-heading
- [x] Write first UI Test
- [x] Add UI-Tests for all screen sizes and variants
- [ ] ~Write better stories with knobs~ Not needed.
- [x] Maybe fix design incongruencies. @TommyHoliday told me that we could discuss it. I took the style from the old axa-title and axa-title-secondary. He has a slightly adapted version on figma: https://www.figma.com/file/6zurYk3bJpzUg0H2THSxGF/AXA-UI-Kit?node-id=0%3A58 - Status: We will implement the changes later.

### DIFF
--- a/src/components/10-atoms/heading/README.md
+++ b/src/components/10-atoms/heading/README.md
@@ -2,7 +2,7 @@
 
 [Changelog](./CHANGELOG.md)
 
-TODO Description
+The title component provides a way to conveniently use headings and titles, that are in line with the style guide. Keep in mind that you need to import the fonts separately.
 
 ## Usage
 
@@ -15,7 +15,7 @@ npm install @axa-ch/heading
 ```js
 import '@axa-ch/heading';
 ...
-<axa-heading></axa-heading>
+<axa-heading rank="1">H1 Heading</axa-heading>
 ```
 
 ### React
@@ -32,7 +32,7 @@ export default AXAHeadingReact;
 ```
 
 ```js
-<AXAHeadingReact onClick={handler} />
+<AXAHeadingReact rank="1">H1 Heading</AXAHeadingReact>
 ```
 
 ### Pure HTML pages
@@ -49,7 +49,7 @@ Import the heading-defining script and use a heading like this:
     <title>Your awesome title</title>
   </head>
   <body>
-    <axa-heading></axa-heading>
+    <axa-heading rank="1">H1 Heading</axa-heading>
     <script src="node_modules/@axa-ch/heading/dist/index.js"></script>
   </body>
 </html>
@@ -59,18 +59,7 @@ Import the heading-defining script and use a heading like this:
 
 ### Variant
 
-| Attribute       | Details         |
-| --------------- | --------------- |
-| `variant="foo"` | Desc of Variant |
-
-### Bar
-
-The attribute `bar` specifies...
-
-### onClick
-
-The function-valued attribute `onClick` can be used as a callback prop for React and other frameworks.
-
-### Migration Notes
-
-You don't have to pay attention to anything for upgrading to newer version.
+| Attribute             | Details                                                                                                                        |
+| --------------------- | ------------------------------------------------------------------------------------------------------------------------------ |
+| `variant="secondary"` | Secondary variant will use publico as font. Default: "primary"                                                                 |
+| `rank="1"`            | Rank holds a number from one to six, which represents the underlying html heading tag. `rank="1"` will lead to a `h1` element. |
--- a/src/components/10-atoms/heading/README.md
+++ b/src/components/10-atoms/heading/README.md
@@ -1,0 +1,76 @@
+# Heading
+
+[Changelog](./CHANGELOG.md)
+
+TODO Description
+
+## Usage
+
+**Important:** If this component needs to run in Internet Explorer 11, [you need to use our polyfill](https://github.com/axa-ch/patterns-library/tree/develop/src/components/05-utils/polyfill).
+
+```bash
+npm install @axa-ch/heading
+```
+
+```js
+import '@axa-ch/heading';
+...
+<axa-heading></axa-heading>
+```
+
+### React
+
+Create a React-ified heading with the createElement function from your React version and then use it like this:
+
+```js
+import { createElement } from 'react';
+import createAXAHeadingReact from '@axa-ch/heading/lib/index.react';
+
+const AXAHeadingReact = createAXAHeadingReact(createElement);
+
+export default AXAHeadingReact;
+```
+
+```js
+<AXAHeadingReact onClick={handler} />
+```
+
+### Pure HTML pages
+
+Import the heading-defining script and use a heading like this:
+
+```html
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <title>Your awesome title</title>
+  </head>
+  <body>
+    <axa-heading></axa-heading>
+    <script src="node_modules/@axa-ch/heading/dist/index.js"></script>
+  </body>
+</html>
+```
+
+## Properties
+
+### Variant
+
+| Attribute       | Details         |
+| --------------- | --------------- |
+| `variant="foo"` | Desc of Variant |
+
+### Bar
+
+The attribute `bar` specifies...
+
+### onClick
+
+The function-valued attribute `onClick` can be used as a callback prop for React and other frameworks.
+
+### Migration Notes
+
+You don't have to pay attention to anything for upgrading to newer version.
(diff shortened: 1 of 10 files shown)


danyball:
Is this a fix for #1088 ? If so, pls assign to issue.

src/components/10-atoms/heading/index.js
```diff
@@ -0,0 +1,90 @@
...
+    // the HTML attribute has been set before defining the custom element
+    applyDefaults(this);
+  }
+
+  render() {
+    const classes = classMap({
+      'a-heading--secondary': this.variant === 'secondary',
+    });
+
+    switch (this.rank) {
```
danyball: Cant you write something like

`
          <h${rank} class="a-heading ${classes}">
            <slot></slot>
          </h${rank}>
        `

src/components/10-atoms/heading/index.js
```diff
@@ -0,0 +1,90 @@
...
+            <slot></slot>
+          </h6>
+        `;
+    }
+  }
+
+  disconnectedCallback() {
+    super.disconnectedCallback();
+
+    // Cleanup and reset (i.e event listeners)
```
danyball: Delete line 83 and 84

src/components/10-atoms/heading/index.react.d.ts
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+export type Variant = 'foo' | 'bar';
```
danyball: replace foo and bar

src/components/10-atoms/heading/index.react.d.ts
```diff
@@ -0,0 +1,16 @@
...
+export type Rank = 1 | 2 | 3 | 4 | 5 | 6;
+
+export interface AXAHeadingProps {
+  className?: string;
```
danyball: this is not a prop in your component isnt it?!

danyball (changes requested):

src/components/10-atoms/heading/index.react.d.ts
01231: Please delete :)

01231: or use primary and secondary

src/components/10-atoms/heading/story.js
```diff
@@ -0,0 +1,37 @@
...
+    <axa-heading rank="4">H4 Primary Heading</axa-heading>
+    <axa-heading rank="5">H5 Primary Heading</axa-heading>
+    <axa-heading rank="6">H6 Primary Heading</axa-heading>
+
+    <axa-heading variant="secondary" rank="1">H1 Secondary Heading</axa-heading>
+    <axa-heading variant="secondary" rank="2">H2 Secondary Heading</axa-heading>
+    <axa-heading variant="secondary" rank="3">H3 Secondary Heading</axa-heading>
+    <axa-heading variant="secondary" rank="4">H4 Secondary Heading</axa-heading>
+    <axa-heading variant="secondary" rank="5">H5 Secondary Heading</axa-heading>
+    <axa-heading variant="secondary" rank="6">H6 Secondary Heading</axa-heading>
```
01231, line 32: I would suggest using a loop?

src/components/10-atoms/heading/index.js
```diff
@@ -0,0 +1,90 @@
...
+    applyDefaults(this);
+  }
+
+  render() {
+    const classes = classMap({
+      'a-heading--secondary': this.variant === 'secondary',
+    });
+
+    switch (this.rank) {
+      case 1:
```
01231: I would use a loop here. So that you don't have to write 6x the same stuff

src/components/10-atoms/heading/ui.test.js
```diff
@@ -0,0 +1,646 @@
...
+  .page(`${host}/iframe.html?id=atoms-heading--heading`)
+  .afterEach(async t => {
+    await t.maximizeWindow();
+  });
+
+const TAG = 'axa-heading';
+
+const mobileBreakpoint = 767;
+const tabletBreakpoint = 991;
+const desktopBreakpoint = 992;
```
01231, line 15: breakpoint are global accessible aren't they?

01231 (changes requested):

src/components/10-atoms/heading/index.js
raphaellueckl: I tried that (and similar ways) and it did not work, that's why I approached it like that. Somehow lit-html does not accept interception outside of attribute values. But I agree, I should maybe take another look at this.

raphaellueckl: I can't make it to work.

src/components/10-atoms/heading/story.js
raphaellueckl: Imho this is more readable, since it's a declarative approach. More opinions on this are welcome though. Good input!

This would be the result. Unreadable and doesn't save any lines. Could be optimized, but If this doesn't become less than three lines, it's imho not worth the effort.

```
storyAXAHeading.add('Heading', () => {
  const wrapper = document.createElement('div');

  const template = [];
  for (let i = 1; i < 7; ++i) {
    template.push(html`
      <axa-heading rank="${i}">H${i} Primary Heading</axa-heading>
    `);
  }
  for (let i = 1; i < 7; ++i) {
    template.push(html`
      <axa-heading rank="${i}" variant="secondary"
        >H${i} Primary Heading</axa-heading
      >
    `);
  }

  render(template, wrapper);
  return wrapper;
});
```

src/components/10-atoms/heading/ui.test.js
raphaellueckl: Yes, but only in the scss context. We could think about adding a global testfile-variables file though.

src/components/10-atoms/heading/index.js
raphaellueckl: This is impossible as I see it. Or do you have an idea how to approach that?

src/components/10-atoms/heading/story.js
01231: You are right. It's definitely more readable.

src/components/10-atoms/heading/ui.test.js
01231: 👍

src/components/10-atoms/heading/index.js
01231: No not really, I think there would be the same problem as with the approach of Dany.

src/components/10-atoms/heading/index.js
danyball: ok.

src/components/10-atoms/heading/README.md
```diff
@@ -0,0 +1,65 @@
+# Heading
+
+[Changelog](./CHANGELOG.md)
```
danyball: Pls remove that line. We dont need it if we merged the "landingpage" branch.

src/components/10-atoms/heading/ui.test.js
```diff
@@ -0,0 +1,646 @@
...
+  await t.expect($axaElemShadowEl.getStyleProperty('font-weight')).eql('400');
+  await t.expect($axaElemShadowEl.getStyleProperty('font-style')).eql('normal');
+  await t.expect($axaElemShadowEl.getStyleProperty('font-size')).eql('16px');
+  await t.expect($axaElemShadowEl.getStyleProperty('line-height')).eql('18px');
+  await t
+    .expect($axaElemShadowEl.getStyleProperty('letter-spacing'))
+    .eql('-0.16px');
+}).before(async t => {
+  await t.resizeWindow(mobileBreakpoint, mobileBreakpoint);
+});
```
danyball, line 646: wow, a huge test file, nice ;)

src/components/10-atoms/heading/index.js
LucaMele: @raphaellueckl @NoahLiechti Agree, remove the switch case please

Can you create the element with `document.createElement`?

01231 (dismissed):
Nice!

LucaMele (changes requested):

src/components/10-atoms/heading/index.js
LucaMele: There must be a way.

LucaMele: Or somthing like

```
const opening = `<${rank} class="a-heading ${classes}">`;
const closing = `</${rank}>`;

html`${opening}<slot></slot>${closing}`
```

?

raphaellueckl: According to this question, there is not: https://stackoverflow.com/questions/59259400/is-there-a-way-to-make-tags-in-lit-html-templates-dynamic

but maybe we can find a workaround

LucaMele (approved):

src/components/10-atoms/heading/index.js
LucaMele: We found a workaround.

danyball (approved):

PR #1592, opened 2020-01-09
Remove title primary and title secondary

### LINKED ISSUE #1589: Refactor published axa-title and axa-title-secondary
Luca Mele:
========

Action plan:

* Update Readme on NPM: https://www.npmjs.com/package/@axa-ch/title-section to something like: Deprecated, please check out new repo: https://www.npmjs.com/package/@axa-ch/title

* ~Refactor title-primary and title-secondary to only title having variant primary and secondary and size in 1,2,3,4,5,6. -> Reason: visible channel should always match with semantic. Big fat title is a h1. If we need to give more importance to a lower heading title than a higher one, we add a readme to explain how to wrap lower, sub section title in `<article>, <section>, <nav>, <aside>, <axa-container>`~

* ~~for legacy system, allow H overriding with flag: `hoverride`~~

* ~~`axa-text` should not wrap with `<span>` and should be NoShadow -> Reason: span is not correct as users could pass `<p>` or other tags that are not allowd inside span. We decided therefore no wrapper. No shadow was the idea that outside application most likely have to style inner text with colors, etc. Therefore noShadow~~ -> fixed with #1136

* ~~Add styling into mixing.~~ -> done already

REFs:
- https://html.spec.whatwg.org/#headings-and-sections
- https://html.spec.whatwg.org/#sectioning-content-2
- https://html.spec.whatwg.org/#outlines

_Originally posted by @LucaMele in https://github.com/axa-ch/patterns-library/issues/1088#issuecomment-503011701_

raphaellueckl:
Fixes #1589


LucaMele (approved):
LGTM

01231 (approved):

danyball (approved):

PR #1618, opened 2020-01-15
Update DOM on update children WC

### LINKED ISSUE #1615: Checkbox not rerendering child label
Version: "@axa-ch/checkbox": "^2.1.1"

Axa checkbox does not rerender child node when child is changed.
Initial render is correct but subsequent changes are not rendered correctly.
This bug only happens when label is provided as child node, behavior is correct if label is passed as prop.

**Reproduction link:**
[https://codesandbox.io/s/eloquent-cdn-ebhnq](https://codesandbox.io/s/eloquent-cdn-ebhnq)

**Buggy behavior:**
Test label as child
- label shows AAAAAA (initial)
- label value is changed to BBBBBB
-  label shows AAAAAA

**Expected behavior:**
Test label as child
- label shows AAAAAA (initial)
- label value is changed to BBBBBB
-  label shows BBBBBB

LucaMele:
Fixes #1615.

# Done is when (DoD):
- [x] I have added UI/Unit tests for my changes
- [x] Old tests and eslint rule are not broken
- [ ] I added modified component properties to the typescript typings
- [ ] Design has been reviewed with (here name of reviewer)
- [x] My changes are accordingly to our [Best Practices](https://github.com/axa-ch/patterns-library/blob/develop/CONTRIBUTION.md#best-practices)
- [ ] My changes are well documented in the README and showcased in the stories 
- [x] My changes are shortly added to CHANGELOG.md with the issue number
- [x] Tested with IE

# Checklist:
- [x] I have commented my code, particularly in hard-to-understand areas (focus on **WHY**)
- [x] I have made/updated the ChangeLog Section in the README

### DIFF
--- a/src/components/10-atoms/checkbox/index.js
+++ b/src/components/10-atoms/checkbox/index.js
@@ -215,6 +215,15 @@ class AXACheckbox extends NoShadowDOM {
       this.querySelector('input').checked = true;
       this.state.native = true;
     }
+
+    // install observer to watch for changes in the component's children,
+    // rerendering if any such change occurs
+    this._observer = new MutationObserver(() => this.requestUpdate());
+    this._observer.observe(this, {
+      attributes: true,
+      childList: true,
+      subTree: true,
+    });
   }
 
   // workaround because react has no innerHTML in constructor
@@ -265,6 +274,11 @@ class AXACheckbox extends NoShadowDOM {
     } else {
       this.removeAttribute('checked');
     }
+  }
+
+  disconnectedCallback() {
+    // remove installed observer
+    this._observer.disconnect();
   }
 }
 
--- a/src/components/10-atoms/checkbox/index.js
+++ b/src/components/10-atoms/checkbox/index.js
@@ -82,8 +82,13 @@ class AXACheckbox extends NoShadowDOM {
   wrapChildren() {
     const childWrapper = document.createElement('axa-text');
     childWrapper.variant = 'size-3';
-    childWrapper.className = 'a-checkbox__children-inline';
-    childWrapper.innerHTML = this.innerHTML;
+    childWrapper.className = 'a-checkbox__children-inline js-checkbox__children-inline';
+
+    // Clone live DOM Node so that we dont have to use mutation obersver which causes a
+    // deadlock because we dont have shadow dom here and it would infintily update - render - update
+    // We will set content of live dome AFTER render is done in order that dom nodes are aligned.
+    // Only needed if we have childs.
+    [ ...this.children ].forEach((el) => childWrapper.appendChild(el));
 
     this.labelTextElement = childWrapper;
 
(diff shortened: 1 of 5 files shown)


src/components/10-atoms/checkbox/index.js
```diff
@@ -83,7 +83,12 @@ class AXACheckbox extends NoShadowDOM {
     const childWrapper = document.createElement('axa-text');
     childWrapper.variant = 'size-3';
     childWrapper.className = 'a-checkbox__children-inline';
-    childWrapper.innerHTML = this.innerHTML;
+
+    // Clone live DOM Node so that we dont have to use mutation obersver which causes a
+    // deadlock because we dont have shadow dom here and it would infintily update - render - update
+    // We will set content of live dome AFTER render is done in order that dom nodes are aligned.
+    // Only needed if we have childs.
```
raphaellueckl:
```
// Clone live DOM node, so that we do not have to use mutation observer, which causes a
// deadlock, because we do not use shadow dom here and it would infinitely update - render - update.
// We will set the content of live DOM AFTER render is done, in order that DOM nodes are aligned.
// Only needed if we there are children.
```

^ I adapted some typos and and wordings. See here: https://www.diffchecker.com/bkdNZPE2

src/components/10-atoms/checkbox/react/DemoUpdateLabelChildren.jsx
```diff
@@ -0,0 +1,21 @@
+import React, { useState, useEffect } from 'react';
+import AXACheckbox from './AXACheckboxReact';
+
+export default function App() {
+  const [label, setLabel] = useState("AAAAAA");
```
raphaellueckl: What about "Meaningful Testdata"? :D 🗡

src/components/10-atoms/checkbox/ui.test.js
```diff
@@ -162,3 +162,17 @@ test('should show checkbox variant checkmark', async t => {
   const measuredColor = await getIconBackgroundColor();
   await t.expect(measuredColor).eql('rgb(0, 0, 143)');
 });
+
+fixture('Checkbox - DOM update works also on children').page(
+  `${host}/iframe.html?id=atoms-checkbox-react-demo--checkbox-updates-also-as-child`
+);
+
+test.only('should update checkbox when its children changes', async t => {
```
raphaellueckl: The .only will be the reason why the build fails.

Typo: changes => change

raphaellueckl (changes requested):
Nice, even if I don't know anything about live dom yet.

src/components/10-atoms/checkbox/react/DemoUpdateLabelChildren.jsx
LucaMele: Is the example our customer provided, but i can make it better on our example

src/components/10-atoms/checkbox/index.js
01231:
```
// Clone live DOM node, so that we do not have to use mutation observer. Because we do not use shadow DOM 
// here, it would infinitely update - render - update.
// Only needed if we there are children.
```

Shortened it a little bit...

raphaellueckl (dismissed):

raphaellueckl (approved):

PR #1621, opened 2020-01-16
Update footer small with children, etc

raphaellueckl:
Already released


src/components/20-molecules/footer-small/CHANGELOG.md
```diff
@@ -1,23 +1,25 @@
+### 4.0.3
+
+- Fixes dynamic updates of footer-small childelements. [#1549](https://github.com/axa-ch/patterns-library/issues/1549)
```
01231, line 3: I would write `child-elemets` or `child elements`

01231 (approved):

PR #1628, opened 2020-01-17
Update DOD so that all PRs can select all the checks

raphaellueckl:
Opinionated PR:
============

In my opinion, in a DOD, you should select EVERYTHING before the PR is ready to merge. This was not the case before, because not all bullet points could be checked in all the cases. But this is true now. When you are ready to have your PR reviewed, you can first make sure that you checked the whole list and signed every point.

Some stuff was redundant (like the changelog things), one point was useless (because the pipeline makes sure that nobody can merge something not compatible with eslint rules or that breaks tests) and some others needed a bit more text so that they can be marked as done without lying.

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -3,7 +3,7 @@ Fixes #(issue)
 # Done is when (DoD):
 - [ ] My changes ship with tests, if needed.
 - [ ] I added modified component properties to the typescript typings, if needed.
-- [ ] Design changes have been reviewed with a designer.
+- [ ] Design changes have been reviewed with a designer, if needed.
 - [ ] My changes are accordingly to our [Best Practices](https://github.com/axa-ch/patterns-library/blob/develop/CONTRIBUTION.md#best-practices)
 - [ ] My changes are well documented and showcased in the stories.
 - [ ] My changes are shortly added to CHANGELOG.md with the issue number.
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -1,15 +1,11 @@
 Fixes #(issue)
 
 # Done is when (DoD):
-- [ ] My changes ship with tests, if needed.
-- [ ] I added modified component properties to the typescript typings, if needed.
-- [ ] Design changes have been reviewed with a designer, if needed.
-- [ ] My changes are accordingly to our [Best Practices](https://github.com/axa-ch/patterns-library/blob/develop/CONTRIBUTION.md#best-practices)
-- [ ] My changes are well documented and showcased in the stories.
-- [ ] My changes are shortly added to CHANGELOG.md with the issue number.
-- [ ] Tested with IE, if affected.
-
-# Checklist:
-- [ ] My code follows the style guidelines of this project.
-- [ ] I have performed a self-review of my own code.
-- [ ] I have commented my code, particularly in hard-to-understand areas (focus on **WHY**), if comments are needed.
+- [ ] The modifications are reflected locked tested.
+- [ ] A self-review of the code has been performed.
+- [ ] The modified component properties have been added to the typescript typings.
+- [ ] Potential design changes have been reviewed by a designer.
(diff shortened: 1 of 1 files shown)


PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,17 +1,15 @@
...
+Fixes #(issue)
 
 # Done is when (DoD):
-- [ ] I have added UI/Unit tests for my changes
-- [ ] Old tests and eslint rule are not broken
-- [ ] I added modified component properties to the typescript typings
-- [ ] Design has been reviewed with (here name of reviewer)
+- [ ] My changes ship with tests, if needed.
+- [ ] I added modified component properties to the typescript typings, if needed.
+- [ ] Design changes have been reviewed with a designer.
```
01231: not always necessary ist it?

PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,17 +1,15 @@
...
 - [ ] My changes are accordingly to our [Best Practices](https://github.com/axa-ch/patterns-library/blob/develop/CONTRIBUTION.md#best-practices)
-- [ ] My changes are well documented in the README and showcased in the stories 
-- [ ] My changes are shortly added to CHANGELOG.md with the issue number
-- [ ] Tested with IE
+- [ ] My changes are well documented and showcased in the stories.
+- [ ] My changes are shortly added to CHANGELOG.md with the issue number.
+- [ ] Tested with IE, if affected.
```
01231: always needed in my opinon

raphaellueckl: Not if you make changes on the build level, or if you change readmes.

01231: True 👍

01231:
How about a section with criteria's which must be done for every single pr and a special section with criteria's which only apply to some PR's? Then we don't have to add a `if needed` to every sentence

danyball (dismissed):

PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,16 +1,11 @@
...
-- [ ] My changes are accordingly to our [Best Practices](https://github.com/axa-ch/patterns-library/blob/develop/CONTRIBUTION.md#best-practices)
-- [ ] My changes are well documented in the README and showcased in the stories 
-- [ ] My changes are shortly added to CHANGELOG.md with the issue number
-- [ ] Tested with IE
-
-# Checklist:
-- [ ] My code follows the style guidelines of this project
-- [ ] I have performed a self-review of my own code
-- [ ] I have commented my code, particularly in hard-to-understand areas (focus on **WHY**)
+- [ ] The modifications are reflected locked tested.
```
LucaMele: @raphaellueckl what does this mean? `reflected locked tested.`

PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,16 +1,11 @@
...
+- [ ] A self-review of the code has been performed.
+- [ ] The modified component properties have been added to the typescript typings.
+- [ ] Potential design changes have been reviewed by a designer.
+- [ ] The modifications are well documented and showcased in the stories.
```
LucaMele: @raphaellueckl here i would still refer to the `README.md` so that is clear also for new devs

LucaMele (commented):
I like @NoahLiechti Idea of splitting it, but also like this is fine for me.

I added 2 small comments and I personally would like to keep this:

`I have commented my code, particularly in hard-to-understand areas (focus on **WHY**)`

But no strong opinion

PULL_REQUEST_TEMPLATE.md
raphaellueckl: My bad... thx

danyball (dismissed):

PULL_REQUEST_TEMPLATE.md
raphaellueckl: Changed to `Modifications within components are reflected by tests.`

danyball (approved):

01231 (approved):

PR #1645, opened 2020-01-31
Refactor icon search and display only material icon names

### LINKED ISSUE #1640: Remove path characters from materials overview
Screenshot 2020-01-28 at 12 53 14

raphaellueckl:
Fixes #1640 

# Done is when (DoD):

### DIFF
--- a/src/components/00-materials/story.js
+++ b/src/components/00-materials/story.js
@@ -126,6 +126,11 @@ storiesOf('Materials', module)
           outline-offset: -2px;
           color: #000;
         }
+        .image-container {
+          display: flex;
+          align-items: center;
+          margin-bottom: 4px;
+        }
       </style>
       <h2>
         Note: The green borders reveal the dimensions of the SVGs.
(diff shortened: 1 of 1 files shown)


danyball (commented):
Isn't it useful to know that the files have the ending ".svg.js" ? If you want to import a icon you have to know that the filename is "xyz.svg.js"

raphaellueckl:
@danyball If you use the axa-icon component, this details does not matter. It will not work when you also provide the ending. Other than that, I would say, everyone uses autocompletion. and we would otherwise think about providing the whole path. That might be a useful addition. Instead of 2 columns, having 3 and one containing the path from `@axa-ch/materials/` on.

What I would recommend over what you say is, that we adapt the filename itself. Because `.svg.js` is kind of a weird file endling and we have eslint rules that say we should not import the whole file including extension. So currently, when you import and SVG from a js file, the import looks like this: `import checkmark from '@axa-ch/materials/icons/checkmark.svg'`. It looks like an SVG is being imported, while javascript just adds a `.js` at the end automatically. What you actually import with this expression, is a javascript file.

What do you think?

danyball:
@raphaellueckl first paragraph: jep. Providing the whole path to copy&paste would be great. like `@axa-ch/materials/icons/checkmark.svg`

second: Oh yes good idea. While creating the js file out of the svg, we should rename that file. Something like "xasdf-svg.js"

01231 (approved):
Nice

danyball (approved):

PR #1657, opened 2020-02-12
Handle undefined React properties

### LINKED ISSUE #1613: Datepicker: react datepicker throws error when onBlur udefined
v 5.0.1

when passing `undefined` to `onBlur`, error happens:

image

image

image

image

### LINKED ISSUE #1614: Handling undefined - probably a general react wrapper issue
I am not sure on this, but it seems to me, that the new kind of react wrapper which passes down all properties (aka seleniumid) has a general problem with handling `undefined` js type.
In my opinion `undefined` js type should be regarded as "not set / not present / absent". In reality we see (see bellow) several occurrences where components think that a `undefined` prop is present, and this causes problems.

I fear that if this issue isn't handled properly now, then we will face all kind of similar problems in the upcoming months.

Related issues:
- https://github.com/axa-ch/patterns-library/issues/1613
- https://github.com/axa-ch/patterns-library/issues/1602

markus-walther:
Fixes #1613.
Fixes #1614.

It works by preventing - in the React wrapper - overwriting the applied defaults of a component by React properties that are undefined.

A cleanup should ensure that all components have sensible defaults for all declared properties, and that no undeclared properties are used (incl. undeclared callback props).


LucaMele (dismissed):
LGTM

LucaMele (dismissed):
LGTM

LucaMele (approved):
LGTM

PR #1665, opened 2020-02-18
Fix failing checkbox test

### LINKED ISSUE #1664: Failing checkbox test
The test `should be clickable + change state and render parent label` can fail due to different screen geometry causing the embedded child-label link to be clicked.

This has actually happened for the Azure CI/CD pipeline.

Sharpen  and rework the test so that only the non-link portion of the child label is automatically clicked on, and fix any other errors in that test as well.

markus-walther:
Fixes #1664.


01231 (approved):
looks good and thanks for fixing

PR #1666, opened 2020-02-18
Rework axa-text to facilitate dynamic child updates

### LINKED ISSUE #1619: axa-text do not show text in storybook
Reproducer:

Use axa-text in story.js this way:

`<axa-text>${color.name}</axa-text>`

-->do not render any string.

`<axa-text>Hallo Welt</axa-text>` works correctly.

See `src/components/00-materials/styles/story.js` for example.

### LINKED ISSUE #1673: Tool to fix broken build
The new `npm run build` may accidentally break, leaving behind 'broken' package.json files in component subdirectories.

The sense in which they are broken is that these files contain a `main` key which was supposed to be deleted via restoring the original, main-free, package.json.

Write a tool that can repair such broken component package.jsons automatically.

markus-walther:
Fixes #1619.
Fixes #1673 en passant (it hindered development of axa-text).

This is a clean re-implementation that allows us to sidestep all dynamic update issues that plagued axa-text. It uses the same approach as axa-fieldset.

### DIFF
--- a/scripts/build/fix.js
+++ b/scripts/build/fix.js
@@ -1,0 +1,46 @@
+#!/usr/bin/env node
+
+/*
+*  Fix incorrect component package.json files left over by a broken build.
+*  Example invocation:
+*  patterns-library$ ./scripts/build/fix.js
+*/
+
+// imports
+const { cwd } = process;
+const fs = require("fs");
+const path = require('path');
+
+// synchronous directory-walker iterator
+function *walkSync(dir, filter = /.*/, acceptable = () => true) {
+    const files = fs.readdirSync(dir);
+
+    for (const file of files) {
+        const pathToFile = path.join(dir, file);
+        const isDirectory = fs.statSync(pathToFile).isDirectory();
+        if (isDirectory) {
+            yield *walkSync(pathToFile, filter, acceptable);
+        } else if (filter.test(pathToFile) && acceptable(dir, file, pathToFile)) {
+            yield pathToFile.split(path.sep);
+        }
+    }
+}
+
+// helpers
+const requirePackageJSON = dir => fs.existsSync(path.join(dir, 'package.json'));
+
+const currentDirectory = cwd();
+
+// find all importable components ...
+for (const file of walkSync(`src${path.sep}components`, /(?<!node_modules.*)package\.json$/, requirePackageJSON)) {
+    // construct full path to component package.json
+     const filePath = [currentDirectory, ...file].join(path.sep);
+     // load package.json's content
+     const packageJSON = require(filePath);
+     // remove 'main' key
+     delete packageJSON.main;
+     // write it back, properly formatted
+     fs.writeFileSync(filePath, JSON.stringify(packageJSON, /* no custom replacer */ null, /* number of spaces for formatting */ 2) + '\n');
+     // tell user
+     console.log('FIXED:', filePath);
+}
--- a/src/components/10-atoms/text/index.js
+++ b/src/components/10-atoms/text/index.js
@@ -1,4 +1,3 @@
-/* global Text */
 import { css, unsafeCSS } from 'lit-element';
 import NoShadowDOM from '../../../utils/no-shadow';
 
@@ -6,6 +5,8 @@ import NoShadowDOM from '../../../utils/no-shadow';
 import defineOnce from '../../../utils/define-once';
 import { applyDefaults } from '../../../utils/with-react';
 import styles from './index.scss';
+
+const TEXT_NODE_TYPE = 3;
 
 // N.B. This custom element is 'open' (no ShadowDOM) so that screen readers can discover
 // p(aragraph) tags inside
@@ -54,7 +55,7 @@ class AXAText extends NoShadowDOM {
   render() {
     // do we have 'pure', non-empty text, e.g. <axa-text>Hello, World</axa-text>?
     const nonWhitespaceTextNodes = [...this.childNodes].filter(
-      node => node instanceof Text && node.textContent.trim()
+      node => node.nodeType === TEXT_NODE_TYPE && node.textContent.trim()
     );
     if (nonWhitespaceTextNodes.length) {
       // yes, wrap it in 1 <p> node to keep screenreaders happy (they offer
--- a/src/components/10-atoms/text/index.js
+++ b/src/components/10-atoms/text/index.js
@@ -1,3 +1,4 @@
+/* global Text */
 import { css, unsafeCSS } from 'lit-element';
 import NoShadowDOM from '../../../utils/no-shadow';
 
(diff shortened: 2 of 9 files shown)


scripts/build/fix.js
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env node
```
LucaMele, line 1: @markus-walther why do you write a script for something that git revert does for you? In case of a broken build, better revert via git instead of writing a script that basically does the same

LucaMele (changes requested):

scripts/build/fix.js
LucaMele: Reason: keep repo as clean as possible.

src/components/10-atoms/text/index.js
```diff
@@ -19,47 +19,14 @@ class AXAText extends NoShadowDOM {
...
+      variant: { type: String, reflect: true },
     };
   }
 
   constructor() {
     super();
     applyDefaults(this);
   }
-
-  render() {
```
LucaMele, line 31: @markus-walther why are you removing this? as a reminder for axa text: AXA Text should always print semantically correct HTML, ergo text should alwys be wrapped via a `span` or a `p` tag. Removing this logic means that you will entirely rely on the user that he surrounds the text via a `span` or a `p` tag. The deleted logic takes care of that. Did this logic break anything or did u solve differently the requirement that `<axa-text>Text</axa-text>` and  `<axa-text><span>something</span></axa-text>` both create a TEXT based HTML tag once?

LucaMele (changes requested):

src/components/10-atoms/text/index.scss
```diff
@@ -1,39 +1,29 @@
...
     @include typo(primary, large);
   }
-}
 
-p.a-text {
   margin: 0;
 }
 
-.a-text--size-2 {
+axa-text[variant="size-2"], axa-text[variant="size-2"] > * {
```
LucaMele: I like this approach

scripts/build/fix.js
markus-walther: Because I don't want to do precision git reverts on package.jsons , taking care that other stuff is *not* reverted. And because my script fixes the underlying problem, so can be trusted - unlike git revert, which knows *nothing* about the problem, so might produce inadvertant other results.

LucaMele: @markus-walther i have no strong opinion on this one, seems just not necessary but if you prefer it so, then ok to me

src/components/10-atoms/text/index.js
markus-walther: axa-text is itself the wrapper. The deleted logic was brittle and lead to the bug ticket that this one fixes, elegantly.

scripts/build/fix.js
LucaMele: And also, that the build breaks is a very rare case so that precision work would be very rare

LucaMele: tests will fail anyway if you try to build the project with the `main` field there

src/components/10-atoms/text/index.js
LucaMele: @markus-walther requirement must be full filled. If a customer does not wrap its text in a p or span tag, we need to make sure is semantically correct. As idea, set role=“paragraph” IF direct child IS not a span or a p tag

markus-walther: Done

src/components/10-atoms/text/index.scss
markus-walther: Thanks!

src/components/10-atoms/text/index.js
```diff
@@ -28,38 +31,51 @@ class AXAText extends NoShadowDOM {
...
 
-    // nodeType === 3 is Text
-    if (firstChild && firstChild.nodeType === 3 && firstElementChild === null) {
-      const p = document.createElement('p');
-      p.innerHTML = firstChild.textContent;
-      p.className = classes.join(' ');
+  render() {
+    // do we have 'pure', non-empty text, e.g. <axa-text>Hello, World</axa-text>?
+    const nonWhitespaceTextNodes = [...this.childNodes].filter(
+      node => node instanceof Text && node.textContent.trim()
```
LucaMele: @markus-walther better use NodeType here instead of relying on Class Type. 

https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType

src/components/10-atoms/text/index.js
```diff
@@ -28,38 +31,51 @@ class AXAText extends NoShadowDOM {
...
+    );
+    if (nonWhitespaceTextNodes.length) {
+      // yes, wrap it in 1 <p> node to keep screenreaders happy (they offer
+      // paragraph-to-paragraph-jumping keyboard shortcuts that work by detecting <p>s):
+      const paragraph = document.createElement('p');
+      paragraph.textContent = this.textContent;
+      // pause child-update watcher
+      this.watch('stop');
+      // delete the children, now that their content has been harvested
       this.innerHTML = '';
```
LucaMele, line 68: @markus-walther this will disable live dom effect. Is it not better if we keep dom life?

LucaMele (changes requested):

src/components/10-atoms/text/index.js
LucaMele: Just be aware that u need to use the number, not the static property cause IE only knows the number type

markus-walther: You forgot to give a reason. I found https://stackoverflow.com/questions/13894644/having-trouble-with-dom-nodes-and-instanceof, but the case discussed there does *not* apply here.

Their case was testing a Text node across an iframe boundary.

Here we are either completely outside or completely inside an iframe, no crossing involved.

Therefore testing with `instanceof Text` is totally fine, which also explains why it works in StoryBook, which _does_ render the component in an iframe.

And: https://developer.mozilla.org/en-US/docs/Web/API/Text shows it is supported by IE, no special considerations necessary.

src/components/10-atoms/text/index.js
markus-walther: This is copied from your original code, so I'm a bit puzzled.

Now we have a MutationObserver, which takes care of any 'live' updating.

As discussed, the new summary assignment of `this.textContent` captures also the mixed case of some text nodes interspersed with some real DOM nodes (cf. https://developer.mozilla.org/de/docs/Web/API/Node/textContent).

src/components/10-atoms/text/index.js
LucaMele: As discussed, will be changed in favour of nodeType.

src/components/10-atoms/text/index.js
LucaMele: ok thanks

LucaMele (approved):
Nice job!!

PR #1683, opened 2020-03-10
CodePreview: show framework specific code

danyball:
We have codepreview right now:
image

What do you think about this solution?
https://axa-ch.github.io/plib-feature/1656_codepreview_3rdpartymodule/?path=/story/components-atoms-button--button (only preview for Button is integrated)
image


danyball:
Closed because:

Its not very handy to create code snippets for every framework-code. We need a solution that creates those code snippets automatically. And the story-code should be the template for that. So if we add a knob, the framework-code-snippets automatically updates.

I do not delete the branch on the server because that code could be interesting in future.

PR #1700, opened 2020-03-23
Rework axa-checkbox' child-label handling

### LINKED ISSUE #1699: Rework axa-checkbox code dealing with component children
The code dealing with component children is complex and hard to reason about.

Rework it to increase trust in that aspect of the component.

markus-walther:
Fixes #1699.


01231 (dismissed):
Very clean!
I would add the new attribute to the knobs and add a demo. (As far as I see you can't write HTML in the label knobs input-field)

PR #1702, opened 2020-03-24
remove browser default margins for axa-heading

### LINKED ISSUE #1693: Add option to control Heading's margin
The `Heading` component works great!
However, I am unable to control the top/bottom margin of the underlying (in my case) `h2` dom element.

In my case I need to reduce the margins to 0.
Please instruct how to proceed on consumer side, or implement this option.

Thanks!

image

raphaellueckl:
Fixes #1693

 OPTION 1! Read here about the in my opinion better option: #1703 

This is a very controversial PR. I would like to receive many opinions on it.

The need was to get resettable margins (and optionally customizable margins).

The way I went with that:
1. I use the CSS-Selector `:host` to make the `axa-heading` element displayed as a block element. `:host` is still in draft state, but my change is tested in Chrome, Safari and IE. Still, we once decided to not use a feature because it was in draft state.
2. We use predefined margins for the heading (as expected). But for the user to counter this margin, he needs to apply a negative margin to the root-element, which does not sound that nice, but makes the component extremely flexible.
3. I could have gone one step further and apply ALL margins directly on the root element, which would make negative margins (issues by the user) obsolete. But this would rely heavily on the host-property. This is my most liked approach and I will open a separate PR for this option, since we get the best compromise with this one.
4.  Not using the `:host` feature means a huge effort or the denial to our users to be able to manipulate the margins on their own.

Difference between Option 1 and 2:
```
Option 1:

<axa-heading rank="1">I am a heading</axa-heading>

<style>
  axa-heading[rank="1"] {
    margin-bottom: -15px;
    /* The margin-bottom is now 5px, because the default margin is 20px for H1, the user needs to know that and rely on it for the future. */
  }

Option 2:

<axa-heading rank="1">I am a heading</axa-heading>

<style>
  axa-heading[rank="1"] {
    margin-bottom: 5px;
    /* The margin-bottom is now 5px. */
  }

</style>
```

What do you think? In my opinion, we should allow the host property, just because it already works everywhere (it seems) and if, for whatever reason, support would get cancelled, we will probably note that ourselves or get failing tests. The IE support will never drop, for chrome we can write tests, only Safari and Firefox could go unseen, if nobody tracks this information.

# Done is when (DoD):
- [ ] Modifications within components are reflected by tests.
- [ ] A self-review of the code has been performed.
- [ ] The modified component properties have been added to the typescript typings.
- [ ] Potential design changes have been reviewed by a designer.
- [ ] The modifications are well documented (README.md, etc.) and showcased in the stories.
- [ ] The modifications are documented in the code, particularly in hard-to-understand areas (focus on **WHY**).
- [ ] The modifications are shortly added to CHANGELOG.md with the issue number.
- [ ] The modifications still work in IE.
- [ ] The modifications are according to our [Best Practices](https://github.com/axa-ch/patterns-library/blob/develop/CONTRIBUTION.md#best-practices)

### DIFF
--- a/src/components/10-atoms/heading/index.js
+++ b/src/components/10-atoms/heading/index.js
@@ -37,9 +37,14 @@ class AXAHeading extends LitElement {
       this.variant === 'secondary' ? 'a-heading--secondary' : '';
 
     const template = `
-      <h${this.rank} class="a-heading ${secondaryVariant}">
-        <slot></slot>
-      </h${this.rank}>
+    <style>
+      :host {
+        display: block;
+      }
+    </style>
+    <h${this.rank} class="a-heading ${secondaryVariant}">
+      <slot></slot>
+    </h${this.rank}>
     `;
 
     return html`
--- a/src/components/10-atoms/heading/story.js
+++ b/src/components/10-atoms/heading/story.js
@@ -19,17 +19,28 @@ storyAXAHeading.add('Heading', () => {
   const wrapper = document.createElement('div');
   const template = html`
     <axa-heading rank="1">H1 Primary Heading</axa-heading>
+    <br />
     <axa-heading rank="2">H2 Primary Heading</axa-heading>
+    <br />
     <axa-heading rank="3">H3 Primary Heading</axa-heading>
+    <br />
     <axa-heading rank="4">H4 Primary Heading</axa-heading>
+    <br />
     <axa-heading rank="5">H5 Primary Heading</axa-heading>
+    <br />
     <axa-heading rank="6">H6 Primary Heading</axa-heading>
+    <br />
 
     <axa-heading variant="secondary" rank="1">H1 Secondary Heading</axa-heading>
+    <br />
     <axa-heading variant="secondary" rank="2">H2 Secondary Heading</axa-heading>
+    <br />
     <axa-heading variant="secondary" rank="3">H3 Secondary Heading</axa-heading>
+    <br />
     <axa-heading variant="secondary" rank="4">H4 Secondary Heading</axa-heading>
+    <br />
     <axa-heading variant="secondary" rank="5">H5 Secondary Heading</axa-heading>
+    <br />
     <axa-heading variant="secondary" rank="6">H6 Secondary Heading</axa-heading>
   `;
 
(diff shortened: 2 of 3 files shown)


src/components/10-atoms/heading/index.js
```diff
@@ -37,9 +37,14 @@ class AXAHeading extends LitElement {
       this.variant === 'secondary' ? 'a-heading--secondary' : '';
 
     const template = `
-      <h${this.rank} class="a-heading ${secondaryVariant}">
-        <slot></slot>
-      </h${this.rank}>
+    <style>
```
markus-walther, line 40: Why use a `<style>` tag for this?? Put it into index.scss.

src/components/10-atoms/heading/index.js
```diff
@@ -37,9 +37,14 @@ class AXAHeading extends LitElement {
...
+      :host {
```
markus-walther, line 41: `:host` is OK. It's already used for `<axa-icon>`.

src/components/10-atoms/heading/index.js
```diff
@@ -37,9 +37,14 @@ class AXAHeading extends LitElement {
...
+        display: block;
```
markus-walther, line 42: Write a comment in the CSS why `display:block` is crucial, and the default for custom element doesn't work here.

markus-walther (changes requested):
What I reviewed is basically OK, just needing cleanup as indicated.

BUT: This is arguably incomplete, since the ticket asks for an _option_. In other words, the behaviour of margins is not conditioned on any attribute. I suggest either `margin` with numeric value (default: 0) or Boolean `noMargin`.

src/components/10-atoms/heading/index.js
raphaellueckl: Yes, I will do that, if we decide to go with this approach. :)

src/components/10-atoms/heading/story.js
```diff
@@ -1,38 +1,36 @@
...
+    ? html`
+        <axa-heading rank="${rank}" variant="secondary"
+          >H1 Primary Heading</axa-heading
+        >
+        <axa-text>${loremIpsum}</axa-text>
+      `
+    : html`
+        <axa-heading rank="${rank}">H1 Primary Heading</axa-heading>
+        <axa-text>${loremIpsum}</axa-text>
+      `;
```
01231, line 33: I would use this in a demo, because the I think the slim overview of the different sizes is very helpful.

PR #1703, opened 2020-03-25
Feature/remove heading margins v2

### LINKED ISSUE #1693: Add option to control Heading's margin

raphaellueckl:
Fixes #1693

OPTION 2! Read about Option 1: https://github.com/axa-ch/patterns-library/pull/1702

This is a very controversial PR. I would like to receive many opinions on it.

Tested on: Chrome, IE11, Brave, Firefox, Safari

The need was to get resettable margins (and optionally customizable margins).

The way I went with that:
1. I use the CSS-Selector `:host` to make the `axa-heading` element displayed as a block element. `:host` is still in draft state, but seems to work fine in all major browsers. Still, we once decided to not use a feature because it was in draft state (slotted).
2. The user does not need to apply a negative margin to the element, but he can overwrite the default margin by simply styling the root element.
3. Not using the `:host` feature means a huge effort or the denial to our users to be able to manipulate the margins on their own.

Difference between Option 1 and 2:
```
Option 1:

<axa-heading rank="1">I am a heading</axa-heading>

<style>
  axa-heading[rank="1"] {
    margin-bottom: -15px;
    /* The margin-bottom is now 5px, because the default margin is 20px for H1, the user needs to know that and rely on it for the future. */
  }

Option 2:

<axa-heading rank="1">I am a heading</axa-heading>

<style>
  axa-heading[rank="1"] {
    margin-bottom: 5px;
    /* The margin-bottom is now 5px. */
  }

</style>
```

What do you think? In my opinion, we should allow the host property, just because it already works everywhere (it seems) and if, for whatever reason, support would get cancelled, we will probably note that ourselves or get failing tests. The IE support will never drop, for chrome we can write tests, only Safari and Firefox could go unseen, if nobody tracks this information.

# Done is when (DoD):

### DIFF
--- a/src/components/10-atoms/heading/index.js
+++ b/src/components/10-atoms/heading/index.js
@@ -30,6 +30,15 @@ class AXAHeading extends LitElement {
     // this functions applies default values per type and verifies if
     // the HTML attribute has been set before defining the custom element
     applyDefaults(this);
+
+    this.topBottomMarginsByRank = {
+      1: '20px',
+      2: '18px',
+      3: '16px',
+      4: '14px',
+      5: '12px',
+      6: '10px',
+    };
   }
 
   render() {
@@ -40,6 +49,7 @@ class AXAHeading extends LitElement {
     <style>
       :host {
         display: block;
+        margin: ${this.topBottomMarginsByRank[this.rank]} 0;
       }
     </style>
     <h${this.rank} class="a-heading ${secondaryVariant}">
--- a/src/components/10-atoms/heading/index.js
+++ b/src/components/10-atoms/heading/index.js
@@ -37,9 +37,14 @@ class AXAHeading extends LitElement {
       this.variant === 'secondary' ? 'a-heading--secondary' : '';
 
     const template = `
-      <h${this.rank} class="a-heading ${secondaryVariant}">
-        <slot></slot>
-      </h${this.rank}>
+    <style>
+      :host {
+        display: block;
+      }
+    </style>
+    <h${this.rank} class="a-heading ${secondaryVariant}">
+      <slot></slot>
+    </h${this.rank}>
     `;
 
     return html`
--- a/src/components/10-atoms/heading/index.js
+++ b/src/components/10-atoms/heading/index.js
@@ -45,13 +45,10 @@ class AXAHeading extends LitElement {
     const secondaryVariant =
       this.variant === 'secondary' ? 'a-heading--secondary' : '';
 
+    this.style.marginTop = TOP_BOTTOM_MARGINS_BY_RANK[this.rank];
+    this.style.marginBottom = TOP_BOTTOM_MARGINS_BY_RANK[this.rank];
+
     const template = `
-    <style>
-      :host {
-        display: block;
-        margin: ${TOP_BOTTOM_MARGINS_BY_RANK[this.rank]} 0;
-      }
-    </style>
     <h${this.rank} class="a-heading ${secondaryVariant}">
       <slot></slot>
     </h${this.rank}>
--- a/src/components/10-atoms/heading/index.js
+++ b/src/components/10-atoms/heading/index.js
@@ -6,6 +6,15 @@ import defineOnce from '../../../utils/define-once';
 import { applyDefaults } from '../../../utils/with-react';
 import styles from './index.scss';
 
+const TOP_BOTTOM_MARGINS_BY_RANK = {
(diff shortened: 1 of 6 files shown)


src/components/10-atoms/heading/index.js
```diff
@@ -30,16 +30,31 @@ class AXAHeading extends LitElement {
     // this functions applies default values per type and verifies if
     // the HTML attribute has been set before defining the custom element
     applyDefaults(this);
+
+    this.topBottomMarginsByRank = {
```
markus-walther: ugh, this instance variable defines fixed constants that never change per instance! And it uses non-string keys in an object, which is illegal (and likely invisibly coerced to strings anyways).

`const TOP_BOTTOM_MARGINS_BY_RANK = {'1': '20px',.....}` outside of the class is much better.

src/components/10-atoms/heading/index.js
```diff
@@ -30,16 +30,31 @@ class AXAHeading extends LitElement {
...
       this.variant === 'secondary' ? 'a-heading--secondary' : '';
 
     const template = `
-      <h${this.rank} class="a-heading ${secondaryVariant}">
-        <slot></slot>
-      </h${this.rank}>
+    <style>
+      :host {
+        display: block;
+        margin: ${this.topBottomMarginsByRank[this.rank]} 0;
```
markus-walther: I think generated style tags with variable content are hacky, even though I get the idea and it's hidden under ShadowDOM.

Since this is per-instance variation, I recommend `this.style.marginTop = TOP_BOTTOM_MARGINS_BY_RANK[this.rank];` and similar for `marginTop`.

markus-walther: If my suggestion is followed, the remaining `:host { display: content; }` should be moved out to index.scss.

markus-walther (changes requested):
Unfortunately I again found a few things ;-)

Was super-confused by host-property in the announcement - no such thing was used in the generated CSS, only plain :host.

And like the related PR, I'd like to see this controlled by a new property, e.g. Boolean `nomargin`, which is also good to ensure backwards compatibility (if omitted).

src/components/10-atoms/heading/story.js
```diff
@@ -1,37 +1,36 @@
...
+    ? html`
+        <axa-heading rank="${rank}" variant="secondary"
+          >H1 Primary Heading</axa-heading
+        >
+        <axa-text>${loremIpsum}</axa-text>
+      `
+    : html`
+        <axa-heading rank="${rank}">H1 Primary Heading</axa-heading>
+        <axa-text>${loremIpsum}</axa-text>
+      `;
```
01231, line 33: I would use this in a demo, because the I think the slim overview of the different sizes is very helpful.

01231 (commented):
I like option 2 better, because it is just more convenient to use a normal margin than a negative one.

src/components/10-atoms/heading/index.js
raphaellueckl: Yeah I tried to make this a static property but stopped at some point (didn't work at first try). You're right.

src/components/10-atoms/heading/index.js
raphaellueckl: Genius! This is way better.

src/components/10-atoms/heading/story.js
danyball: The overview is helpful to see all the sizes. But i think its better to show an example with text below the header. So no overview for me - but no strong opinion.

src/components/10-atoms/heading/index.js
```diff
@@ -36,10 +45,13 @@ class AXAHeading extends LitElement {
     const secondaryVariant =
       this.variant === 'secondary' ? 'a-heading--secondary' : '';
 
+    this.style.marginTop = TOP_BOTTOM_MARGINS_BY_RANK[this.rank];
+    this.style.marginBottom = TOP_BOTTOM_MARGINS_BY_RANK[this.rank];
```
danyball, line 49: could be a one liner if you want.

danyball (approved):
Good solution! Very handy for the users.

PR #1717, opened 2020-04-06
axa-file-upload fixes

### LINKED ISSUE #1715: File-Upload max file size not considered for all file types
I have created large files of total `131.6MB` including images and pdfs.
The images show "File size exceeds maximum size" error, but not the pdfs.
I got large files from:
- https://www.hq.nasa.gov/alsj/a17/A17_FlightPlan.pdf
- https://effigis.com/en/solutions/satellite-images/satellite-image-samples/

https://patterns.axa.ch/?path=/story/components-molecules-file-upload-demos--feature-file-upload-in-a-form

ezgif com-optimize (1)

https://patterns.axa.ch/?path=/story/components-molecules-file-upload--file-upload

Screenshot 2020-04-06 at 10 22 51

markus-walther:
Fixes #1715.

To do that, applies some design-change suggestions from designers to allow coexistence of filename and per-file error message (approved as implemented).

Also fixes bugs w.r.t. updating of over-the-global-size-limit error message, e.g. when deleting files results in the remaining ones still being over the limit.

Finally, refactors and beautifies code a bit.

Please review carefully.


src/components/20-molecules/file-upload/index.js
```diff
@@ -208,42 +208,48 @@ class AXAFileUpload extends LitElement {
     this.requestUpdate();
   }
 
+  validateOverallSize(fileSize = 0) {
+    const {
+      sizeOfAllFilesInBytes,
+      maxSizeOfAllFilesKB,
```
01231, line 221: `maxSizeIfAllFilesInKB`? For consistency I would also ad the `In` here

01231 (changes requested):
In my opinion the labels should be centered, as they were before.
Screenshot 2020-04-07 at 12 23 19

01231 (changes requested):
I uploaded a 697kb big Pdf. The max size of a single file is 1kb. The Pdf shows no error but the images do.
Screenshot 2020-04-07 at 12 49 34

01231 (commented):
Maybe it would make sense to make sure when the maxSizeOfSingleFileKB is bigger than the maxSizeOfAllFilesKB, to make the maxSizeOfAllFilesKB as big as the maxSizeOfAllFilesKB. 
At the moment we have in this situation a global error message that shows that we have exceeded the max file size, but the file itself has no error.

src/components/20-molecules/file-upload/index.js
```diff
@@ -208,42 +208,48 @@ class AXAFileUpload extends LitElement {
...
-        this.sizeOfAllFilesByte + compressedImages[i].size >
-        maxSizeOfAllFilesByte
-      ) {
-        this.globalErrorMessage = this.filesTooBigStatusText;
-      } else if (compressedImages[i].size > maxSizeOfSingleFileByte) {
-        faultyFiles.push(compressedImages[i]);
+    const finalFiles = [];
+    const files = [...notImagesFiles].concat(compressedImages);
+
+    for (let i = 0, n = files.length, file, fileSize; i < n; i++) {
```
LucaMele: @markus-walther using the for loop var declaration for in scope not loop relevant var declaration is in my opinion not the cleanest way. Better declare on one line below the vars file and fileSize. Is more readable

src/components/20-molecules/file-upload/index.js
markus-walther: OK, makes sense

markus-walther:
> In my opinion the labels should be centered, as they were before.
> Screenshot 2020-04-07 at 12 23 19

Good catch!

01231 (approved):
Looks good. Tested it again for some cases🙌🏼

PR #1708, opened 2020-03-27
Feature/datepicker remove confirm cancel buttons

raphaellueckl:
This PR will remove the "Accept" and "Cancel" Buttons and also change the color of initially "preselected" day value (that look like they are selected but are actually not) to another color.

# Done is when (DoD):
- [x] Modifications within components are reflected by tests.
- [x] A self-review of the code has been performed.
- [x] The modified component properties have been added to the typescript typings.
- [x] Potential design changes have been reviewed by a designer.
- [x] The modifications are well documented (README.md, etc.) and showcased in the stories.
- [x] The modifications are documented in the code, particularly in hard-to-understand areas (focus on **WHY**).
- [x] The modifications are shortly added to CHANGELOG.md with the issue number.
- [x] The modifications still work in IE.
- [x] The modifications are according to our [Best Practices](https://github.com/axa-ch/patterns-library/blob/develop/CONTRIBUTION.md#best-practices)

PR #1718, opened 2020-04-07
Feature/datepicker remove confirm cancel buttons reworked

markus-walther:
This is an alternative version of PR #1708.

It tries to improve code clarity, but remain feature compatible.


src/components/20-molecules/datepicker/index.js
```diff
@@ -87,6 +88,35 @@ const parseAndFormatAllowedYears = (allowedyears = [], setYear) => {
...
+
+  if (typeof year === 'number' && year >= 0) {
+    date.setFullYear(year);
+  }
+
+  if (typeof month === 'number' && month >= 0) {
+    date.setMonth(month);
+    // month not changed as desired since target day unavailable?
+    // (e.g. July 31 =/=> June 31)
+    if (date.getMonth() !== month) {
```
raphaellueckl, line 103: I don't understand. This IF statement seems obsolete to me?

src/components/20-molecules/datepicker/index.js
```diff
@@ -278,18 +309,21 @@ class AXADatepicker extends NoShadowDOM {
     style.width = getFormattedStyle(width); // set width to component's css
     const formattedHeight = getFormattedStyle(height); // set height to input-wrap element because of optional label
 
-    const cellClasses = ({ sameMonth, today, selected, inactive }) =>
-      classMap({
+    const cellClasses = ({ sameMonth, today, selected, inactive, value }) => {
```
raphaellueckl: A comment would imho be helpful about what defines a selected cell. Somehow either if it is user selected, if it is preselected AND if it's 'today', right?

src/components/20-molecules/datepicker/index.js
```diff
@@ -444,14 +478,13 @@ class AXADatepicker extends NoShadowDOM {
...
     }
 
-    if (!this.isControlled) {
-      this.initDate(
-        null,
-        year || startDate.getFullYear(),
-        month || startDate.getMonth(),
-        day || startDate.getDate()
-      );
+    if (this.isControlled) {
```
raphaellueckl, line 480: Easier readable would imho be: 

```
if (!this.isControlled) {
  // Rest of the method...
}
```

But this is just personal preference.

src/components/20-molecules/datepicker/index.js
markus-walther: OK, will do. It's selected iff the same day as the internal this._date.

src/components/20-molecules/datepicker/index.js
markus-walther: This is what I had before, as the diff shows. The disadvantage is more nesting/indentation, which is usually harder to read. As you say, it's a preference, but those were my reasons.

src/components/20-molecules/datepicker/index.js
markus-walther: Even better action: remove reference to selected here, it's useless ;-)

PR #1721, opened 2020-04-08
feature/no stretch on mobile

### LINKED ISSUE #1690: Radio buttons mobile (due date 1.5.20)
It should be possible to display radio buttons on mobile next to each other as stated in the original comment by @TommyHoliday :
as discussed in slack already; the exact opposite behaviour should be possible for radio buttons. meaning, that on mobile there should be a possibility to choose flexible width. 
flexible-width
100-width

_Originally posted by @TommyHoliday in https://github.com/axa-ch/patterns-library/issues/1686#issuecomment-598689961_

01231:
Fixes #1690

# Done is when (DoD):

### DIFF
--- a/src/components/10-atoms/fieldset/README.md
+++ b/src/components/10-atoms/fieldset/README.md
@@ -74,9 +74,9 @@ The `horizontal` attribute is typically used to ensure correct layout for a grou
 
 `horizontal` mode is responsive: when crossing the `sm(all)` breakpoint (575px), this component forces conformant children to assume 100% width and rearranges them into a vertical layout with correct vertical spacing.
 
-### disableresponsivestretch
+### enableresponsivestretch
 
-Boolean `disableresponsivestretch` prevents the radio buttons to use the full length in a horizontal fieldset. Instead they are aligned side by side. It must be combined with `button`and `nogap` attributes from the axa-radio.
+Boolean `enableresponsivestretch` prevents the radio buttons to use the full length in a horizontal fieldset. Instead they are aligned side by side. It must be combined with `button`and `nogap` attributes from the axa-radio.
 
 ### error\*
 
--- a/src/components/10-atoms/fieldset/README.md
+++ b/src/components/10-atoms/fieldset/README.md
@@ -74,9 +74,9 @@ The `horizontal` attribute is typically used to ensure correct layout for a grou
 
 `horizontal` mode is responsive: when crossing the `sm(all)` breakpoint (575px), this component forces conformant children to assume 100% width and rearranges them into a vertical layout with correct vertical spacing.
 
-### nostretchonmobile
+### disableresponsivestretch
 
-Boolean `nostretchonmobile` prevents the radio buttons to use the full length in a horizontal fieldset. Instead they are aligned side by side. It must be combined with `button`and `nogap` attributes from the axa-radio.
+Boolean `disableresponsivestretch` prevents the radio buttons to use the full length in a horizontal fieldset. Instead they are aligned side by side. It must be combined with `button`and `nogap` attributes from the axa-radio.
 
 ### error\*
 
--- a/src/components/10-atoms/fieldset/README.md
+++ b/src/components/10-atoms/fieldset/README.md
@@ -68,15 +68,11 @@ Starred property names are reflected to attributes.
 
 ### horizontal\*
 
-Boolean `horizontal` changes the layout direction of the element's children to horizontal (default: false). It also imposes the correct horizontal gaps between children, except for children with a `nogap` attribute.
+String `horizontal` changes the layout direction of the element's children to horizontal (default: false). It also imposes the correct horizontal gaps between children, except for children with a `nogap` attribute.
 
 The `horizontal` attribute is typically used to ensure correct layout for a group of &lt;axa-radio&gt; buttons.
 
-`horizontal` mode is responsive: when crossing the `sm(all)` breakpoint (575px), this component forces conformant children to assume 100% width and rearranges them into a vertical layout with correct vertical spacing.
-
-### enableresponsivestretch
-
-Boolean `enableresponsivestretch` prevents the radio buttons to use the full length in a horizontal fieldset. Instead they are aligned side by side. It must be combined with `button`and `nogap` attributes from the axa-radio.
+`stretch` mode is responsive: when crossing the `sm(all)` breakpoint (575px), this component forces conformant children to assume 100% width and rearranges them into a vertical layout with correct vertical spacing.
 
 ### error\*
 
--- a/src/components/10-atoms/fieldset/README.md
+++ b/src/components/10-atoms/fieldset/README.md
@@ -74,6 +74,10 @@ The `horizontal` attribute is typically used to ensure correct layout for a grou
 
 `horizontal` mode is responsive: when crossing the `sm(all)` breakpoint (575px), this component forces conformant children to assume 100% width and rearranges them into a vertical layout with correct vertical spacing.
 
+### nostretchonmobile
+
+Boolean `nostretchonmobile` prevents the radio buttons to use the full length in a horizontal fieldset. Instead they are aligned side by side. It must be combined with `button`and `nogap` attributes from the axa-radio.
+
 ### error\*
 
 The string-valued `error` attribute, when nonempty, displays its value as an error message below the fieldset's children.
--- a/src/components/10-atoms/fieldset/demo.js
+++ b/src/components/10-atoms/fieldset/demo.js
@@ -19,12 +19,12 @@ storiesOf('Components|Atoms/Fieldset/Demos', module)
     const template = html`
       <axa-text variant="size-3">
         Resize your browser to a tablet or mobile width to see the effect of
-        "enableResponsiveStretch"</axa-text
+        'horizontal="stretch"'</axa-text
       >
       <br />
       <axa-text>How old are you?</axa-text>
       <br />
-      <axa-fieldset horizontal enableresponsivestretch>
+      <axa-fieldset horizontal="stretch">
         <axa-radio button name="contract" label="18-45" value="1"></axa-radio>
         <axa-radio button name="contract" label="46-54" value="2"></axa-radio>
         <axa-radio button name="contract" label="55+" value="3"></axa-radio>
--- a/src/components/10-atoms/fieldset/demo.js
+++ b/src/components/10-atoms/fieldset/demo.js
@@ -1,0 +1,53 @@
+/* global document */
+import { storiesOf } from '@storybook/html';
+import { html, render } from 'lit-html';
+import './index';
+
+import Readme from './README.md';
+import Changelog from './CHANGELOG.md';
+
+storiesOf('Components|Atoms/Fieldset/Demos', module)
+  .addParameters({
+    readme: {
+      sidebar: Readme,
+    },
+    changelog: Changelog,
+  })
+  .add('Fieldset - enableresponsivestretch', () => {
+    const root = document.createElement('div');
+
+    const template = html`
+      <axa-text variant="size-3">
+        Resize your browser to a tablet or mobile width to see the effect of
+        "enableResponsiveStretch"</axa-text
+      >
+      <br />
+      <axa-text>How old are you?</axa-text>
+      <br />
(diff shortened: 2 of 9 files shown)


src/components/10-atoms/fieldset/README.md
```diff
@@ -74,6 +74,10 @@ The `horizontal` attribute is typically used to ensure correct layout for a grou
 
 `horizontal` mode is responsive: when crossing the `sm(all)` breakpoint (575px), this component forces conformant children to assume 100% width and rearranges them into a vertical layout with correct vertical spacing.
 
+### nostretchonmobile
+
+Boolean `nostretchonmobile` prevents the radio buttons to use the full length in a horizontal fieldset. Instead they are aligned side by side. It must be combined with `button`and `nogap` attributes from the axa-radio.
```
danyball: "must be" combined? Why? This attribute makes also sense in different cases, isnt it? I mean, if you set this attr but no button or nogap, this also should work?

src/components/10-atoms/fieldset/index.js
```diff
@@ -19,6 +19,7 @@ class AXAFieldset extends NoShadowDOM {
       horizontal: { type: Boolean, reflect: true },
       error: { type: String, reflect: true },
       slot: { type: String },
+      noStretchOnMobile: { type: Boolean, reflect: true },
```
danyball: Just "onMobile"? Or even on Tablet? .... Another name would be better i think. something like "disableResponsiveStretch" or... ?? dont know. I would remove that "mobile" word

src/components/10-atoms/fieldset/index.scss
```diff
@@ -37,6 +37,14 @@ axa-fieldset {
     }
   }
 
+  &[nostretchonmobile]{
+    @include breakpoint($mediaquery-xsm) {
+      axa-radio[button][nogap] {
+        width: auto !important;
```
danyball: ` !important` should be your your last choice.

src/components/10-atoms/radio/story.js
```diff
@@ -40,7 +40,7 @@ storiesOf('Components|Atoms/Radio', module)
 
     const template = html`
       <p>Knobs with a * only affect the first radio button</p>
-      <axa-fieldset horizontal>
+      <axa-fieldset horizontal nostretchonmobile>
```
danyball: And if someone wants to see the "stretch" feature in action?! I would let these "foreign" story clean. You have this option in fieldset story, you dont need to add it here.

danyball (changes requested):

src/components/10-atoms/fieldset/README.md
01231: Yes you are right, my fault

src/components/10-atoms/fieldset/index.js
01231: Good idea👍🏼

src/components/10-atoms/radio/story.js
NoahLiechti: Yes that's that's true, but then we could also argue if `horizontal` is in the right place here.

src/components/10-atoms/fieldset/index.scss
01231: I think this is needed, because we have to override axa-radio from the fieldset component. But I will check that again.

01231: In my opinion `!important` is needed here. The specificity from the other selectors seems to be higher.

src/components/10-atoms/radio/story.js
01231: Removed it, because it now works for all kinds of radios. You can use the knobs from fieldset to test ist. No demo necessary.

src/components/10-atoms/fieldset/index.scss
raphaellueckl: I am also for alternatives, as long as they don't make the CSS ugly. If not possible, I guess we need it, but we should really try hard to not do that.

raphaellueckl (changes requested):
Nice, but there is no test validating the behavior.

src/components/10-atoms/radio/story.js
danyball: You are right. I would also remove horizontal. These are `axa-fieldset` features and should be shown at fieldset story or demo.

src/components/10-atoms/fieldset/story.js
```diff
@@ -17,10 +17,15 @@ storiesOf('Components|Atoms/Fieldset', module)
   .add('Fieldset', () => {
     const error = text('error', '');
     const horizontal = boolean('horizontal', false);
+    const disableresponsivestretch = boolean('disableResponsiveStretch', false);
```
danyball: This knob do not have any effect at this story. Try to show that knob in action (perhaps with some hints for the user like "resize your browser to a mobile width") or create a fieldset demo.

src/components/10-atoms/fieldset/index.scss
danyball: Try to rewrite the selector for this line: `width: 100% !important;`. That line should not be selected if "nostretch" is set.

src/components/10-atoms/fieldset/index.js
```diff
@@ -19,6 +19,7 @@ class AXAFieldset extends NoShadowDOM {
       horizontal: { type: Boolean, reflect: true },
       error: { type: String, reflect: true },
       slot: { type: String },
+      disableresponsivestretch: { type: Boolean, reflect: true },
```
danyball: The "auto stretch" feature is part of "horizontal" feature. So rename that new one like "horizontal-disableresponsivestretch"

danyball (changes requested):

src/components/10-atoms/fieldset/demo.js
```diff
@@ -0,0 +1,53 @@
...
+
+    const template = html`
+      <axa-text variant="size-3">
+        Resize your browser to a tablet or mobile width to see the effect of
+        "enableResponsiveStretch"</axa-text
+      >
+      <br />
+      <axa-text>How old are you?</axa-text>
+      <br />
+      <axa-fieldset horizontal enableresponsivestretch>
```
markus-walther: This line shows that `enableresponsivestretch` has redundant information and hence should be shortened. It's a Boolean attribute, so the presence automatically 'enables' things. Delete the 'enable' bit therefore. Naming matters...

src/components/10-atoms/fieldset/README.md
```diff
@@ -74,6 +74,10 @@ The `horizontal` attribute is typically used to ensure correct layout for a grou
 
 `horizontal` mode is responsive: when crossing the `sm(all)` breakpoint (575px), this component forces conformant children to assume 100% width and rearranges them into a vertical layout with correct vertical spacing.
 
+### enableresponsivestretch
+
+Boolean `enableresponsivestretch` prevents the radio buttons to use the full length in a horizontal fieldset. Instead they are aligned side by side. It must be combined with `button`and `nogap` attributes from the axa-radio.
```
markus-walther: This description sentence nicely illustrates that `enableresponsivestretch` is actually a submode of `horizontal`! This in turn means that `horizontal` should not be a Boolean attribute but a string-valued one. Inspection of the present CSS for axa-fieldset shows that no selector refers to the actual value of `horizontal`, i.e. such a change is entirely compatible with the existing CSS.

I suggest to rework this PR to capture the dependency of the new behaviour on `horizontal`!

markus-walther: A possible name for the value of `horizontal` is `responsivestretch` or just `stretch` (I prefer the latter one).

src/components/10-atoms/fieldset/demo.js
markus-walther: Forget the previous comment if you accept the change of `horizontal` data type to string from above...

markus-walther (changes requested):
Hi Noah,

nice work overall. I just suggest to recognize the fundamental dependency of your new behaviour on `horizontal` and rework the whole thing as a submode to capture the dependency formally.
This would also eliminate a class of errors where people forget to set `horizontal`....

--Markus

src/components/10-atoms/fieldset/index.react.d.ts
```diff
@@ -1,9 +1,10 @@
 import React from 'react';
 
 export interface AXAFieldsetProps {
-  horizontal?: boolean;
+  horizontal?: string;
   error?: string;
   slot?: string;
+  enableresponsivestretch?: boolean;
```
danyball: forgot to delete?

src/components/10-atoms/fieldset/story.js
```diff
@@ -16,24 +16,46 @@ storiesOf('Components|Atoms/Fieldset', module)
   })
   .add('Fieldset', () => {
     const error = text('error', '');
-    const horizontal = boolean('horizontal', false);
+    const options = {
+      None: null,
+      Stretch: 'stretch',
+      Horizontal: 'horizontal',
```
danyball: I would name these "variables" without upper letter. So they are same to the value and a dev can see valid values immediately.

src/components/10-atoms/radio/story.js
```diff
@@ -39,8 +39,11 @@ storiesOf('Components|Atoms/Radio', module)
     const noAutoWidth = boolean('noAutoWidth', false);
 
     const template = html`
-      <p>Knobs with a * only affect the first radio button</p>
-      <axa-fieldset horizontal>
+      <axa-text variant="size-3"
+        >Knobs with a * only affect the first radio button</axa-text
+      >
+      <br />
+      <axa-fieldset>
```
danyball, line 46: without horizontal. good!

src/components/10-atoms/fieldset/index.scss
```diff
@@ -26,14 +27,16 @@ axa-fieldset {
...
-        > * > * {
-          margin-right: 0 !important;
-          width: 100% !important;
-        }
+      > *,
+      > * > * {
+        margin-right: 0;
+        width: 100%;
+      }
+
```
danyball, line 39: nice refactoring!

src/components/10-atoms/fieldset/index.react.d.ts
```diff
@@ -1,9 +1,10 @@
 import React from 'react';
 
 export interface AXAFieldsetProps {
-  horizontal?: boolean;
+  horizontal?: string;
```
danyball: Instead of string you could write the 2 strings that are allowed in that way: `'string' | 'horizontal'`.

Or like the button d.ts as a extra TS type: `type Size = 'small' | 'large';`

danyball (changes requested):

src/components/10-atoms/fieldset/index.scss
01231: Thanks🙃

src/components/10-atoms/fieldset/story.js
01231: This variables represent how the text is shown in the dropdown. So I wanted them to be uppercase. Bu ya it doesn't really matter. Will change them to lowercase, so that they look better in code.

src/components/10-atoms/fieldset/index.react.d.ts
01231: jep :)

src/components/10-atoms/fieldset/index.react.d.ts
01231: Done🙌🏼

src/components/10-atoms/fieldset/index.js
01231: (outdated)

src/components/10-atoms/fieldset/story.js
01231: (outdated)

src/components/10-atoms/radio/story.js
01231: (outdated)

src/components/10-atoms/fieldset/demo.js
01231: Done

src/components/10-atoms/fieldset/README.md
01231: Done

danyball (approved):

PR #1723, opened 2020-04-09
remove datepicker specific stylings from axa dropdown

### LINKED ISSUE #1720: Datepicker unnecessary padding, cross-component styling
The dropdown has some styling information in it, that only belongs to datepicker components and at least some of that seems even cumbersome.

Screenshot 2020-04-07 at 18 02 36
The padding-right should be the same as the other elements, to allow for more text.

The padding-right breaks the dropdown, if the text gets cropped, so we should keep it as small as possible. This would be a change that needs the opinions from the designers. It could be, that because of the dropdown caret, they don't want the text to use more space. But I guess this rule doesn't make much sense and also, the dropdown has styling information about the datepicker in its css at the moment.

raphaellueckl:
Fixes #1720

This is discussed with the designers and ok for them.

@markus-walther I didn't understand, why there is a min-width on the dropdown. If there is a reason for that, which I guess there is, since I remember that there has been a discussion on the topic once, we could re-evaluate it. But your approval will be required before this PR gets live.

# Done is when (DoD):
- [ ] The modifications are shortly added to CHANGELOG.md with the issue number.
- [ ] The modifications still work in IE.

### DIFF
--- a/src/components/20-molecules/datepicker/index.js
+++ b/src/components/20-molecules/datepicker/index.js
@@ -384,7 +384,6 @@ class AXADatepicker extends NoShadowDOM {
                       maxheight="270"
                       items="${JSON.stringify(this.monthitems)}"
                       defaulttitle="${this.monthtitle}"
-                      usecase="datepicker"
                     >
                     </axa-dropdown>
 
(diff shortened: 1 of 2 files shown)


markus-walther (approved):
This looks good to me.

I trust you did a comprehensive manual check of the UI look-and-feel to ensure things are not visually broken for some special situations.

It did a cursory manual check, and things look good to me.

As for why the min-width, I think it was a defensive move to try and make a minimal-diff change that avoided checking all the use cases for dropdown just because we have a special need from within datepicker. So, a more general solution that avoids special cases is welcome, esp. with designer approval.

PR #1726, opened 2020-04-16
Feature/add changelogs

01231:


src/components/20-molecules/datepicker/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 7.0.1
+
+- Remove datepicker specific stylings from axa dropdown. (#1720)
```
raphaellueckl, line 3: IMHO we don't need a changelog for internal refactorings. We also did not write changelog for all the releases we did, for instance, when someone is doing a global release.

src/components/20-molecules/dropdown/CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## 7.0.1
```
raphaellueckl, line 1: IMHO we don't need a changelog for internal refactorings. We also did not write changelog for all the releases we did, for instance, when someone is doing a global release.

src/components/20-molecules/file-upload/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 2.0.9
+
+- Apply design-change suggestions from designers to allow coexistence of filename and per-file error message. (#1715)
```
raphaellueckl: This is the only message needed, nobody cares about the rest. This was a "patch" update, which means, some bugs were fixed and so on. I would leave out the other two points. Patches are for bugfixes and refactorings, there is no need to repeat that.

src/components/20-molecules/datepicker/CHANGELOG.md
NoahLiechti: In this case there were actually changes in the dropdown component

src/components/20-molecules/dropdown/CHANGELOG.md
01231: In this case there were actually changes in the dropdown component

src/components/20-molecules/file-upload/CHANGELOG.md
01231: But the user could be interested which bugs exactly are fixed with the version. Especially if he reported the bug.

src/components/20-molecules/datepicker/CHANGELOG.md
danyball: If you changed something at datepicker code i would say that should be part of the changelog, even if its internal. So we can also read the changelog to know what we have doing in past. But the changed need to be big enough to describe them.
It would be more interesting if you describe what have changed for the costumers. Is that fix visible somewhere? Or do you just use the new datepicker in that release?

src/components/20-molecules/file-upload/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
...
+- Minor bugfixes.
+- Code is refactored and beautified.
```
danyball: These 2 points are not necessary as raphi wrote. If you dont have a issue number for these bugfixes or nobody can see what you have fixed - remove that.

danyball (changes requested):

src/components/20-molecules/file-upload/CHANGELOG.md
01231: Done (issue is actually for the bug fixes)

danyball (dismissed):

PR #1736, opened 2020-04-21
[obsolete] Suppress duplicate style insertion

### LINKED ISSUE #1727: Multiple instances of a components cause multiple CSS appends to document head
IS: Multiple component make multiple entries in the head
SHOULD: for one component always and only one css should be rendererd in the `head` tag. Issue relys in NO Shadow. example: https://www.acc.axa.ch/de/rechner/rechtsschutzversicherung.html#/start

INFO: This has nothing to do with the possibility of having multiple versions of the component. 1 HTML = 1 JS = 1 CSS. This must be there as is a constraint of custom element itself.  Once we have done a multiversion stratgey, we will most likly end up in having CE Names like this: `axa-button-1-1-1`. Here again rule apply: 1 HTML = 1 JS = 1 CSS. Therefore the fix in no-shadow dom is also in the best intrest of Versioning and multi tenant.

markus-walther:
Fixes #1727.

Contains an en-passant fix for a subtle bug in InlineStyles that _also_ has to do with style attachment (but did not prove to be the root cause here, just something that showed up while investigating the issue).

The fix changes the build artifacts, i.e. all components need to be re-released to be safe.

I tested by modifying an example pod locally to contain axa-button, axa-fieldset and axa-radio, and copied relevant lib folder files from a patterns-library checkout containing the fix, then observing the duplication to go away.
 (The commit message 'Fix ...' contains the wrong issue number, sorry).

**Update**: Steps left for thorough review

- [ ] Is there a .css file in the export?
- [ ]  review sass-loader code for side effects other than head insertion
- [ ] sample testing under different scenarios: AEM, React, dist export, lib export (at least all
components in podaempatternslibrary)
- [ ] manual major-alpha test release
- [ ] check effects of different timing and CSS selector specificity due to different style application sequence after the  bugfix
-  [ ] test all-in-one component overview page under AEM


LucaMele (changes requested):
@markus-walther im just blocking merge cause the solution you found could work but i have to deeply test everything ad there is soooooo much involved and side effects

markus-walther:
> @markus-walther im just blocking merge cause the solution you found could work but i have to deeply test everything ad there is soooooo much involved and side effects

That's fine, I agree it needs thorough testing.

src/utils/inline-styles.js
```diff
@@ -24,11 +23,10 @@ export default class InlineStyles extends LitElement {
...
       crossTagReferenceCounts[this.tagName] || new WeakMap();
-    // one more reference, caused by this instance (undefined | 0 === 0)
-    const referenceCount = (referenceCounts.get(root) | 0) + 1;
-    // persist reference-count metadata
-    referenceCounts.set(root, referenceCount);
-    crossTagReferenceCounts[this.tagName] = referenceCounts;
+    // one more reference, caused by this instance
+    const info = referenceCounts.get(root) || {};
+    const referenceCount = (info.referenceCount || 0) + 1;
+    info.referenceCount = referenceCount;
```
danyball, line 29: At line 28 you add +1 to `info.referenceCount` and save it into a variable. And here you save this variable back to info.referenceCount? I would say an extra variable is not necessary here. It is better readable without it.

src/utils/inline-styles.js
```diff
@@ -57,15 +59,18 @@ export default class InlineStyles extends LitElement {
       return;
     }
     const root = this._rootNode || this.ownerDocument;
-    const referenceCount = (referenceCounts.get(root) | 0) - 1;
+    const info = referenceCounts.get(root) || {};
+    const referenceCount = (info.referenceCount || 0) - 1;
```
danyball, line 78: These two lines 62 + 63 could be transformed to a class method because of reuse at line 27. (and if you do that, dont forget to unit test that :) )

src/utils/inline-styles.js
```diff
@@ -57,15 +59,18 @@ export default class InlineStyles extends LitElement {
...
     // remove or update reference-count metadata
     if (referenceCount < 1) {
       referenceCounts.delete(root); // give an early hint to GC
-      if (styleTagNode && styleTagNode.parentNode) {
-        styleTagNode.parentNode.removeChild(styleTagNode); // remove style node as well
+      const { style } = info;
+      if (style && style.parentNode) {
+        style.parentNode.removeChild(style); // remove style node in DOM
```
danyball, line 84: This is still working, so this extra bugfix is ok. (Have testet my "old" bug with your new refactoring code.)

danyball (commented):
How to test that new solution? possible at storybook?

danyball:
Manual Test Result:

Works perfect for radio-buttons. Also if i remove parent elements from dom the style from head is removed correctly. But it is not working for `axa-text`? I have deleted the whole body element and the axa-text style was not deleted from header.

markus-walther:
> Manual Test Result:
> 
> Works perfect for radio-buttons. Also if i remove parent elements from dom the style from head is removed correctly. But it is not working for `axa-text`? I have deleted the whole body element and the axa-text style was not deleted from header.

Good catch! It's a bug in axa-text, which forgets to call super.disconnectedCallback().. Will fix in separate commit in this same branch.

src/utils/inline-styles.js
markus-walther: You forgot line 31, where thanks to the variable I can get rid of an unnecessary `this` reference. So, overall, I think my solution is preferrable.

src/utils/inline-styles.js
markus-walther: I decide not too, since it seems too small and the offsets are different (+1 vs -1).

PR #1754, opened 2020-04-28
Bugfix/axa radio dom removal bug

### LINKED ISSUE #1752: AXA Radio Exception with Disconnected Callback
Issue reported by Yannick Streit from Zühlke Engineering:

"We’re using the AXA radio button component in version 2.1.5 for one of our features and are facing a strange issue. We would be glad if you can help us since it seems to be an issue with the AXA Radio component.
 
When we navigate forth an back in our wizard our React component containing the radio buttons gets mounted/unmounted accordingly. When navigating back to the component containing the radio buttons the “disconnectedCallback” of the AXA radio button gets called unexpectedly. Since the component is not set up yet, a “null pointer” exception is thrown by this component (see screenshot below from @axa-ch/radio/lib/index.react.js:751).
 
This is an excerpt the code where we use the radio buttons in a fieldset:
 
```
<div className="checkout-actions">
  <AXAFieldset horizontal>
       <AxaRadioButton
              name="service-offer-response"
             data-test-id="add-product"
             label={t('price.acceptOffer')}
             value="true"
             button
             noGap
             checked={props.selected === true}
             onChange={() => onToggle(true)}
       />
       <AxaRadioButton
             name="service-offer-response"
             data-test-id="remove-product"
             label={t('price.declineOffer')}
             value="false"
             button
             noGap
             checked={props.selected === false}
             onChange={() => onToggle(false)}
       />
  </AXAFieldset>
</div>
```
 
Could you help us out?"
Screenshot

markus-walther:
Fixes #1752.

### DIFF
--- a/src/components/10-atoms/radio/index.js
+++ b/src/components/10-atoms/radio/index.js
@@ -184,7 +184,7 @@ class AXARadio extends NoShadowDOM {
         ? html`
             <label class="a-radio__wrapper">
               ${inputElement}
-              <div class="a-radio__content">${label}</div>
+              <div class="a-radio__content js-radio__content">${label}</div>
             </label>
           `
         : html`
@@ -198,7 +198,7 @@ class AXARadio extends NoShadowDOM {
   firstUpdated() {
     this.input = this.querySelector('input');
     const { name, button, noAutoWidth } = this;
-    const ourButton = this.querySelector('.a-radio__content');
+    const ourButton = this.querySelector('.js-radio__content');
     radioButtonGroup[name] = radioButtonGroup[name] || new Set();
     radioButtonGroup[name].add(ourButton);
 
@@ -206,7 +206,7 @@ class AXARadio extends NoShadowDOM {
       // give DOM some time to paint before measuring width
       window.requestAnimationFrame(() => {
         const { width: labelTextWidth } = this.querySelector(
-          '.a-radio__content'
+          '.js-radio__content'
         ).getBoundingClientRect();
         maxWidth[name] = Math.max(labelTextWidth | 0, maxWidth[name] | 0);
         // equalize width for all <axa-radio button> with same name:
@@ -257,7 +257,7 @@ class AXARadio extends NoShadowDOM {
     delete selectedRadioButton[name]; // help GC
     delete maxWidth[name]; // help GC
     // one button less in the set
-    const ourButton = this.querySelector('.a-radio__content');
+    const ourButton = this.querySelector('.js-radio__content');
     const successfullyDeleted = radioButtonSet.delete(ourButton);
     // it was the last in the set?
     if (successfullyDeleted && radioButtonSet.size === 0) {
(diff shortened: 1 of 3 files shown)


src/components/10-atoms/radio/index.js
```diff
@@ -244,11 +244,25 @@ class AXARadio extends NoShadowDOM {
...
+      return;
+    }
+    // clean up
     delete selectedRadioButton[name]; // help GC
     delete maxWidth[name]; // help GC
-    radioButtonGroup[name].delete(this.querySelector('.a-radio__content'));
-    if (radioButtonGroup[name].size === 0) {
-      delete radioButtonGroup[name]; // help GC
+    // one button less in the set
+    const ourButton = this.querySelector('.a-radio__content');
```
LucaMele: We always separate css and JS Dom selection. please use `js-radio__content` instad of `a-radio__content`.

src/components/10-atoms/radio/react/DemoRadioButtonReact.jsx
```diff
@@ -60,32 +63,45 @@ const DemoRadiobuttonsControlled = () => {
...
         &nbsp;error&nbsp;
       </label>
+      <label htmlFor="visible">
+        <input
+          id="visible"
+          type="checkbox"
+          checked={!!visible}
+          onChange={handleVisible}
+        />
+        &nbsp;visible&nbsp;
```
raphaellueckl, line 73: Why the two whitespaces?

src/components/10-atoms/radio/react/DemoRadioButtonReact.jsx
```diff
@@ -60,32 +63,45 @@ const DemoRadiobuttonsControlled = () => {
         />
         &nbsp;error&nbsp;
       </label>
+      <label htmlFor="visible">
+        <input
+          id="visible"
+          type="checkbox"
+          checked={!!visible}
```
raphaellueckl, line 70: Why do you assign `visible` to `checked` instead of `checked`?

01231 (dismissed):
LGTM (approved when that Luca's comment is implemented)

src/components/10-atoms/radio/react/DemoRadioButtonReact.jsx
```diff
@@ -6,6 +6,7 @@ const DemoRadiobuttonsControlled = () => {
   const [button, setButton] = useState(true);
```
raphaellueckl, line 6: I don't really understand what more this demo-file is actually demoing now. Comments would be helpful.

src/components/10-atoms/radio/index.js
markus-walther: OK, although this was there already in the old code.

src/components/10-atoms/radio/react/DemoRadioButtonReact.jsx
markus-walther: // rendering axa-radios invisible cleanly removes them from DOM without exceptions being thrown

is the comment describing the 'what more'. But that one is in ui.test.js, so you're right I should add something also in DemoRadioButtonReact.jsx.

markus-walther: Done

src/components/10-atoms/radio/index.js
markus-walther: Done

raphaellueckl (approved):
lgtm

PR #1758, opened 2020-04-30
fix axa link external attribute not opening in new tab

### LINKED ISSUE #1746: axa-link 'external' property doesn't seem to work anymore
Feedback from @MKaHead 

Reproduce: https://patterns.axa.ch/iframe.html?id=components-atoms-link--link&knob-link=https://www.axa.ch/en/information/data-protection.html&knob-Link%20text=Data%20protection%20statement&knob-external=true&knob-variant=&knob-icon=&knob-Background%20color=white

raphaellueckl:
Fixes #1746

There is no Test of the functionality, because of this comment: https://stackoverflow.com/a/56396445/2311557

I hope that puppeteer will help us here.

# Done is when (DoD):
- [ ] Modifications within components are reflected by tests.

### DIFF
--- a/src/components/10-atoms/link/index.js
+++ b/src/components/10-atoms/link/index.js
@@ -80,7 +80,7 @@ class AXALink extends LitElement {
           if (typeof this.onClick === 'function') {
             this.onClick();
           } else {
-            window.location.href = this.href;
+            window.open(this.href, this.external ? '_blank' : '_top');
           }
         }}"
         target="${this.external ? '_blank' : '_top'}"
--- a/src/components/10-atoms/link/index.js
+++ b/src/components/10-atoms/link/index.js
@@ -83,7 +83,6 @@ class AXALink extends LitElement {
             window.open(this.href, this.external ? '_blank' : '_top');
           }
         }}"
-        target="${this.external ? '_blank' : '_top'}"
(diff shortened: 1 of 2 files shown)


src/components/10-atoms/link/index.js
```diff
@@ -80,7 +80,7 @@ class AXALink extends LitElement {
           if (typeof this.onClick === 'function') {
             this.onClick();
           } else {
-            window.location.href = this.href;
+            window.open(this.href, this.external ? '_blank' : '_top');
           }
         }}"
         target="${this.external ? '_blank' : '_top'}"
```
danyball: This line seems to be obsolet now. I think every browser can handle window.open and target is not a "must" for a tag.

danyball (dismissed):

src/components/10-atoms/link/index.js
raphaellueckl: Right, but I left it there to make the intent of the link obvious. Maybe this helps a screenreader or something.

LucaMele: I would leave it in case we refactor the JS or in case the JS is not working, better to have it as HTML as well

markus-walther (approved):
Very nice, works also if after 1 x render JS is deactivated.

PR #1762, opened 2020-05-06
Change React-wrapper for Boolean props and fix checkbox as a result

### LINKED ISSUE #1761: Checkbox controlled-mode not working if initial checked property is false
Version: "@axa-ch/checkbox": "^4.0.1"

When initial checked property is false checkbox does not enter controlled-mode.

Reproduction link:
https://codesandbox.io/s/practical-leaf-lc0gs

### LINKED ISSUE #1772: IE: axa-radio "noGap" not working correctly
Reported by "Nick Rascher" over Slack. 

Open this page with IE11 (**React** demo. VanillaJS demo is working!):

https://patterns.axa.ch/?path=/story/components-atoms-radio-react-demos--controlled-component-react-ified-radio-button

There is a gap between buttons:
image

There's a margin at the first button. The CSS rule is not working?
image

markus-walther:
Fixes #1761.
fixes #1772 

This needs thorough thinking and reviewing, since it changes the behaviour for **all** Boolean properties under React.

### DIFF
--- a/src/utils/with-react.js
+++ b/src/utils/with-react.js
@@ -109,7 +109,7 @@ const distributeProperties = (properties, componentClass) => {
   let map;
   // iterate over all properties
   Object.keys(properties).forEach(name => {
-    let value = properties[name];
+    const value = properties[name];
     // classify property by type to select correct map object
     // (note that unregistered properties are classified as attr(ibute)s via their undefined .type)
     let type;
@@ -132,11 +132,8 @@ const distributeProperties = (properties, componentClass) => {
       case Array:
       case Object:
       case Function:
-        map = props;
-        break;
       case Boolean:
-        map = attrs;
-        value = value ? '' : null; // ''/null: canonicalize Boolean values s.t. val(...) sets or removes the attribute
+        map = props;
         break;
       default:
         map = declaredType ? props : attrs;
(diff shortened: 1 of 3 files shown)


src/utils/with-react.js
```diff
@@ -132,11 +132,8 @@ const distributeProperties = (properties, componentClass) => {
       case Array:
       case Object:
       case Function:
-        map = props;
-        break;
       case Boolean:
-        map = attrs;
-        value = value ? '' : null; // ''/null: canonicalize Boolean values s.t. val(...) sets or removes the attribute
+        map = props;
```
danyball, line 136: No need for a switch-case anymore. A simple if should be enough?!

danyball (commented):
Can you add a knob or a demo for that?

LucaMele (approved):
LGTM

markus-walther:
> Can you add a knob or a demo for that?

Difficult, since it affects the internal 'plumbing' only, its infrastructure code.

src/utils/with-react.js
markus-walther: As discussed bilaterally, this has its own downsides (repeated type === ...), so I prefer to keep it as-is.

PR #1763, opened 2020-05-06
Experimental PoC for component versioning

markus-walther:


src/components/20-molecules/datepicker/index.js
```diff
@@ -324,6 +336,39 @@ class AXADatepicker extends NoShadowDOM {
...
+      const newTag = closing => `${oldTag(closing) + '-' + aVersion}`;
+      return html(
+        strings.map(string =>
+          string.replace(oldTag(), newTag()).replace(oldTag('/'), newTag('/'))
+        ),
+        ...args
+      );
+    };
+
+    const monthDropdown = versionedHtml('axa-dropdown')`
```
LucaMele: versionedHtml\`axa-dropdown\`

src/utils/with-react.js
```diff
@@ -150,15 +151,16 @@ const distributeProperties = (properties, componentClass) => {
   return { attrs, props };
 };
 
-export default (createElement, componentClass) => {
+export default (createElement, componentClass, version) => {
   const { tagName } = componentClass;
-  const displayName = pascalCase(tagName);
+  const finalTagName = `${tagName}${version ? '-' + version : ''}`;
```
LucaMele: can be replaced with default params (the expression)

src/components/20-molecules/datepicker/index.js
```diff
@@ -324,6 +336,39 @@ class AXADatepicker extends NoShadowDOM {
       });
     };
 
+    const versionedHtml = tagName => (strings, ...args) => {
+      const aVersion = AXADatepicker.version[1][tagName];
+      const oldTag = (closing = '') => `<${closing}${tagName}`;
+      const newTag = closing => `${oldTag(closing) + '-' + aVersion}`;
+      return html(
```
LucaMele: Live DOM?

LucaMele:
IE CSS Collision when ShadoDom is mocked?

src/components/20-molecules/datepicker/index.js
```diff
@@ -324,6 +336,39 @@ class AXADatepicker extends NoShadowDOM {
...
+        strings.map(string =>
+          string.replace(oldTag(), newTag()).replace(oldTag('/'), newTag('/'))
```
LucaMele: var str = "Mr Blue has a blue house and a blue car";
var res = str.replace(/blue/g, "red");

OR 

SPLIT & JOIN

src/components/20-molecules/datepicker/index.js
```diff
@@ -324,6 +336,39 @@ class AXADatepicker extends NoShadowDOM {
...
+    };
+
+    const monthDropdown = versionedHtml('axa-dropdown')`
+        <axa-dropdown
+          @axa-change="${this.handleChangeDropdownMonth}"
+          class="m-datepicker__dropdown m-datepicker__dropdown-month js-datepicker__dropdown-month"
+          maxheight="270"
+          items="${JSON.stringify(this.monthitems)}"
+          defaulttitle="${this.monthtitle}"
+        >
```
LucaMele, line 341: `<axa-dropdown><axa-button><axa-icon></axa-icon></axa-button></axa-dropdown>`?

src/utils/with-react.js
```diff
@@ -167,5 +169,9 @@ export default (createElement, componentClass) => {
   // displayName is important for React testing (e.g. enzyme) and Chrome DevTools plugins
   reactStatelessComponent.displayName = displayName;
 
+  if (version) {
+    defineOnce(finalTagName, class extends componentClass {});
```
LucaMele: this needs to be done for native, meaing that native components also must define tag name like so:

`defineOnce('axa-dropdown', AXADropdown);`
`defineOnce('axa-dropdown-1-2-3', class extends AXADropdown {});`

axa-dropdown -> commonjs -> rsv -> axa-button
axa-dropdown -> pod-aem-plib -> axa-button

LucaMele: `defineOnce(${AXADropdown.tagName}-${AXADropdown.version[0]}, AXADropdown);`

LucaMele:
JEST / Selenium ?

LucaMele:
`axa-dropdown-rsv` ✅✅✅✅✅✅❤️❤️❤️❤️❤️

src/components/20-molecules/datepicker/index.js
markus-walther: ... Luca's comment is about multiple instances of the _same_ tagName in to-be-replaced string.

Additionally, a production version of a `versionedHTML` helper needs to generalize to multiple instances of _different_ tagNames, e.g. dropdown and button in same HTML template string.

src/components/20-molecules/datepicker/index.js
```diff
@@ -123,6 +123,13 @@ class AXADatepicker extends NoShadowDOM {
     return 'axa-datepicker';
   }
 
+  static get version() {
+    return [
+      '7.0.2',
+      { 'axa-dropdown': '7.0.2'.replace(/\./g, '-'), 'axa-button': '2.0.15' },
+    ];
+  }
+
```
danyball: Maybe 2 getter to have a array with same objects. Something like "get version" and "get versionDependencies".

src/components/20-molecules/datepicker/react/AXADatepickerReact.jsx
```diff
@@ -1,5 +1,5 @@
 import { createElement } from 'react';
 import createAXADatepickerReact from '../index.react';
 
-const AXADatepickerReact = createAXADatepickerReact(createElement);
+const AXADatepickerReact = createAXADatepickerReact(createElement, 'rsv');
```
danyball: Where to inject this information (rsv) if you use our WC without react? (edit: i think i get it - but we need to talk about it)

src/components/20-molecules/datepicker/index.js
```diff
@@ -324,6 +336,39 @@ class AXADatepicker extends NoShadowDOM {
       });
     };
 
+    const versionedHtml = tagName => (strings, ...args) => {
+      const aVersion = AXADatepicker.version[1][tagName];
+      const oldTag = (closing = '') => `<${closing}${tagName}`;
+      const newTag = closing => `${oldTag(closing) + '-' + aVersion}`;
```
danyball: What do you think to have this version-logic at a centered place? "TAGNAME-VERSION" is located here and in `with-react.js`.

markus-walther: yes, that will be factored out in a DRY fashion.

package.json
```diff
@@ -7,10 +7,12 @@
     "build": "NODE_ENV=production npm run build-components",
     "watch": "npm run watch-components",
     "build-components": "lerna run build --parallel",
+    "build-changed-components": "lerna run build --parallel --since",
     "watch-components": "lerna run watch --parallel",
     "setup-npm-links": "node scripts/setup-npm-link.js",
     "postinstall": "lerna bootstrap --hoist && cross-env NODE_ENV=production npm run build-components",
-    "release": "npm whoami && npx lerna clean && rm -rf node_modules && npm install && npm run test && lerna version && lerna publish from-package",
+    "prepublish": "cross-env NODE_ENV=production npm run build-changed-components",
```
LucaMele: @markus-walther this will trigger on install as well: https://docs.npmjs.com/misc/scripts#description. You might want `prepublishOnly `

scripts/build/version_info.js
```diff
@@ -0,0 +1,80 @@
+// GATHER VERSION INFO FROM COMPONENTS
+const path = require('path');
+const fs = require('fs');
+const {cwd} = process;
+
+// helpers
+const clean = name => name.replace(/^@axa\-ch\//, 'axa-');
```
LucaMele, line 7: @markus-walther maybe on Windows OS is `\` so please add that in the regex as well

LucaMele: pretty sure is not relevant just double check to be safe

LucaMele (commented):
@markus-walther sieht gut aus aufs erste, well done. Sag mir bescheid für finale review

package.json
markus-walther: Good point! Done

scripts/build/version_info.js
markus-walther: I checked, and https://docs.npmjs.com/misc/scope mandates a forward slash in package.json names, which is what `clean(...)` operates over.

So I conclude it's fine to leave it as it is.

LucaMele:
@markus-walther Did you test it in IE11? I think in IE11 due to the polyfill this code wont work (I could be that dependency is NOT instanceof HTMLElement):

```js
const componentClass =
      dependency instanceof HTMLElement ? dependency.constructor : dependency;
```

src/utils/component-versioning.js -> Zeile 53

src/utils/component-versioning.js
```diff
@@ -0,0 +1,99 @@
...
+  // process all dependant components that it lists...
+  let versionedTagName = '';
+  dependencies.forEach(dependency => {
+    const componentClass =
+      dependency instanceof HTMLElement ? dependency.constructor : dependency;
+    // ordinary, non-POD versioning?
+    if (!customVersion) {
+      // yes, inject version info into component-defining class
+      const { tagName } = componentClass;
+      componentClass.versions = versionInfo[tagName];
```
LucaMele: @markus-walther this is super dangerous. Please don't use the the class as a memory/state object, make a memory object away from it instead. What if a dev or new api starts defined `static versions` in the class? Performance wise is negligible the difference.

LucaMele (changes requested):

scripts/build/version_info.js
```diff
@@ -0,0 +1,80 @@
...
+  });
+  return result;
+}
+
+const ignoreNonComponents = (packages, ignores) => packages.filter( package => {
+    // ignore components classified under our Atomic-Design scheme as starting with 0
+    // (i.e. currently 00-materials, 05-utils)
+    const ignore = package.indexOf('src/components/0') > -1;
+    if (ignore) {
+        const [_, lastPathComponent] = package.match(/([a-z]+)$/);
```
LucaMele: @markus-walther i guess `obj[str.length]` would be more readable but is fine as well like this

LucaMele (changes requested):
scripts/build/version_info.js

Please fix this:

Screenshot 2020-05-26 at 09 20 43

scripts/build/version_info.js
```diff
@@ -0,0 +1,80 @@
...
+    const ignore = {};
+    const components = ignoreNonComponents(packages, ignore);
+    return {ignore, components, directoryPath};
+};
+
+// rollup version, is invoked in each component directory
+const gatherVersions = (aPath = cwd(), versionInfo = {}, ignore) => {
+    // determine to-be-ignored dependencies, if not passed-in already
+    if (ignore === undefined) {
+        ({ignore} = parseLernaJSON(path.resolve(aPath, '..', '..', '..', '..')));
```
LucaMele: @markus-walther super hack re-writing the ignore like this. Absolutely a no go from me because we tweak around what is intended in ES6 just to save 1 line of code. Please fix it with something like this `let toIgnore = ignore;` and then simply re-write `toIgnore`

scripts/build/version_info.js
```diff
@@ -0,0 +1,80 @@
...
+    }
+    // construct full path to component package.json
+    const filePath = `${aPath}${path.sep}package.json`;
+    // bail out if package.json missing
+    if (!fs.existsSync(filePath)) {
+        return '{}';
```
LucaMele: @markus-walther better like this: `return JSON.stringify({})`

LucaMele (changes requested):

scripts/build/version_info.js
LucaMele: I had this in our donts

src/components/20-molecules/top-content-bar/index.js
```diff
@@ -33,6 +36,9 @@ class AXATopContentBar extends LitElement {
   constructor() {
     super();
     applyDefaults(this);
+    /* eslint-disable no-undef */
+    defineVersioned([AXAButton, AXAButtonLink, AXAContainer], __VERSION_INFO__);
```
LucaMele, line 40: @markus-walther i really really like how you solved this. It also has the implicit benefit of hard dependency locking

LucaMele:
@markus-walther CODE REVIEW DONE. I will test now locally. Very very well done! Thank you so much!

LucaMele:
@markus-walther Please add a demo for versioning WITH version number in Storybook. We also need to beta release the button or datepicker and see if it works outside of storybook

markus-walther:
> @markus-walther Did you test it in IE11? I think in IE11 due to the polyfill this code wont work (I could be that dependency is NOT instanceof HTMLElement):
> 
> ```js
> const componentClass =
>       dependency instanceof HTMLElement ? dependency.constructor : dependency;
> ```
> 
> src/utils/component-versioning.js -> Zeile 53
image (2)

I.e., this apparently works just fine with IE11 under Windows 10 (tested with patterns.axa.ch).

src/utils/component-versioning.js
markus-walther: As discussed, I will check for the presence of a pre-existing `versions` class property and throw an error if found.

Rationale for automatic injection of `versions` here was twofold:

1. Why automatic? Remove an error case for developers who might forget to version a component otherwise
1. Why to component class? To aid live debugging using dev tools console, a la `window.customElements.get('axa-datepicker').versions`

scripts/build/version_info.js
markus-walther: I find mine a little bit more idiomatic because the first match is always the entire string covered by regexp, and `_` is the customary variable to say: I ignore this value.

So, since you leave it open here, I'd like to stick with the present code.

markus-walther:
> scripts/build/version_info.js
> 
> Please fix this:
> 
> Screenshot 2020-05-26 at 09 20 43

Ok, fixed by renaming to `aPackage`

src/components/20-molecules/top-content-bar/index.js
markus-walther: Thanks!

scripts/build/version_info.js
markus-walther: You are right. I followed your recommendation when rewriting that section of the code.

scripts/build/version_info.js
markus-walther: Done.

LucaMele (approved):
LGTM

PR #1764, opened 2020-05-07
Feature/radio support materials icon

### LINKED ISSUE #1738: Radio button/icon: DESIGN issue: support material icons
**Radio button with icon**

This demo bellow looks as it should...

https://patterns.axa.ch/?path=/story/components-atoms-radio-demos--radio-button-icon

```jsx
const icon = `<svg xmlns="http://www.w3.org/2000/svg" width="60.2" height="34.74">
<defs/>
<g transform="translate(.1 .1)">
..... transform="translate(-8.473 -23.037)"/>
</g>
</svg>`;

<axa-radio
          button
          name="insurance"
          label="Option 1"
          .icon="${icon}"
        ></axa-radio>

```
... as you can see the icon is inlined.

**Output:**

image

**With material icons used:**

image

I would like to keep the same design (icon size/margins) when using icon imported from materials:

```js
import car_svg from '@axa-ch/materials/images/car.svg.js';
```

Notice the difference in size/margins between the two screenshots above.

Discussed with Jonas on 21.4.2020.

Thanks!

01231:
Fixes #1738

# Done is when (DoD):


src/components/10-atoms/radio/index.scss
```diff
@@ -166,18 +166,19 @@ axa-radio {
...
+          width: 48px;
+          height: 48px;
           fill: currentColor;
         }
 
         .a-radio__content {
-          padding-bottom: 15px;
-          padding-top: inherit;
+          padding-bottom: 10px;
+          padding-top: 10px;
```
raphaellueckl, line 181: Not sure why it was inherit, maybe there was a reason. But if you let it as "inherit" it will still be 10px.

But just a comment, I have no opinion on this.

raphaellueckl:
Screenshot 2020-05-07 at 12 35 16

It looks better as before, but it still looks crappy. Does the design really request it to look like that?

src/components/10-atoms/radio/index.scss
danyball: Same like Raphi. If you exactly know why we have `inherit` here and it is not necessary, you can remove it. Otherwise not.

src/components/10-atoms/radio/story.js
```diff
@@ -7,16 +7,28 @@ import './index';
...
-    <path d="M32.889 25.928h-.039c-4.621.287-11.623 1.155-11.7 1.162a.653.653 0 0 0-.574.653v5.952a3.544 3.544 0 0 0 3.537 3.524H52.7a.653.653 0 0 0 .463-1.109c-.441-.445-10.772-10.75-20.274-10.182zm-11 7.76v-5.371c1.658-.2 6.6-.777 10.344-1.038v8.634h-8.116a2.232 2.232 0 0 1-2.232-2.226zm11.65 2.225v-8.706c6.97-.059 14.613 6.122 17.5 8.7z" transform="translate(-8.473 -23.037)"/>
-    </g>
-  </svg>`;
+const carIconSVG = `<svg width="48" height="48" viewBox="0 0 48 48" fill="currentColor" xmlns="http://www.w3.org/2000/svg">
```
danyball: Please use our material icons here. Do not copy paste the code of the SVGs to this story. Use something like `import car, boat, xyz from 'materials'`.

src/components/10-atoms/radio/story.js
```diff
@@ -39,17 +51,21 @@ storiesOf('Components|Atoms/Radio', module)
...
-      <axa-text variant="size-3"
-        >Knobs with a * only affect the first radio button</axa-text
-      >
+      <axa-text variant="size-3">
+        Knobs with a * only affect the first radio button
+      </axa-text>
+      <br />
+      <axa-text variant="size-2">
+        What type of vehicle do you want to insure?
+      </axa-text>
```
danyball, line 39: Good idea to change the example!

danyball (changes requested):

danyball:
> Screenshot 2020-05-07 at 12 35 16
> 
> It looks better as before, but it still looks crappy. Does the design really request it to look like that?

I think this combination of knobs should never be used. But ask a designer...

src/components/10-atoms/radio/story.js
01231: done

src/components/10-atoms/radio/index.scss
01231: Inherit just get its value from its parent element. `.a-radio__content` has a parent `.a-radio__wrapper`. The parents value for `padding-top` is also `10px`. In this case It's either way
10px. But because the padding of the wrapper and the content are independent, this is the only way to do it in my opinion.

01231: Commit where it was implemented. 
https://github.com/axa-ch/patterns-library/commit/606ad523db71b55817807fd01fa5a6e0562f32e7

src/components/10-atoms/radio/index.scss
```diff
@@ -166,18 +166,19 @@ axa-radio {
       height: auto;
```
danyball, line 166: I think the `padding: inherit` could be related to this line. So, "inherit" for the padding below could make sense.

src/components/10-atoms/radio/story.js
```diff
@@ -1,4 +1,6 @@
 /* global document */
+/* eslint-disable import/no-extraneous-dependencies */
+import { CarSvg, SailBoatSvg, PlaneSvg } from '@axa-ch/materials/images';
```
danyball, line 3: YES!:)

danyball (changes requested):

src/components/10-atoms/radio/story.js
01231: Due the fact, that the images have no width per default, they are not shown if only the icon attribute is set. Is this wanted?

danyball: Thats no problem if my statement here is correct: https://github.com/axa-ch/patterns-library/pull/1764#issuecomment-625176982

Or you can set width and height for svg for this case, too.

01231:
Update: The combination above is not in use. Approved by @TommyHoliday

src/components/10-atoms/radio/index.scss
01231: `padding: inherit` would never inherit something from the `height` property. `padding-top` from `a-radio__content` inherits from `padding` from the class `.a-radio__wrapper`. This creates a dependency that shouldn't be. We definitely have to remove `inherit`.

danyball: What i mean: if height is variable it could have a side-effect to the padding. But if i think about it... that possible wrong. So .. you are doing right:)

src/components/10-atoms/radio/index.js
```diff
@@ -176,7 +176,7 @@ class AXARadio extends NoShadowDOM {
         : html`
             <span class="a-radio__icon"></span>
           `}
-      ${icon ? svg([icon]) : html``}
+      ${icon && button ? svg([icon]) : html``}
```
danyball, line 179: nice

danyball (approved):

PR #1781, opened 2020-05-19
Fix nondeterministic axa-radio exceptions

### LINKED ISSUE #1780: axa-radio nondeterministically throws exceptions
From Yannick Streit:
"Hi guys, after updating to @axa-ch/radio@2.1.6 we're seeing an uncaught exception in the browser's console thrown sometimes after the component gets unmounted (this does not happen all the time). This is leading to flaky UI tests on our side since we assert that we have no errors in our console during test execution.
The problem might be related to the previous bug: https://github.com/axa-ch/patterns-library/issues/1752
Could you please have a look? I attached a screenshot of the place where the exception is thrown (radioButtonGroup[name] seems to be undefined)."
Screenshot 2020-05-12 at 16 53 29

markus-walther:
Fixes #1780.

Needs IE test since it relies on isConnected polyfill.

### DIFF
--- a/src/components/10-atoms/checkbox/CHANGELOG.md
+++ b/src/components/10-atoms/checkbox/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.0.3
+
+- Fixes checkbox controlled mode problem. #1761
+
 ## 4.0.1
 
 - Fixes an optical bug on mobile devices if trying to uncheck with a tap (#1709)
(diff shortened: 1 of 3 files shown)


01231 (commented):
Looks good to me. But is it already tested in IE or is this still TODO? Would be helpful if the PR checklist is visible.

markus-walther:
> Looks good to me. But is it already tested in IE or is this still TODO? Would be helpful if the PR checklist is visible.

I just tested it in IE, works.

PR #1783, opened 2020-05-20
make footer link less vulnerable to external hover stylings

raphaellueckl:
Any `a:hover` would overwrite the text-decoration of footer before. This PR changes that. That was a problem on axa.ch.

### DIFF
--- a/src/components/30-organisms/footer/child.scss
+++ b/src/components/30-organisms/footer/child.scss
@@ -55,6 +55,7 @@ axa-footer {
 
     &:hover {
       color: rgba($color-prim-white, 0.65);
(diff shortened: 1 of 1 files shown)


01231 (approved):
Looks good

danyball (approved):
Do we need something like that for other components as well? All the links at axa.ch are underlinded, not just these at the footer.

raphaellueckl:
@danyball I would say we should wait on issues, because there is no end to this loophole.

The best thing to do IMHO would be, to use shadowDom in all components, and also style the children with `::slotted`, but sadly that is still a draft. Then we could save a lot of rules.

PR #1809, opened 2020-06-12
Bugfix/1805 datepicker range

### LINKED ISSUE #1805: Datepicker: allowed years adds current year
When I set the allowedyears to ['1999-2000'] the date picker also adds the current year.
Version 7.0.8
Bildschirmfoto 2020-06-05 um 18 16 49
Bildschirmfoto 2020-06-05 um 18 16 06

### LINKED ISSUE #1810: Datepicker: auto select date if "currentYear" is not in range
Discussion on slack.

Datepicker is empty if range=1999 and currentyear=2020. Should autoselect a date.

"Ich setzte dann auf “aktuellerTag . aktuellerMonat . KleinesJahrInDerRange”"

danyball:
Fixes #1805 
fixes #1810 

# Done is when (DoD):

### DIFF
--- a/src/components/20-molecules/datepicker/README.md
+++ b/src/components/20-molecules/datepicker/README.md
@@ -54,7 +54,7 @@ _Note 2: The locale `it-CH` is mapped to date format `de-CH`. Example: `8.1.1982
 
 ### year
 
-The number-valued `year` defines the start year of the datepicker (default: current year).
+The number-valued `year` defines the start year of the datepicker (default: current year). If `year` do not match any entries at `allowedyears`, `year` will be the first entry of `allowedyears` to avoid an emtpy datepicker at startup.
 
 ### month
 
@@ -77,9 +77,9 @@ Input-field content is validated.
 
 ### allowedyears
 
-The array-valued `allowedyears` specifies the years that will be visible in the date-picker dropdown. The `year` property value is an automatic member of `allowedyears`.
+The array-valued `allowedyears` specifies the years that will be visible in the date-picker dropdown. The `year` property value is set as `allowedyears` if you do not specify any `allowedyears` to avoid an empty datepicker.
 
-Both individual years `[2019,202]`,
+Both individual years `[2019,2020]`,
 year ranges `["1989-2010"]`, and a mix of the two `["1989-2010", 2012, 2014]` are supported.
 
 ### placeholder, monthtitle, yeartitle, invaliddatetext
--- a/src/components/20-molecules/datepicker/README.md
+++ b/src/components/20-molecules/datepicker/README.md
@@ -54,7 +54,7 @@ _Note 2: The locale `it-CH` is mapped to date format `de-CH`. Example: `8.1.1982
 
 ### year
 
-The number-valued `year` defines the start year of the datepicker (default: current year). If `year` do not match any entries at `allowedyears`, the first entry of `allowedyears` will be set as `years` to avoid an emtpy datepicker at startup.
+The number-valued `year` defines the start year of the datepicker (default: current year). If `year` do not match any entries at `allowedyears`, the first entry of `allowedyears` will be set as `years` to avoid that datepicker don't show any days to select and looks like "emtpy".
 
 ### month
 
--- a/src/components/20-molecules/datepicker/README.md
+++ b/src/components/20-molecules/datepicker/README.md
@@ -54,7 +54,7 @@ _Note 2: The locale `it-CH` is mapped to date format `de-CH`. Example: `8.1.1982
 
 ### year
 
-The number-valued `year` defines the start year of the datepicker (default: current year). If `year` do not match any entries at `allowedyears`, `year` will be the first entry of `allowedyears` to avoid an emtpy datepicker at startup.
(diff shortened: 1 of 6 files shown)


src/components/20-molecules/datepicker/index.js
```diff
@@ -481,6 +488,16 @@ class AXADatepicker extends NoShadowDOM {
...
+    if (
+      allowedyears &&
+      !allowedyears.includes(startDate.getFullYear()) &&
+      !allowedyears.includes(year)
+    ) {
+      // to avoid an empty datepicker at startup set a valid year
+      const [newStartYear] = allowedyears;
+      this.year = newStartYear;
+    }
+
```
danyball: @markus-walther I think this part of code is not at the correct place. Why? Because if i change the html attribute "year" after everything is renderd nothing is happend.

danyball:
@markus-walther This test is failing: `test('should allow year ranges'` - and i think its because of the part of code at the wrong place.

src/components/20-molecules/datepicker/README.md
```diff
@@ -54,7 +54,7 @@ _Note 2: The locale `it-CH` is mapped to date format `de-CH`. Example: `8.1.1982
 
 ### year
 
-The number-valued `year` defines the start year of the datepicker (default: current year).
+The number-valued `year` defines the start year of the datepicker (default: current year). If `year` do not match any entries at `allowedyears`, the first entry of `allowedyears` will be set as `years` to avoid an emtpy datepicker at startup.
```
markus-walther: wording: `year` attribute

markus-walther: wording: does not match...
wording: will be set as `year`
spelling: empty

And: what does 'empty' datepicker mean? That's not a well-defined notion that everybody can understand. Can you express your concept better?

src/components/20-molecules/datepicker/README.md
```diff
@@ -77,9 +77,9 @@ Input-field content is validated.
 
 ### allowedyears
 
-The array-valued `allowedyears` specifies the years that will be visible in the date-picker dropdown. The `year` property value is an automatic member of `allowedyears`.
+The array-valued `allowedyears` specifies the years that will be visible in the date-picker dropdown. The `year` property value is set as `allowedyears` if you do not specify any `allowedyears` to avoid an empty datepicker.
```
markus-walther: Major confusion here with value types.  Year is a number, so cannot be set as allowedyears, which is an array. Suggestion: `allowedyears` defaults to a length-1 array containing the `years` value.

For 'empty' datepicker, see above.

src/components/20-molecules/datepicker/index.js
```diff
@@ -54,9 +54,15 @@ const applyEffect = self =>
     }, 0 /* execute after render() */);
   });
 
-const parseAndFormatAllowedYears = (allowedyears = [], setYear) => {
+export const parseAndFormatAllowedYears = (allowedyears = [], setYear) => {
```
markus-walther, line 60: I was confused why this is exported until I saw the unit test in a separate file. I suggest a comment
/* needed in unit tests */.

src/components/20-molecules/datepicker/react/story.jsx
```diff
@@ -5,7 +5,7 @@ import {
   boolean,
   number,
   select,
-  withKnobs,
+  withKnobs, object,
```
markus-walther: all other datatypes are on their own lines, maybe put object before withKnobs on its own line too?

src/components/20-molecules/datepicker/index.js
markus-walther: I'm guessing you mean parseAndFormatAllowedYears and 'is not INVOKED at ALL the correct places'.

If so, I suggest to add `|| changedProperties.has('year')` at the end of the first `if` condition in `shouldUpdate`.

markus-walther: OK, after getting clarification from @danyball : there is only one place currently where the property `year` is written, and that is in the `initDate` method.

So, I suggest to move this part of the code - i.e. the big if block - to that method. I also suggest trying to make the complicated if condition simpler, if possible.

markus-walther (changes requested):

src/components/20-molecules/datepicker/README.md
```diff
@@ -77,9 +77,9 @@ Input-field content is validated.
 
 ### allowedyears
 
-The array-valued `allowedyears` specifies the years that will be visible in the date-picker dropdown. The `year` property value is an automatic member of `allowedyears`.
+The array-valued `allowedyears` specifies the years that will be visible in the date-picker dropdown. If you do not specify any `allowedyears`, the `year` property value will be added to `allowedyears` to avoid an empty year dropdown.
```
01231, line 80: If there is no `year` and `allowedyears` then the datepicker is broken. Is this something to cover?

01231 (commented):
LGTM

src/components/20-molecules/datepicker/README.md
danyball: Did you test that? Year should never be empty because it has a default value (current year).

markus-walther: I just tested it with creating an axa-datepicker using `document.createElement`, and attaching it to `document.body`, i.e. as bare as it gets. It works, showing that the defaults work correctly in the absence of `year` and `allowedyears`.

markus-walther (approved):

PR #1844, opened 2020-07-02
Feature/component flat list

### LINKED ISSUE #1777: [Child of #1773] Reorder our components to a flat list
We do not want to group our components any more. Lets display them as a flatlist. But do not change the filesystem of PL.

And remove the component names out of the `demos` that these demos are not found by a storybook search.

01231:
Fixes #1777

# Done is when (DoD):


raphaellueckl (dismissed):
LGTM, but I don't like that all the "main stories" are now called "Story" instead of by their specific component name.

.storybook/config.js
```diff
@@ -19,6 +19,8 @@ addParameters({
       brandUrl: 'https://github.com/axa-ch/patterns-library/tree/develop',
       brandImage: logo,
     }),
+    storySort: (a, b) =>
+      a[1].kind === b[1].kind ? 0 : a[1].id.localeCompare(b[1].id, undefined, { numeric: true }),
```
danyball: nice!

danyball (changes requested):
- can you configure Storybook to always try to open "Story" if the user clicks on a folder? Maybe the word "story" is not understood by our users. Maybe something like "Example" or "Showcase" would be better?!

- is it possible to set the same icon on the components? It looks a little bumpy.
image

01231:
> * can you configure Storybook to always try to open "Story" if the user clicks on a folder? Maybe the word "story" is not understood by our users. Maybe something like "Example" or "Showcase" would be better?!
> * is it possible to set the same icon on the components? It looks a little bumpy.
>   image

- The folder icon appears when there are subfolders in the component and the blue one when there is no subfolder. The word examples or showcase are synonyms for demo in my opinion. Our stories are demos itself. The name story gives it a reason to be in this superordinate position from the demos.
- Storybook seems to only open a story automatically when there is no subfolder (blue icon). In the passt they all where blue and therefore opened the story on click.

danyball:
> > * can you configure Storybook to always try to open "Story" if the user clicks on a folder? Maybe the word "story" is not understood by our users. Maybe something like "Example" or "Showcase" would be better?!
> > * is it possible to set the same icon on the components? It looks a little bumpy.
> >   image
> 
> * The folder icon appears when there are subfolders in the component and the blue one when there is no subfolder. The word examples or showcase are synonyms for demo in my opinion. Our stories are demos itself. The name story gives it a reason to be in this superordinate position from the demos.
> * Storybook seems to only open a story automatically when there is no subfolder (blue icon). In the passt they all where blue and therefore opened the story on click.

I dont like `story`. Our `demos` showing just special behaviour and are not "showcases" that shows all the cases of a component.

01231:
> * this
Screenshot 2020-07-08 at 09 39 08

It's called storybook. If we can expect from the user to now what a storybook is we can expect from him that he will understand what a story is. And if he doesn't now what a story is, he'll probably know it as soon as he clicks on the component, because this is the first thing that opens. And due the fact that story will be used throughout the whole storybook this will not be much of a challenge for him.

If you really don't like the name story I'll need a better suggestion, because the old naming was not good and redundant. What i don't like about "examples" or "showcase" is, that this would also work for all the demos in my opinion. I know we could also say that about the name "story". What about "Nativ component" or "native story"? This would descriptive and unambiguously.

LucaMele:
I personally prefer story for the reaosons Noah explained

danyball (approved):

PR #1852, opened 2020-07-06
move documentation into docs folder

### LINKED ISSUE #1851: Reorder documentation files
We have so many MD files, that we should restructure the project. Add a docs folder for those and update the links.

Screenshot 2020-07-03 at 15 02 47

raphaellueckl:
Fixes #1851

BEFORE:

image

AFTER:

image

I updated all links and explicitly searched the whole repo (after the manual work) for every changed path.


LucaMele (changes requested):
Docs is not made for readmes, but for static HTML. This PR also will conflict with the big one of flexbox.

LucaMele:
@raphaellueckl i think is much better in the root because there is where you expect them. See other community repos: 

https://github.com/facebook/react
https://github.com/angular/angular
https://github.com/vuejs/vue
etc

And moving them will kill all already in placed links to all documents like in midgard etc. For me, this PR is to close

PR #1855, opened 2020-07-06
fix rendering empty accordion, if no footer links are provided

### LINKED ISSUE #1836: axa-footer: renders empty accordion on mobile
Hi,

MyAXA portal uses` "@axa-ch/footer": "3.0.8"` and I noticed bug in mobile design. 

We are using only one "column" section, not two as yours demo.
Empty accordion is rendered on mobile resolution.

I tried to reproduce it with the latest version, but it is not possible because of another issue #1832 .

Repo to reproduce (branch `myaxa-version`):
https://github.com/branislav-remen/patterns-lib-footer/tree/myaxa-version

image
image

raphaellueckl:
Fixes #1836

This PR should probably not be merged, before the component versioning is fixed.

### DIFF
--- a/src/components/30-organisms/footer/CHANGELOG.md
+++ b/src/components/30-organisms/footer/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.1.2
+
+- Do not render an empty accordion on mobile if no content is given. #1836
+
 ## 4.1.0
 
 - Replaced old typography with new one. This changes could have changed the components design. (#1796 and #1750)
--- a/src/components/30-organisms/footer/CHANGELOG.md
+++ b/src/components/30-organisms/footer/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 4.1.2
 
-- Do not render an empty accordion on mobile if no content is given. #1836
+- Only render accordion on mobile when content is given. #1836
 
 ## 4.1.0
 
--- a/src/components/30-organisms/footer/demo.js
+++ b/src/components/30-organisms/footer/demo.js
@@ -63,4 +63,9 @@ storiesOf('Components|Footer/Demos', module)
     });
 
     return wrapper;
+  })
+  .add('Without content', () => {
+    const wrapper = document.createElement('div');
+    wrapper.innerHTML = `<axa-text>This is an emtpy footer. Just for testing.</axa-text><axa-footer></axa-footer>`;
+    return wrapper;
   });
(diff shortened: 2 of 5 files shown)


danyball:
Azure failing?!

src/components/30-organisms/footer/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 4.1.2
+
+- Do not render an empty accordion on mobile if no content is given. #1836
```
01231: Would write: Only render accordion on mobile when content is given

src/components/30-organisms/footer/demo.js
```diff
@@ -62,5 +62,10 @@ storiesOf('Components|Footer/Demos', module)
       span.innerText = `Last clicked link: ${ev.detail}`;
     });
 
+    return wrapper;
+  })
+  .add('Without content', () => {
+    const wrapper = document.createElement('div');
+    wrapper.innerHTML = `<axa-text>This is an emtpy footer. Just for testing.</axa-text><axa-footer></axa-footer>`;
```
01231: not very eloquent in my opinion. I would add a footer like this but with the header `AXA WELTWEIT` in addition (only the header):
Screenshot 2020-07-23 at 09 46 14

And then write something like: No carousel should be rendered on mobile for `AXA WELTWEIT`

01231 (changes requested):

src/components/30-organisms/footer/demo.js
danyball: This is not a common use case. Who wants to have a column with a header but without content? See the issue for that PR. The user do not use column 2. No header, no items.

src/components/30-organisms/footer/CHANGELOG.md
danyball: Wont fix. Minor.

01231: Why not? `ctrl C` + `ctrl V`
Double negative is difficult to read.

src/components/30-organisms/footer/demo.js
01231: Then don't add the header. I just don't like that there is just a blue square in the demo. You can't tell what the example is about even with a description. It looks broken. In my opinion there should be either the `AXA & SIE` column or the `IN KONTAKT BLEIBEN` column.

danyball: Ah sorry. Misunderstand that. I want to test that no emtpy accordion is rendered even for column 0. So both columns needs to be emtpy. But i could add the "in kontakt bleiben" column if that is better?

01231 (approved):
Nice. Looks good!

PR #1867, opened 2020-07-09
Feature/update storybook 5 3

### LINKED ISSUE #1774: [Child of #1773] Update Storybook to version 6 [Due Date: 03.08.20]
We dont want to redesign storybook on a old version to avoid doubled work. The updated causes some breaking changes and we want to fix them first.

raphaellueckl:
Fixes #1774

My changes are according to the most recent documentation. We use the very newest versions here. Feel free to adapt, but MAKE SURE that the following stuff works on IE11, otherwise we cannot run it on IE11.

It was a real struggle to get to this point. If you see minor issues, I HIGHLY advise you to merge it first and change the smaller stuff afterwards. If you think there are major issues, please take over the PR and add your ideas.

This project was created with the Storybook-Cli. :)

- `npm run test`
- `npm run build-storybook && cd storybook-static && npx http-server` -> Open IE11 on `http://localhost:8080/`
- `npm start` -> Open IE11 on `http://localhost:6006/`
- Verify that the addons work at all times (Keyboard Shortcut: `D`)
- Make sure that assets like the axa logo and the favicons are displayed in all browsers.


.storybook/preview.js
```diff
@@ -1,4 +1,4 @@
-import '@axa-ch/patterns-library-polyfill';
+import './polyfills-ie11.js';
```
LucaMele: @raphaellueckl why this?

LucaMele: i mean, why not keep our ones?

danyball (dismissed):

.storybook/preview.js
raphaellueckl: We keep our ones, but "array.split" or something was not supported. But feel free to reduce them. That part caused a lot of work, but in the end, I never fully optimized it, now that you mention it. Good point.

BUT "our ones" are still there, just within that file. :)

danyball: It is still everything working with just our polyfills. Changed that and clicked trough all our components.

danyball:
JUST with npm start!!

We have 1 error on the console. But everything works. And its just on the first component you are clicking on. 

It seems that this is a open storybook-bug:
https://github.com/storybookjs/storybook/issues/10204

The error is gone if we remove `changelog` and `codepreview` addons.

danyball:
JUST with npm start!!

We have a second warning - but it is not reproducable:
`Warning: unmountComponentAtNode(): The node you're attempting to unmount was rendered by React and is not a top-level container. Instead, have the parent component update its state and rerender in order to remove this component.`

.storybook/preview.js
danyball: I have splitted that polyfill files into 2 imports:
- the additional polyfills for storybook 5.3
- our good old PL polyfills for our components

.storybook/main.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+    addons: [
+      'storybook-readme/register',
+      './.storybook/addons/changelog/register.js',
+      '@storybook/addon-knobs/register',
+      './.storybook/addons/codepreview/register.js',
+      '@storybook/addon-a11y/register'
+    ],
+};
```
01231, line 9: New line?

danyball:
I suggest to update directly to the new SB 6, which will be releases at the end of july (https://github.com/storybookjs/storybook/issues/9311). Because there is a problem at a SB internal file with the 5.3. 

Have installed the 6-RC and it works with just our PL polyfills. No extra pf needed. We just have to some extra work on our SB plugins. There are breaking changes on how to load them. But the showcases are working:
image


We can go these 2 ways:

1. Merge this PR with SB5.3 and the extra polyfills. After upgrading to 6 remove the extra polyfills.
2. Keep the branch of this PR as a basis for SB6 upgrade.

What do you think @NoahLiechti @LucaMele ?

01231:
I would choose 1. because we have some tasks for the new storybook/welcomepage design and I think the  design tool integration which storybook provides with version 5.3 would be helpful for these tasks.

LucaMele:
@danyball I would wait to the one that works without polyfill hacks. End of July is only in 2 weeks and everyone is on holiday anyway @NoahLiechti so i guess it wont hurt the designer if they have to wait 2 weeks in holiday times for a cleaner setup

danyball:
We decided to not merge this PR (see comments). All checks have passed on this PR right now. So, the branch is ready to upgrade to 6.X.

We are closing the PR but do not delete the branch.

PR #1879, opened 2020-07-17
bugfix/1878_gap-below-component

### LINKED ISSUE #1878: input-file & button-link: gap below component
There is always a little gap below an input-file/button-link.

Screenshot 2020-07-16 at 09 54 48

raphael-suter:
Fixes #1878 

# Done is when (DoD):

### DIFF
--- a/src/components/10-atoms/input-file/index.scss
+++ b/src/components/10-atoms/input-file/index.scss
@@ -1,3 +1,7 @@
+axa-input-file {
+  display: inline-block;
+}
+
 .a-input-file {
   @include text(t4);
   position: relative;
@@ -7,6 +11,7 @@
   padding: 0 30px;
   margin: 0;
   min-height: 40px;
+  vertical-align: bottom;
   hyphens: auto;
   overflow-wrap: break-word;
   word-wrap: break-word;
--- a/src/components/10-atoms/input-file/index.scss
+++ b/src/components/10-atoms/input-file/index.scss
@@ -6,6 +6,7 @@ axa-input-file {
   @include text(t4);
   position: relative;
   display: inline-block;
+  width: 100%;
   text-align: center;
   box-sizing: border-box;
   padding: 0 30px;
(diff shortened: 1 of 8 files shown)


src/components/10-atoms/input-file/index.scss
```diff
@@ -1,3 +1,7 @@
+axa-input-file {
```
danyball, line 1: This selector could be invalid if someone uses the new component versioning like `axa-input-file-pod_xy`. Better set display to the internal elements or create a wrapper div for that.

src/components/20-molecules/datepicker/index.js
```diff
@@ -1,28 +1,27 @@
 /* eslint-disable import/no-extraneous-dependencies */
+import AXADropdown from '@axa-ch/dropdown';
```
danyball, line 2: Something changed at that file? many changes, but just reordering?

danyball (changes requested):

src/components/20-molecules/datepicker/index.js
raphael-suter: No, there aren't any changes here, except the reordered imports (that was my commit to re-run the checks).

src/components/10-atoms/input-file/index.scss
raphael-suter: As discussed before, this shouldn't be a problem.

src/components/10-atoms/input-file/index.scss
```diff
@@ -1,3 +1,7 @@
...
+  display: inline-block;
```
danyball, line 2: Why do you add display? At my chrome everything works without it.

raphael-suter: Without it the component doesn't get selected properly & we could use it to allow 100% width. (Would have to change another display inside the component)

Screenshot 2020-07-21 at 08 49 07

danyball: Very nice!!! Please add this new functionality to the readme.

danyball (changes requested):

danyball (dismissed):

danyball (approved):

PR #1911, opened 2020-08-04
axa-dropdown: rename onchange event object item "name" to "optionLabel" 

### LINKED ISSUE #1910: Dropdown: Add the name of the component to onChange
From Slack channel pattern-lib-devs:

Use case seems like this:
```
formStore
{
   name: "hans" -> text input
   address: "123" -> text input
   lang: 'de' -> select(dropdown)
   job: 'troll' -> select(dropdown)
}
```
generic select handler
`setFormStore({ [event.currentTarget.name]: event.target.value })
` generic input text handler
`setFormStore({ [event.target.name]: event.target.value }

### LINKED ISSUE #1923: axa-dropdown: various differences to native select
axa-dropdown degrades to native &lt;select&gt; in narrow viewports, and can also be configured to always appear in `native` mode.

1. In this mode, a closed but focussed dropdown can be opened via up/down arrow keys. (n/a)
1. <del>Also, disabled items - realized as options - are greyed out but visible (n/a), and skipped over by keyboard navigation.</del>

In enhanced mode, the above points are not available (n/a), creating a difference in UX.

Resolve the differences  in favour of native &lt;select&gt;'s behaviour.

danyball:
Fixes #1910.
Fixed #1923.

The idea is to merge this PR first, then adapt https://github.com/axa-ch/patterns-library/pull/1920 slightly to the improved keyboard navigation code introduced here, then merge the latter, too.

# Done is when (DoD):

### DIFF
--- a/src/components/20-molecules/dropdown/README.md
+++ b/src/components/20-molecules/dropdown/README.md
@@ -35,16 +35,11 @@ export default AXADropdownReact;
 ```js
 <AXADropdownReact
   items={[
-    {
-      name: 'Please Select',
-      value: 'Please Select',
-      selected: true,
-      disabled: true,
-    },
     { name: 'Item 1', value: 'Item 1' },
     { name: 'Item 2', value: 'Item 2' },
     { name: 'Item 3', value: 'Item 3' },
   ]}
+  defaultTitle="Please Select"
   onChange={event => console.log(`value changed ${event.target.value}`)}
 />
 ```
@@ -59,8 +54,6 @@ The String-valued attribute `defaultTitle` sets the initial title of the closed 
 
 Its intended use is primarily for native-HTML situations where server-generated `items` describe the choices proper,
 and a separate title like `defaulttitle="Please select"` prompts the user to make a choice.
-
-See the above React example for an alternative parametrization strategy that relies on using `items` only.
 
 ### value
 
@@ -109,7 +102,6 @@ Is an array of objects to set the options of the dropdown. The objects must have
 | name     | The text that is displayed                                       |
 | value    | The value that is submitted via form submit                      |
 | selected | boolean: just set true to one of the options to preselect option |
-| disabled | boolean: disables the option                                     |
 
 ## Callback Properties
 
--- a/src/components/20-molecules/dropdown/README.md
+++ b/src/components/20-molecules/dropdown/README.md
@@ -1,5 +1,7 @@
 # AXA Dropdown
 
+This component represents a styled, enhanced replacement for native HTML &lt;select&gt;. The component is responsive, falling back to native look-and-feel under mobile viewports. Keyboard operation is supported.
+
 ## Usage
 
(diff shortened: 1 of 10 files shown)


src/components/20-molecules/dropdown/index.js
```diff
@@ -271,15 +271,20 @@ class AXADropdown extends NoShadowDOM {
...
+    const [{ label }] = this.findByValue(value);
     // allow idiomatic event.target.value in onChange callback!
-    const syntheticEvent = { target: { value, index: integerIndex, name } };
+    const details = {
+      value,
+      name,
+      optionIndex: integerIndex,
+      optionLabel: label,
+    };
+    const syntheticEvent = { target: details };
```
danyball, line 300: We are mixing the Option Object (https://www.w3schools.com/jsref/dom_obj_option.asp) with the Event Object of onChange. 
Why? To simulate the `target` at the syntheticEvent.

HTML Standard for `event.target.name` is the name of the select component. There is no "e.target.index" or "e.target.label". So i added them with the specific name "optionIndex". Its the index of the option, not the index of the select.

markus-walther: I reverted the optionIndex part only (back to `index`), keeping the optionLabel. The reason is that we would like to remain backwards compatible and not break application code unnecessarily. Apps that relied on `index` do not have to be updated, since nothing changed semantically. optionLabel in contrast represents a meaningful change, but as a new key in the event object is not in danger of conflicting with existing app code.

src/components/20-molecules/dropdown/README.md
```diff
@@ -98,13 +100,32 @@ If you do not set this attribute, all the options will be displayed without any
 
 This property is most useful in space-constrained scenarios or with many options, e.g. in a country selector.
 
+### items
+
+Is an array of objects to set the options of the dropdown. The objects must have a `name` and `value` set. See an example at the top of this Readme.
+
+| item     | Details                                                          |
+| -------- | ---------------------------------------------------------------- |
+| name     | The text that is displayed                                       |
```
01231, line 102: What about calling `name`--> `label` as suggested in #1910

src/components/20-molecules/dropdown/index.js
danyball: Dont like it.

danyball (commented):
Please dont forget to add tests.

src/components/20-molecules/dropdown/README.md
danyball: Markus and myself decided to not change it because its a unnecessary breaking change.

src/components/20-molecules/dropdown/index.js
markus-walther: As discussed orally, we have now consensus that the revert of optionIndex => index is justified, since  it narrows the potential range of users who might suffer consequences of a breaking change.

markus-walther:
> Please dont forget to add tests.

I just did.

PR #1914, opened 2020-08-04
Fix axa-footer delayed-children problem

### LINKED ISSUE #1890: axa-footer sometimes does not style its children correctly
Suppose axa-footer is used natively. Also suppose that the defining JavaScript _dist_ bundle is loaded and evaluated prior to the occurrence of a `<axa-footer>` tag with its slotted children.

In this situation, the browser's HTML parser, upon encountering the opening `<axa-footer>` tag, may immediately construct the element and also render it.

At this point in time its children are not yet in the DOM tree.

As a result the visual rendering of this component is wrong.

Fix this, and properly accomodate all possible construction orders of the component vs its children.

Reproduction: https://codepen.io/tsaadaasmarino/pen/VweqVJa

markus-walther:
Fixes #1890.

Tested by simple HTML page consisting of **inline script** with dist export, followed by axa-footer with lots of children.
It is hypothesized that an inline script is going to be loaded **and** evaluated much faster than a network-loaded (i.e. `src`) script, which is why the subtle bug mentioned in #1890 deterministically occurs in this scenario, and only if the script is first in textual order.

The bigger task of making the entire footer children dynamically updateable has to wait for another day, as the convoluted implementation makes a simple modification towards supporting this scenario difficult to implement for now.

# Done is when (DoD):

### DIFF
--- a/src/components/30-organisms/footer/index.js
+++ b/src/components/30-organisms/footer/index.js
@@ -1,4 +1,3 @@
-// TODO fix that stuff
 /* eslint-disable import/no-extraneous-dependencies */
 import { html, svg, css, unsafeCSS } from 'lit-element';
 import { repeat } from 'lit-html/directives/repeat';
@@ -40,8 +39,8 @@ class AXAFooter extends InlineStyles {
     `;
   }
 
-  // Parent class InlineStyles needs a static method to retrive styles
-  // name of such method is passed when calling: this.inlineStyles('resetHeadingCss');
+  // Parent class InlineStyles needs a static method to retrieve styles;
+  // the name of such method is passed when calling this.inlineStyles('resetHeadingCss');
   static get resetHeadingCss() {
     return childStyles;
   }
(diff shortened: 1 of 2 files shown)


src/components/30-organisms/footer/index.js
```diff
@@ -85,13 +84,13 @@ class AXAFooter extends InlineStyles {
   }
 
   /**
-   * A slot element needs to be a direct child of a webcomponent. Since AEM
+   * A slot element needs to be a direct child of a web component. Since AEM
```
01231, line 87: thanks for cleaning up all comments👍🏼

src/components/30-organisms/footer/index.js
```diff
@@ -361,6 +378,12 @@ class AXAFooter extends InlineStyles {
       }
     }
   };
+
+  disconnectedCallback() {
```
01231, line 382: nice

01231 (dismissed):
Tested and works!

PR #1958, opened 2020-09-08
remove console warning "Duplicate title: 'Welcome'"

01231:
Removes warning from welcomepage PR

### DIFF
--- a/src/other/pages/what-is-new/story.js
+++ b/src/other/pages/what-is-new/story.js
@@ -1,4 +1,5 @@
 import { html, render } from 'lit-html';
+import { storiesOf } from '@storybook/html';
 import '../../../components/10-atoms/text';
 import '../../../components/10-atoms/heading';
 import styles from './index.scss';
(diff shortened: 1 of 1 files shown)


src/other/pages/what-is-new/story.js
```diff
@@ -86,6 +86,6 @@ export const WhatIsNew = () => {
     ${contact}
   `;
 
-  render(template, wrapper);
-  return wrapper;
-};
+    render(template, wrapper);
+    return wrapper;
+  });
```
danyball, line 91: Why do we have all these whitespace changes? Dont we use the same prettier config? Or is that eslint? Please check your IDE

danyball (approved):

src/other/pages/what-is-new/story.js
NoahLiechti: I created the file. After only one person edited the file. I have nothing changed in eslint or prettier since then.
If I was the only person to edit this files, my guess would be that there wouldn't be whitespace changes at all. So I think the person who edited the files must has different settings.

01231: The whitespace is justified. Look on how the structure of the code changed.

PR #1960, opened 2020-09-09
Feature/1841 currency

### LINKED ISSUE #1841: Input type text add new property "currency"
It was requested, that we allow automatic separators and rounding for input fields.

For instance: "10250.123" => "CHF 10'250.15"

We will support the INTL Format api, available native in the browser (https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat)

Currency attribute should receive currency value like so: `currency="CHF"`. Locale can be fixed to `de-CH`.

danyball:
Fixes #1841

# Merge Checklist:
- [x] Tests are sufficient.

- [x] is popup change ok? --> #1969

### DIFF
--- a/src/components/10-atoms/input-text/CHANGELOG.md
+++ b/src/components/10-atoms/input-text/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 4.4.0
 
-- You can now set `currency` attribute. This will format the input `1000` for example to `CHF 1’000.00`. For details see README.md.
+- You can now set a `currency` attribute. This will format the input `1000` for example to `CHF 1’000.00`. For details see README.md.
 
 ## 4.3.1
 
--- a/src/components/10-atoms/input-text/CHANGELOG.md
+++ b/src/components/10-atoms/input-text/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.4.0
+
+- You can now set `currency` attribute. This will format the input `1000` for example to `CHF 1’000.00`. For details see README.md.
+
 ## 4.3.1
 
 - Fix: prevent duplicate style attachment. (#1727)
--- a/src/components/10-atoms/input-text/README.md
+++ b/src/components/10-atoms/input-text/README.md
@@ -160,6 +160,14 @@ It is a string value. Useful to open up numeric keyboard on touch devices. See o
 
 The Boolean attribute autofocus, when true, sets keyboard focus on the underlying native &lt;input&gt; element after initial rendering of the component.
 
+### currency
+
+You can set this string value to format numbers. It formats the input when the `onBlur` event of the input element is fired or if you set `value` programmatically. You can set `ISO 4217` [currency codes](https://en.wikipedia.org/wiki/ISO_4217), but the locale "style" is fixed to `de-CH`.
+
+The formatter is just working if attribute `type` is set to `text`.
+
+Please be aware of using this attribute in combination with `maxLength`. The result of formatting could exceed the limits of `maxLength`!
+
 ## Callbacks
 
 ### onChange
--- a/src/components/10-atoms/input-text/README.md
+++ b/src/components/10-atoms/input-text/README.md
@@ -164,7 +164,7 @@ The Boolean attribute autofocus, when true, sets keyboard focus on the underlyin
 
 You can set the `currency` attribute to a valid [ISO 4217 currency code](https://en.wikipedia.org/wiki/ISO_4217) string. When set, &lt;input-text&gt; formats the input whenever the `onBlur` event of its internal input element fires, and also upon setting `value` programmatically. Note that the locale "style" remains fixed to `de-CH`.
 
-Currency formatting is only active when attribute `type` is set to `text`. If the inputted text can not be formatted the attribute `invalid` switches to `true`. A valid input to format is for example `12.20`, invalid is `12.20.1` because of the two dots.
+Currency formatting is only active when attribute `type` is set to `text`. If the inputted text can not be formatted the attribute `invalid` switches to `true`. The Formatter do not accept values which are only letters ("onehundret") or has more than 1 decimal separators ("12.30.1").
 
 Please be mindful of using currency formatting in combination with `maxLength`. The result of formatting could exceed the limits of `maxLength`!
 
--- a/src/components/10-atoms/input-text/README.md
+++ b/src/components/10-atoms/input-text/README.md
@@ -164,7 +164,7 @@ The Boolean attribute autofocus, when true, sets keyboard focus on the underlyin
 
 You can set the `currency` attribute to a valid [ISO 4217 currency code](https://en.wikipedia.org/wiki/ISO_4217) string. When set, &lt;input-text&gt; formats the input whenever the `onBlur` event of its internal input element fires, and also upon setting `value` programmatically. Note that the locale "style" remains fixed to `de-CH`.
 
-Currency formatting is only active when attribute `type` is set to `text`. If the inputted text can not be formatted the attribute `invalid` switches to `true`.
+Currency formatting is only active when attribute `type` is set to `text`. If the inputted text can not be formatted the attribute `invalid` switches to `true`. A valid input to format is for example `12.20`, invalid is `12.20.1` because of the two dots.
 
 Please be mindful of using currency formatting in combination with `maxLength`. The result of formatting could exceed the limits of `maxLength`!
 
--- a/src/components/10-atoms/input-text/README.md
+++ b/src/components/10-atoms/input-text/README.md
@@ -164,7 +164,7 @@ The Boolean attribute autofocus, when true, sets keyboard focus on the underlyin
 
 You can set the `currency` attribute to a valid [ISO 4217 currency code](https://en.wikipedia.org/wiki/ISO_4217) string. When set, &lt;input-text&gt; formats the input whenever the `onBlur` event of its internal input element fires, and also upon setting `value` programmatically. Note that the locale "style" remains fixed to `de-CH`.
 
-Currency formatting is only active when attribute `type` is set to `text`.
+Currency formatting is only active when attribute `type` is set to `text`. If the inputted text can not be formatted the attribute `invalid` switches to `true`.
 
 Please be mindful of using currency formatting in combination with `maxLength`. The result of formatting could exceed the limits of `maxLength`!
 
--- a/src/components/10-atoms/input-text/README.md
+++ b/src/components/10-atoms/input-text/README.md
@@ -162,11 +162,11 @@ The Boolean attribute autofocus, when true, sets keyboard focus on the underlyin
 
 ### currency
 
-You can set this string value to format numbers. It formats the input when the `onBlur` event of the input element is fired or if you set `value` programmatically. You can set `ISO 4217` [currency codes](https://en.wikipedia.org/wiki/ISO_4217), but the locale "style" is fixed to `de-CH`.
+You can set the `currency` attribute to a valid [ISO 4217 currency code](https://en.wikipedia.org/wiki/ISO_4217) string. When set, &lt;input-text&gt; formats the input whenever the `onBlur` event of its internal input element fires, and also upon setting `value` programmatically. Note that the locale "style" remains fixed to `de-CH`.
 
-The formatter is just working if attribute `type` is set to `text`.
+Currency formatting is only active when attribute `type` is set to `text`.
 
-Please be aware of using this attribute in combination with `maxLength`. The result of formatting could exceed the limits of `maxLength`!
+Please be mindful of using currency formatting in combination with `maxLength`. The result of formatting could exceed the limits of `maxLength`!
 
 ## Callbacks
 
--- a/src/components/10-atoms/input-text/index.js
+++ b/src/components/10-atoms/input-text/index.js
@@ -257,7 +257,6 @@ class AXAInputText extends AXAPopupMixin(NoShadowDOM) {
 
     if (currency && type === 'text') {
       const hasAtLeastOneDigit = /\d/.test(value);
-      const dotsCounted = (value.match(/\./g) || []).length;
 
       if (!this.currencyFormatter) {
         // just create a new Intl if it does not exist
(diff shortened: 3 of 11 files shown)


danyball:
Open issues:

- [x] chars left wrong

- [x] currency input: user type "hello" > what should happen?

src/components/20-molecules/popup/index.js
```diff
@@ -7,7 +7,7 @@ export * from './popup-mixin';
 /* eslint-disable no-undef */
 const axaPopupVersion = __VERSION_INFO__;
 /* eslint-enable no-undef */
-const commonVersion = axaPopupVersion['axa-popup']['axa-popup'];
+const commonVersion = axaPopupVersion['axa-popup'];
```
danyball: This has nothing to do with the currency feature directly. But without that change the Unit Tests are failing. @markus-walther can you please check that this change is ok?

danyball:
Tests laufen alle lokal durch

>  408 passed (7m 30s)

src/components/10-atoms/input-text/index.js
```diff
@@ -290,6 +319,12 @@ class AXAInputText extends AXAPopupMixin(NoShadowDOM) {
       _open,
     } = this;
 
+    let formattedValue = value;
+    if (document.activeElement !== this.nativeInput) {
+      // Do not format the value if the user is still typing. We will format its input onBlur.
```
markus-walther: onBlur => on blur. (Comments should be in plain English, onBlur is not a word)

src/components/10-atoms/input-text/ui.test.js
```diff
@@ -209,3 +209,18 @@ test('should display correct default value', async t => {
...
+test('should format value', async t => {
+  const $axaInputElement = await Selector(() => document.querySelector(TAG), {
+    dependencies: { TAG },
+  }).find(CLASS);
+
+  await t.typeText($axaInputElement, '1234', { replace: true });
+  await t.pressKey('tab'); // to blur input-element
+
+  await t.expect($axaInputElement.value).eql('CHF 1’234.00');
+});
```
markus-walther, line 226: I am missing a controlled-mode test. For that it should be defined what controlled mode means when currency is active...

markus-walther (changes requested):
Found some small things, otherwise good work!

Will check later why the popup issue you find was not causing problems before...

src/components/10-atoms/input-text/index.js
```diff
@@ -248,6 +252,31 @@ class AXAInputText extends AXAPopupMixin(NoShadowDOM) {
...
+    if (currency && type === 'text' && !this.currencyFormatter) {
+      // just create a new Intl if it does not exist
+      this.currencyFormatter = new Intl.NumberFormat('de-CH', {
+        style: 'currency',
+        currency,
+      });
+    }
+
+    if (this.currencyFormatter) {
+      const containsNumber = /\d/g;
```
raphaellueckl: does this regex check if a value HAS a number, or if it ONLY is a number (and no letters?). because then the name is a bit misleading.

src/components/10-atoms/input-text/index.js
```diff
@@ -248,6 +252,31 @@ class AXAInputText extends AXAPopupMixin(NoShadowDOM) {
...
+
+    if (currency && type === 'text' && !this.currencyFormatter) {
+      // just create a new Intl if it does not exist
+      this.currencyFormatter = new Intl.NumberFormat('de-CH', {
+        style: 'currency',
+        currency,
+      });
+    }
+
+    if (this.currencyFormatter) {
```
raphaellueckl: i don't really like the structure of those if-else blocks in here.

line 266 checks if currencyformatter is set, which is an obsolete check if the type is "number" (which is already checked previously). this is not a clean approach.

better would be (pseudocode):

```
_formatCurrency(value) {
  if (currency && type=="number") { // if not type currency or number, don't look any further...
    if (!this.currencyFormatter) {
       // INIT
    }
    // proceed the formatting...
  }
  return value
}
`
``

**Review by raphaellueckl — CHANGES_REQUESTED** [review]

File: src/components/10-atoms/input-text/index.js (hunk 4 above)
Comment (danyball) [reply src/components/10-atoms/input-text/index.js]: HAS. Because only if input contains a number the formatter can work correctly

File: src/components/10-atoms/input-text/ui.test.js (hunk 3 above)
Comment (danyball) [reply src/components/10-atoms/input-text/ui.test.js:226]: done.

File: src/components/10-atoms/input-text/index.js (hunk 4 above)
Comment (01231) [reply src/components/10-atoms/input-text/index.js]: what about `const isNumber`?

Comment (danyball) [reply src/components/10-atoms/input-text/index.js]: Do not change the variable, name is perfect.

File: src/components/10-atoms/input-text/index.js (hunk 5 above)
Comment (danyball) [reply src/components/10-atoms/input-text/index.js]: done

**Review by raphaellueckl — DISMISSED** [review]

File: src/components/10-atoms/input-text/CHANGELOG.md [hunk 6]
```diff
@@ -1,3 +1,7 @@
+## 4.4.0
+
+- You can now set `currency` attribute. This will format the input `1000` for example to `CHF 1’000.00`. For details see README.md.
```
markus-walther: wording: A `currency` attribute

src/components/10-atoms/input-text/README.md
```diff
@@ -160,6 +160,14 @@ It is a string value. Useful to open up numeric keyboard on touch devices. See o
 
 The Boolean attribute autofocus, when true, sets keyboard focus on the underlying native &lt;input&gt; element after initial rendering of the component.
 
+### currency
+
+You can set this string value to format numbers. It formats the input when the `onBlur` event of the input element is fired or if you set `value` programmatically. You can set `ISO 4217` [currency codes](https://en.wikipedia.org/wiki/ISO_4217), but the locale "style" is fixed to `de-CH`.
```
markus-walther: Wording:

You can set the `currency` attribute to a valid [ISO 4217 currency code](https://en.wikipedia.org/wiki/ISO_4217) string. When set, &lt;input-text&gt; formats the input whenever the `onBlur` event of its internal input element  fires, and also upon setting `value` programmatically.  Note that the locale "style" remains fixed to `de-CH`.

src/components/10-atoms/input-text/README.md
```diff
@@ -160,6 +160,14 @@ It is a string value. Useful to open up numeric keyboard on touch devices. See o
...
+
+The formatter is just working if attribute `type` is set to `text`.
```
markus-walther: Wording: Currency formatting is only active when attribute ....

src/components/10-atoms/input-text/README.md
```diff
@@ -160,6 +160,14 @@ It is a string value. Useful to open up numeric keyboard on touch devices. See o
...
+
+Please be aware of using this attribute in combination with `maxLength`. The result of formatting could exceed the limits of `maxLength`!
```
markus-walther: Wording: Please be mindful of using currency formatting in combination...

src/components/10-atoms/input-text/index.js
```diff
@@ -248,6 +252,31 @@ class AXAInputText extends AXAPopupMixin(NoShadowDOM) {
     this.nativeInput = this.querySelector('input');
   }
 
+  _formatCurrency(value) {
+    const { currency, type } = this;
+    const containsNumber = /\d/g;
```
markus-walther: Remove the `g` after the regexp, it serves no purpose here.

The semantics is: has at least 1 digit somewhere, you might want to rename the variable to express this better.

src/components/10-atoms/input-text/index.js
```diff
@@ -248,6 +252,31 @@ class AXAInputText extends AXAPopupMixin(NoShadowDOM) {
...
+    if (currency && type === 'text') {
+      if (!this.currencyFormatter) {
+        // just create a new Intl if it does not exist
+        this.currencyFormatter = new Intl.NumberFormat('de-CH', {
+          style: 'currency',
+          currency,
+        });
+      }
+
+      if (containsNumber.test(value)) {
```
markus-walther: containsNumber is not reused, so in better DRY style you'd write

const hasAtLeastOneDigit = /\d/.test(value) above and here only if(hasAtLeastOneDigit) { ...}

src/components/10-atoms/input-text/index.js
```diff
@@ -248,6 +252,31 @@ class AXAInputText extends AXAPopupMixin(NoShadowDOM) {
...
+        this.invalid = false;
+        const valueDecimalsOnly = value.replace(/[^0-9.]/g, '');
+        return this.currencyFormatter.format(valueDecimalsOnly);
```
markus-walther: This means that `mickeymouse19newyork.foobar34` is a valid input... is it good UX to be this lenient?

src/components/10-atoms/input-text/index.js
```diff
@@ -248,6 +252,31 @@ class AXAInputText extends AXAPopupMixin(NoShadowDOM) {
...
   }
 
+  _formatCurrency(value) {
+    const { currency, type } = this;
+    const containsNumber = /\d/g;
+
+    if (currency && type === 'text') {
+      if (!this.currencyFormatter) {
+        // just create a new Intl if it does not exist
+        this.currencyFormatter = new Intl.NumberFormat('de-CH', {
```
markus-walther, line 263: Why does the currency formatter need to be part of the element's state?

src/components/10-atoms/input-text/index.js
```diff
@@ -290,6 +319,12 @@ class AXAInputText extends AXAPopupMixin(NoShadowDOM) {
       _open,
     } = this;
 
+    let formattedValue = value;
+    if (document.activeElement !== this.nativeInput) {
```
markus-walther, line 329: comment missing: // focus not on internal input?

markus-walther (changes requested):
Good job! Some small findings, please address.

src/components/20-molecules/popup/index.js
markus-walther: This change is definitely NOT ok. It leads to definitions like &lt;axa-popup-button-objectObject&gt; ...
Please revert it and adapt the unit tests instead.

src/components/10-atoms/input-text/index.js
danyball: It is not needed. But i dont find a way to unit test `currencyFormatter` if its a variable of the module.

src/components/10-atoms/input-text/index.js
danyball: Will ask thomas, but thats ok for me.

src/components/10-atoms/input-text/index.js
danyball: The comment is at line 325? Or what do you mean?

src/components/10-atoms/input-text/index.js
danyball: It is ok for Thomas.

src/components/20-molecules/popup/index.js
danyball: Reverted and fixed.

src/components/10-atoms/input-text/index.js
01231: you could move the comment one line up

src/components/10-atoms/input-text/index.js
01231: Good input. Can you implement this?

01231 (dismissed):
Nice

src/components/10-atoms/input-text/index.js
```diff
@@ -248,6 +252,33 @@ class AXAInputText extends AXAPopupMixin(NoShadowDOM) {
...
+        this.currencyFormatter = new Intl.NumberFormat('de-CH', {
+          style: 'currency',
+          currency,
+        });
+      }
+
+      if (hasAtLeastOneDigit && dotsCounted <= 1) {
+        this.invalid = false;
+        const valueDecimalsOnly = value.replace(/[^0-9.]/g, '');
+        return this.currencyFormatter.format(valueDecimalsOnly);
```
markus-walther: This is a brittle way to solve the NaN bug that arises with a specific input that has multiple dots. There could be many other patterns.

You need to address the root of the problem with `parseFloat(valueDecimalsOnly)`. If it  parses as a valid floating-point number, it could be used as a currency value.

src/components/10-atoms/input-text/index.js
markus-walther: What I mean is that it is super-unobvious what `if (document.activeElement !== this.nativeInput) ` does without the comment I suggested, placed above or at the end of line 325.

Principle: Any code that requires some deep thinking needs a comment that explains the Why.

This line definitely requires deep thinking or even googling if you didn't do it recently...

markus-walther (dismissed):
Have found two things only this time! After that, it should be fine.

src/components/10-atoms/input-text/index.js
```diff
@@ -267,10 +266,14 @@ class AXAInputText extends AXAPopupMixin(NoShadowDOM) {
...
-        this.invalid = false;
+      if (hasAtLeastOneDigit) {
         const valueDecimalsOnly = value.replace(/[^0-9.]/g, '');
-        return this.currencyFormatter.format(valueDecimalsOnly);
+
+        // eslint-disable-next-line no-restricted-globals
+        if (!isNaN(valueDecimalsOnly)) {
+          this.invalid = false;
+          return this.currencyFormatter.format(valueDecimalsOnly);
+        }
```
01231: Would use parseFloat(valueDecimalsOnly) as @markus-walther recommended. It is cleaner and more speaking.

01231 (changes requested):

src/components/10-atoms/input-text/index.js
01231:
```
if (hasAtLeastOneDigit) {
  const valueDecimalsOnly = value.replace(/[^0-9.]/g, '');
  return this.currencyFormatter.format(parseFloat(valueDecimalsOnly));
 }
```

src/components/10-atoms/input-text/README.md
```diff
@@ -164,7 +164,7 @@ The Boolean attribute autofocus, when true, sets keyboard focus on the underlyin
 
 You can set the `currency` attribute to a valid [ISO 4217 currency code](https://en.wikipedia.org/wiki/ISO_4217) string. When set, &lt;input-text&gt; formats the input whenever the `onBlur` event of its internal input element fires, and also upon setting `value` programmatically. Note that the locale "style" remains fixed to `de-CH`.
 
-Currency formatting is only active when attribute `type` is set to `text`. If the inputted text can not be formatted the attribute `invalid` switches to `true`.
+Currency formatting is only active when attribute `type` is set to `text`. If the inputted text can not be formatted the attribute `invalid` switches to `true`. A valid input to format is for example `12.20`, invalid is `12.20.1` because of the two dots.
```
01231: would write that letters only are also invalid. i wouldn't write that 12.20 is valid because this makes the impression that only this format is allowed. Better just write what isn't allowed (to dots and only letters). there are too many allowed input formats.

01231 (changes requested):

PR #1970, opened 2020-09-22
feature/1968-axa-input-text-add-onInput

### LINKED ISSUE #1968: axa-input-text add onKeyUp support for react
```
function myMethod(event) {
  console.log(event)
}

// ...

<axa-input-text onKeyUp={myMethod}>
```

Requested by @urswiss

urswiss:
Fixes #1968 

# Merge Checklist:
- [ ] Code selfreview has been performed ([Best Practices](https://github.com/axa-ch/patterns-library/blob/develop/CONTRIBUTION.md#best-practices)).
- [ ] Tests are sufficient.
- [ ] Modifications still work in IE.

### DIFF
--- a/src/components/10-atoms/input-text/index.js
+++ b/src/components/10-atoms/input-text/index.js
@@ -60,7 +60,7 @@ class AXAInputText extends AXAPopupMixin(NoShadowDOM) {
       onChange: { type: Function, attribute: false },
       onFocus: { type: Function, attribute: false },
       onBlur: { type: Function, attribute: false },
-      onInputfieldKeyUp: { type: Function, attribute: false },
+      onInput: { type: Function, attribute: false },
     };
   }
 
@@ -338,7 +338,6 @@ class AXAInputText extends AXAPopupMixin(NoShadowDOM) {
                   @input="${this.handleInput}"
                   @focus="${this.handleFocus}"
                   @blur="${this.handleBlur}"
-                  @keyup="${this.onInputfieldKeyUp}"
                 />
               `
             : html`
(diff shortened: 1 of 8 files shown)


marek-polak (commented):
I think we should rather use 'onInput' event instead - as mentioned in this question: https://stackoverflow.com/questions/574941/best-way-to-track-onchange-as-you-type-in-input-type-text, using `onkey*` events is not reliable, since you can right-click the field and choose Paste, and this will change the field without any keyboard input.

Instead, we should use `onInput` as answered few posts below (https://stackoverflow.com/a/26202266)
> It feels like onchange without the need to lose focus on the element. It is HTML5.
> It’s supported by everyone (even mobile), except IE8 and below. 

I noticed there is already support for `onInput` in mixin, so maybe all it needs to be done is to define it in typing for react?

urswiss:
valid point. Ok if I adapt the PR accordingly, @raphaellueckl ?

src/components/10-atoms/input-text/index.js
```diff
@@ -60,6 +60,7 @@ class AXAInputText extends AXAPopupMixin(NoShadowDOM) {
       onChange: { type: Function, attribute: false },
       onFocus: { type: Function, attribute: false },
       onBlur: { type: Function, attribute: false },
+      onInputfieldKeyUp: { type: Function, attribute: false },
```
raphaellueckl: I would rename it to `onKeyUp`

danielnestler (dismissed):
OK

raphaellueckl:
Closing as there is another PR based on this one, merge conflics resolved.

https://github.com/axa-ch/patterns-library/pull/1975

PR #1980, opened 2020-09-28
Fix datepicker issues

### LINKED ISSUE #1974: AxaDatepicker : defaultValue does not affect preselected value in Picker.
defaultValue in Datepicker sets the input field, however when user clicks on the Picker, Today is selected instead of the Date in input. Note: If input is clicked first then picker opened the date is correct.

## Expected Behavior
The default value in picker is same as value in defaultValue

## Current Behavior
The defaultValue does not define, the preselection in picker

## Steps to Reproduce
Demo:
1. https://axa-ch.github.io/esales-ui-components/?path=/story/common-axa-datepicker-pure--default-with-validations
2. Open picker. Today is selected instead of predefined date: 2001-05-04
image

## Context (Environment)
Usability of Picker by the users is impacted.

### LINKED ISSUE #1977: axa-datepicker: do not auto-select dates
Screenshot 2020-09-28 at 10 40 59

As the above screenshot shows, the current date's day may be selected despite a completely underspecified datepicker, i.e. no date-relevant properties have been set.

Furthermore, the `allowedyears` range did include the current year (2020). If the current year is included, it  should be the default rather than the _beginning of the range_ (1971), which should be reserved as a measure of last resort to avoid a completely empty datepicker-month-matrix display.

Desired UX is, however, to only select a date via a) user interaction or b) set properties.

Once a date is selected, it should only be visible for the month and year in which it was set, i.e. navigation away from that month's calendar sheet should clear the selection UI, and later navigation back to said month and year should make it reappear.

Fix this bug.

### LINKED ISSUE #1981: AxaDatepicker: Picker opens on bottom even when there is not enough space there
When Picker in Datepicker is opened the first time, it gets opened in the bottom disregarding if there is space or not.
Tested version: "@axa-ch/datepicker": "^12.0.0",

## Expected Behavior
Positioning of the picker(Top/Bottom) is correct the first time picker is opened, not only after scrolling/resizing the viewport.

## Steps to Reproduce
1. https://axa-ch.github.io/esales-ui-components/?path=/story/common-axa-datepicker--with-some-knobs
2. remove the bottom tab (with knobs and actions)
3. Click on the picker
4. Picker is opened bottom, even there is not enough space
5. After scrolling the Picker moves correctly to top
6. Reopening opens the Picker top. Yet this behaviour is expected the first time picker is opened

## Context (Environment)
This forces the user to scroll, not nice in straight-forward wizard (esales-pod-hh)

markus-walther:
Fixes #1974.
Fixes #1977.
Fixes #1981. closes #2001 

This has been discussed a bit with Thomas Veit.

# Merge Checklist:
- [x] Tests are sufficient.

### DIFF
--- a/src/components/20-molecules/datepicker/index.js
+++ b/src/components/20-molecules/datepicker/index.js
@@ -564,6 +564,7 @@ class AXADatepicker extends NoShadowDOM {
   updated(changedProperties) {
     if (changedProperties.has('open') && this.open) {
       applyEffect(this);
+      this.handleViewportCheck(this.input);
     }
   }
 
@@ -598,7 +599,7 @@ class AXADatepicker extends NoShadowDOM {
       // yes, so in order to prevent the datepicker from showing an "empty" month sheet
       // without any days to select, force-pick the starting year as the first of the range of allowed years
       const [newStartYear] = this.allowedyears;
-      this._year = newStartYear;
+      this.year = newStartYear;
     }
 
     const { _year, _month, _day, allowedyears, locale, startDate } = this;
--- a/src/components/20-molecules/datepicker/index.js
+++ b/src/components/20-molecules/datepicker/index.js
@@ -6,7 +6,6 @@ import {
   Keyboard_arrow_leftSvg,
   Keyboard_arrow_rightSvg,
 } from '@axa-ch/materials/icons';
-import { formatISO } from 'date-fns';
 import { html, svg } from 'lit-element';
 import { classMap } from 'lit-html/directives/class-map';
 import {
@@ -143,6 +142,11 @@ class AXADatepicker extends NoShadowDOM {
   }
 
   static get properties() {
+    const now = new Date();
+    // formally mark the (internal) default date, s.t. it can be distinguished
+    // from dates that are set by applications (exploiting the malleability
+    // of system objects...)
+    now.type = 'default';
     return {
       'data-test-id': { type: String, reflect: true },
       open: { type: Boolean, reflect: true },
@@ -156,7 +160,7 @@ class AXADatepicker extends NoShadowDOM {
           fromAttribute: value => (value ? new Date(value) : null),
           toAttribute: value => (value ? value.toISOString() : null),
         },
-        defaultValue: new Date(),
+        defaultValue: now,
       },
       outputdate: { type: String, reflect: true },
       year: { type: Number, reflect: true, defaultValue: undefined },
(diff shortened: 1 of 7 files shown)


src/components/20-molecules/datepicker/index.js
```diff
@@ -143,6 +145,11 @@ class AXADatepicker extends NoShadowDOM {
   }
 
   static get properties() {
+    const now = new Date();
+    // formally mark the (internal) default date, s.t. it can be distinguished
+    // from dates that are set by applications (exploiting the malleability
+    // of system objects...)
+    now.type = 'default';
```
danyball, line 149: Isn't that dangerous if the standard class Date adds its own prop `type` in future? Maybe name it like `axaType`?

src/components/20-molecules/datepicker/index.js
```diff
@@ -550,32 +550,50 @@ class AXADatepicker extends NoShadowDOM {
...
 
-    this._date = overrideDate(year, month, day, startDate);
+    const status = {};
+    this._date = overrideDate(year, month, day, startDate, status);
     const { _date } = this;
+    // did the start date get actually overridden via one or more of the
+    // date parts (year, month, day)?
+    if (status.set) {
+      // yes, so the resulting _date is application-selected
+      delete _date.type;
```
danyball, line 613: Is it possible to not delete this prop `type` to have more debug informations?

src/components/20-molecules/datepicker/index.js
```diff
@@ -104,15 +103,17 @@ export const parseAndFormatAllowedYears = (allowedyears = [], setYear) => {
 };
 
 // gather all native Date-object setters in one place
-const overrideDate = (year, month, day, date) => {
+const overrideDate = (year, month, day, date, status = {}) => {
   let _day = day;
 
   if (typeof year === 'number' && year >= 0) {
     date.setFullYear(year);
+    status.set = true;
```
danyball: Do you need a object `status` here? is a boolean variable sufficient?

markus-walther: This function returns a `Date` instance. If we don't want to change how the function is _used_, we have no choice but to pass in a modifiable object, and set a key in that to propagate that info outwards.

This way of doing things is the least invasive option, IMHO, that's why I chose it.

src/components/20-molecules/datepicker/index.js
markus-walther: Browser system classes like `Date` are long-term stable, any change must go through a lengthy spec process. We can wait for that to happen...

src/components/20-molecules/datepicker/index.js
markus-walther: It would be incorrect, since  this is semantically no longer a default date. To reflect this, I manipulate the date so that the assertion `_date.type === 'default'` is no longer true. 

This assertion can itself serve as debug info...

danyball (commented):
Have not understood everything down to the smallest detail, but it works and looks quite clean.

danyball (approved):

PR #1982, opened 2020-09-29
Fix versioning for axa-text

### LINKED ISSUE #1979: Text: If version is filled, Variant property does not work
When Version is filled in createAxaText, the Variant property does not Work for Text.

## Expected Behavior
Variant property of Text should work independently of Version

## Steps to Reproduce
1. Create AxaText with version

const version = "esales";
const AxaTextReact = createAXAText(createElement, version);
export const SomeText = () => <AxaTextReact variant="bold">Text</AxaTextReact>

Der text ist nicht bold wie er sein sollte
Wenn mann aber die Version entfernt funktioniert das wie erwartet

## Context (Environment)
We version all components of PL due to possible conflicts with other pods. In this case it is not possible

markus-walther:
Fixes #1979.

# Merge Checklist:
- [x] Tests are sufficient.

### DIFF
--- a/src/components/10-atoms/text/index.js
+++ b/src/components/10-atoms/text/index.js
@@ -1,4 +1,3 @@
-import { css, unsafeCSS } from 'lit-element';
 import NoShadowDOM from '../../../utils/no-shadow';
 
 /* eslint-disable import/no-extraneous-dependencies */
@@ -16,9 +15,7 @@ class AXAText extends NoShadowDOM {
   }
 
   static get styles() {
-    return css`
-      ${unsafeCSS(styles)}
-    `;
+    return styles;
   }
 
   static get properties() {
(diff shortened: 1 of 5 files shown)


danyball (approved):
minimal change, nice

src/components/10-atoms/text/index.js
```diff
@@ -16,9 +15,7 @@ class AXAText extends NoShadowDOM {
   }
 
   static get styles() {
-    return css`
-      ${unsafeCSS(styles)}
-    `;
+    return styles;
```
01231, line 18: shouldn't this be changed on all components then?

01231: for example cookie-disclaimer has `${unsafeCSS(styles)}`

markus-walther: Yes, it should. But that's another issue, I opened https://github.com/axa-ch/patterns-library/issues/2006.

PR #2013, opened 2020-10-26
Feature/fix icon performance

### LINKED ISSUE #1995: Storybook is very slow - especially in a local environment
## Current Behavior

Working locally is a pain. Reloading the page takes forever, and when opening the icon component story everything freezes
## Expected Behavior
The page should reload fast and the icons story should not freeze anymore.

## Steps to Reproduce

1. Reload page when working locally
or
1. Open icons story

## Possible Solution

Adjust how icons are displayed/loaded?

raphaellueckl:
Fixes #1995

# Merge Checklist:
- [x] Tests are sufficient.

### DIFF
--- a/src/components/00-materials/story.js
+++ b/src/components/00-materials/story.js
@@ -92,7 +92,7 @@ export const IconsAndImages = () => {
     });
 
     renderAreaIcons.innerHTML = filteredIcons
-      .map(i => mapToIconItem(i, 'materials__single-icon'))
+      .map(i => mapToIconItem(i))
       .join('');
 
     iconHeader.innerHTML =
@@ -132,13 +132,7 @@ export const IconsAndImages = () => {
       assetsToRender += 50;
 
       for (let i = iconsLoadedAlready; i < assetsToRender; ++i) {
-        iconGroup.appendChild(
-          mapToIconItemNode(
-            icons[i],
-            'image-container',
-            'materials__single-icon'
-          )
-        );
+        iconGroup.appendChild(mapToIconItemNode(icons[i], 'image-container'));
       }
       for (let i = iconsLoadedAlready / 2; i < assetsToRender / 2; ++i) {
         imageGroup.appendChild(
@@ -172,10 +166,6 @@ export const IconsAndImages = () => {
       .materials__asset-group {
         display: flex;
         flex-wrap: wrap;
-      }
-
-      .materials__single-icon {
-
       }
 
       .materials__single-image > svg {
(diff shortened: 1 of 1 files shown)


danyball (changes requested):
Pls remove `Overwrite Colors:` button - users can change that via Knobs. Other change LGTM.

01231 (changes requested):
Would suggest that icons and images start at the same level:
Screenshot 2020-10-28 at 08 39 37

Name should be fully visible on hover. Otherwise the user can't use the icon because he doesn't knows the name.
Screenshot 2020-10-28 at 08 40 54

01231 (changes requested):
Icons column is 260px wide and images column 276px. I would expect, that both columns use equal space.

Screenshot 2020-10-28 at 08 55 23

-------------------------

Screenshot 2020-10-28 at 08 55 31

src/components/00-materials/story.js
```diff
@@ -194,8 +264,15 @@ export const IconsAndImages = () => {
...
 
-      .item-name {
+      .materials__asset-name {
         margin-left: 5px;
+        margin-right: 10px;
+        width: 210px;
+      }
+
+      .materials__asset-name > p {
+        text-overflow: ellipsis;
```
danyball, line 267: The only purpose for the story is to show the name of the icons in my opinion :). I would never ellippsis them. So that developers can copy-paste the names.

We have enough space on the right. Maybe just not set a width to this row (at line 270)

src/components/00-materials/story.js
```diff
@@ -194,8 +264,15 @@ export const IconsAndImages = () => {
         margin: 20px 0;
       }
 
-      .item-name {
+      .materials__asset-name {
         margin-left: 5px;
+        margin-right: 10px;
+        width: 210px;
```
danyball: We have enough space on the right. Maybe just not set a width to this row (at line 270)

danyball (changes requested):

src/components/00-materials/story.js
raphaellueckl: But what if google decides to do this?

`super-duper-ultra-long-and-weird-svg-name-that-does-take-a-lot-of-space.svg`?

There are some super long names and I think, with autocomplete people could still find the icon when they import. The only other solution I see here is to increase the width of the text.

src/components/00-materials/story.js
raphaellueckl: Yeah, this is a good idea I think

src/components/00-materials/story.js
raphaellueckl: I made a compromise: Increased the max-width of the text significantly.

danyball (dismissed):

01231 (changes requested):
Please set "Note: ..." on a new line like it is on dev.
Is:
Screenshot 2020-10-28 at 15 14 13

Should:
Screenshot 2020-10-28 at 15 12 14

01231 (approved):
LGTM

PR #2028, opened 2020-11-10
Feature/remove datepicker width property

### LINKED ISSUE #1894: Datepicker remove width property
As discussed, it isn't necessary to keep the width property because it only sets inline-styles to the component tag.
We can remove it.

Screenshot 2020-07-22 at 13 06 44

01231:
Fixes #1894

# Merge Checklist:
- [x] Tests are sufficient.

### DIFF
--- a/src/components/20-molecules/datepicker/react/story.jsx
+++ b/src/components/20-molecules/datepicker/react/story.jsx
@@ -41,7 +41,7 @@ storiesOf('Examples/Datepicker/React', module)
     const invalid = boolean('invalid', false);
     const placeholder = text('placeholder', 'Please select a date');
     const defaultValue = text('defaultValue', '');
-    const width = text('width', '400');
+    // const inlineWidth = text('width (not an attribute)', '');
     const allowedYears = object('allowedyears', [
       '1971-2000',
       2012,
@@ -78,7 +78,6 @@ storiesOf('Examples/Datepicker/React', module)
             'axa-datepicker[data-test-id="datepicker-react"]'
           ).title = newDate;
         }}
-        width={width}
       />,
       div
     );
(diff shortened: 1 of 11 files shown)


src/components/20-molecules/datepicker/react/story.jsx
```diff
@@ -41,14 +41,18 @@ storiesOf('Examples/Datepicker/React', module)
     const invalid = boolean('invalid', false);
     const placeholder = text('placeholder', 'Please select a date');
     const defaultValue = text('defaultValue', '');
-    const width = text('width', '400');
+    const inlineWidth = text('width (not an attribute)', '');
```
raphaellueckl, line 44: What is this? Do we really need that?

src/utils/with-react.js
```diff
@@ -142,7 +145,19 @@ const distributeProperties = (properties, componentClass) => {
 
     // map property name to value *unless* value is undefined
     if (value !== undefined) {
-      map[name] = value;
+      if (type === 'style') {
+        // {width: "500px"} -> width:500px;
```
raphaellueckl: When I see stuff like that, I usually think an extra method, that explains what it does with its name, makes the code more readable.

src/components/20-molecules/datepicker/react/story.jsx
raphaellueckl: IMHO we should not have that as a prop. This is just general styling and we would somehow have to add it for all other elements too.

raphaellueckl (changes requested):

src/components/20-molecules/datepicker/react/story.jsx
01231: It's nice to have for demo purposes in my opinion. And it's a real use case. We expect the user to use this feature in the datepicker.

01231:
seems like it doesn't work on ie

src/components/20-molecules/datepicker/react/story.jsx
raphaellueckl: A bit weird to me, but up to you.

src/utils/with-react.js
01231: done

raphaellueckl (approved):

PR #2033, opened 2020-11-18
Bugfix/add missing package locks

### LINKED ISSUE #1952: package-lock missing
Check why not all our components have a package-lock.json.

01231:
Fixes #1952

# Merge Checklist:
- [x] Tests are sufficient.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "build-components": "lerna run build --parallel",
     "watch-components": "lerna run watch --parallel",
     "setup-npm-links": "node scripts/setup-npm-link.js",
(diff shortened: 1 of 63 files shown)


danyball (dismissed):

NoahLiechti:
My conclusion would be to remove all `package-lock.json` from submodules. It is a know issue with hoisting which is probably intended. Due the fact that we don't have any impact I think this is reasonable.

NoahLiechti:
According to package-lock.json description at npmjs.com

> One key detail about package-lock.json is that it cannot be published, and it will be ignored if found in any place other than the toplevel package

raphaellueckl (approved):

PR #2048, opened 2020-11-26
Feature/introduce playwright

raphaellueckl:
Introduces the playwright testing framework to the pattern library.

# Merge Checklist:
- [x] Tests are sufficient.


package.json
```diff
@@ -14,15 +14,15 @@
...
-    "test-chrome": "sh scripts/testcafe-chrome.sh",
-    "test-ie": "sh scripts/testcafe-ie.sh",
     "test-jest": "jest --env=jsdom",
+    "test-playwright": "TEST_HOST_STORYBOOK_URL=http://localhost:6006 jest --config=jest.ui.config.js",
     "test-jest-coverage": "jest --env=jsdom --coverage",
     "test-jest-watch": "jest --env=jsdom --watch -o",
-    "test-ui": "TEST_HOST_STORYBOOK_URL=http://localhost:6006 testcafe chrome ./**/ui.test.js",
-    "test-ui-hl": "TEST_HOST_STORYBOOK_URL=http://localhost:6006 testcafe 'chrome:headless' ./**/ui.test.js",
+    "test-ui": "sh scripts/test-ui-testrunner.sh",
+    "test-ui-isolated": "TEST_HOST_STORYBOOK_URL=http://localhost:6006 testcafe chrome ./**/ui.test.js",
```
danyball: isolated? Would rename that to test-ui-testcafe

package.json
```diff
@@ -14,15 +14,15 @@
...
+    "test-ui-hl-isolated": "TEST_HOST_STORYBOOK_URL=http://localhost:6006 testcafe 'chrome:headless' ./**/ui.test.js",
```
danyball: same

scripts/test-playwright.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# npm run test-playwright --config="jest.ui.config.js"
+
+npx jest --config=jest.ui.config.js
+test_result=$?
+
+exit $test_result
```
danyball: Is it possible to reduce complexity and remove this sh script?

scripts/test-ui-testrunner.sh
```diff
@@ -8,13 +7,16 @@ export $(egrep -v '^#' .env | xargs) > /dev/null 2>&1
 npx start-storybook -p $TEST_HOST_STORYBOOK_PORT -c .storybook -s ./src/static --ci --quiet > /dev/null 2>&1 &
 npx wait-on $TEST_HOST_STORYBOOK_URL -t 30000
 
-if [ -z "$1" ]
-then
-    npx testcafe -c ${NUM_CPUS} "chrome:headless" -q ./**/ui.test.js
-else
-    npx testcafe -F "${1}" -c ${NUM_CPUS} "chrome:headless" -q ./**/ui.test.js
+sh ./scripts/test-playwright.sh
```
danyball: `npx jest --config=jest.ui.config.js` here?

src/components/10-atoms/button/pw.ui.test.js
```diff
@@ -0,0 +1,36 @@
...
+
+let browser;
+let page;
+
+const config = {
+  // devtools: true
+};
+
+beforeAll(async () => {
+  browser = await chromium.launch(config);
```
danyball: Is it possible to launch browser only once? So that all the test-files do not need to start chromium on their own. A test should be "dump" and should just say "browser start" instead of "chromium start".

package.json
raphaellueckl: Yeah good point. I didn't really know what the best name would be. Those tasks won't run as long as storybook is not started separately.

I would go with "test-testcafe", ok? Or `test-testcafe-serverless` (and rename the playwrtight one too)?

scripts/test-playwright.sh
raphaellueckl: Would you prefer both scripts inlined? Would maybe be worth a thought.

scripts/test-ui-testrunner.sh
raphaellueckl: imho only if we also inline the testcafe script.

src/components/10-atoms/button/pw.ui.test.js
raphaellueckl: I could add a class that abstracts this method, but imho this introduces more complexity that is being solved.

raphaellueckl: Maybe this makes sense, when we start having shared test-configs and so on. I will give it a try.

01231 (commented):
Would be useful to add playwright to the `contribution.md`.

```
# Testing

- Library uses [TestCafe](https://devexpress.github.io/testcafe/) for ui testing
- Library uses [Jest](https://jestjs.io/) for unit testing
- ...
```

package.json
danyball: "test-testcafe", ok

scripts/test-playwright.sh
danyball: Dont like these shellscripts. Isnt it possible to just start "npx jest..." instead of the sh file?

raphaellueckl: Yeah this is what I mean. To be consistent, we would need to do the same for the testcafe-script, not just playwright. What do you think?

scripts/ui-test-runner.sh
```diff
@@ -1,19 +1,36 @@
 #!/bin/bash
-NUM_CPUS=$(python -c 'import multiprocessing as mp; print(mp.cpu_count())')
 
 # Import environment variables from .env file
 export $(egrep -v '^#' .env.defaults | xargs) > /dev/null 2>&1
 export $(egrep -v '^#' .env | xargs) > /dev/null 2>&1
 
+echo "Running Storybook on address '$TEST_HOST_STORYBOOK_URL' and on port '$TEST_HOST_STORYBOOK_PORT'"
```
danyball, line 7: helpful! nice

danyball (approved):

01231 (approved):
LGTM

PR #2051, opened 2020-11-30
Fix component versioning bug

### LINKED ISSUE #2050: Dependent component version rewriting fails for components that extend shorter same-named components
Suppose we have a component axa-button-link in version 6.3.4 and an axa-button in version 5.0.9.

If another component uses them to define its rendered HTML, dependent component version rewriting will change bare HTML tags into their versioned cousins. 

However, currently this leads to wrong outcomes  with components that extend shorter same-named components, like axa-button-link extending its proper prefix axa-button (cf. rewriting step =2> below):

`<axa-button-link =1> <axa-button-link-6-3-4 =2> <axa-button-5-0-9-link-6-3-4`

Fix this problem in a principled manner.

markus-walther:
Fixes #2050.

# Merge Checklist:
- [x] Tests are sufficient.

### DIFF
--- a/src/utils/component-versioning.js
+++ b/src/utils/component-versioning.js
@@ -1,6 +1,10 @@
 import { html } from 'lit-element';
 import defineOnce from './define-once';
 
+// constants
+const RESERVED_CHARACTER = '{'; // not allowed in HTML tags or their attributes ...
+// and also not part of static template strings due to ${...} syntax!
+
 // helper functions
 const toKebabCase = dottedVersionString =>
   `${dottedVersionString}`.replace(/\./g, '-').replace(/[^A-Za-z0-9-]/g, '');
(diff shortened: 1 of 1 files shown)


src/utils/component-versioning.js
```diff
@@ -9,24 +13,39 @@ const versionedTag = (tagName, version) => `${tagName}-${toKebabCase(version)}`;
...
+    // ... by comparing with master-component tag name
+    .filter(name => name !== tagName)
+    // ... and sorting longest-first (e.g. axa-button-link comes before axa-button)
+    .sort((a, b) => b.length - a.length);
   // pair the list of dependency tag names with their versions
   return [dependencies, dependencies.map(_tagName => versions[_tagName])];
 };
 
-const oldTag = (tagName, closing = '') => `<${closing}${tagName}`;
+const oldTag = (tagName, closing = '', openingBracket = '<') =>
```
raphaellueckl, line 26: `openingBracket` is a useless parameter, it's never being used outside this method.

src/utils/component-versioning.js
```diff
@@ -101,11 +120,16 @@ const versionedHtml = componentInstance => (strings, ...args) => {
...
   // rewrite incoming array of static parts of template literals
-  // in such a way that tags of dependent components are versioned
+  // in such a way that tags of dependent components are versioned:
   let newStrings = strings;
+  // 1. rewriting proper, turning tags into versioned ones with funny initial brackets (see newTag(...) above)
   for (let i = 0, n = tagNames.length; i < n; i++) {
     newStrings = rewrite(newStrings, tagNames[i], versions[i]);
   }
+  // 2. finish rewriting by converting funny initial brackets back to standard ones
+  for (let i = 0, n = newStrings.length; i < n; i++) {
```
raphaellueckl, line 130: Why 2 variables? I am sure you have a reason for that (maybe infinite loop and outofmemory at some point, because array is modified within the body, which COULD BE a potential risk). :)

This is simpler in my opinion:
``` for (let i = 0; i < newStrings.length; i++) {```

File: src/utils/component-versioning.js (hunk 1 above)
Comment (markus-walther) [reply src/utils/component-versioning.js:26]: It is used in line 40, and there are instances using it implicitly (in its default) in line 50 and 55.

File: src/utils/component-versioning.js (hunk 2 above)
Comment (markus-walther) [reply src/utils/component-versioning.js:130]: This is a generic pattern for JS performance: cache the results of a computation that is not changed in a loop.

taking the length of an array is such a computation.

This idiom is used in all of my code... It creates 'standard' for-loops that are easy to check for correctness.

File: src/utils/component-versioning.js [hunk 3]
```diff
@@ -9,24 +13,39 @@ const versionedTag = (tagName, version) => `${tagName}-${toKebabCase(version)}`;
...
 const extractDependencies = componentClass => {
   const { versions, tagName } = componentClass;
-  // extract all dependencies by comparing with master-component tag name
-  const dependencies = Object.keys(versions).filter(name => name !== tagName);
+  // extract all dependencies...
+  const dependencies = Object.keys(versions)
+    // ... by comparing with master-component tag name
+    .filter(name => name !== tagName)
+    // ... and sorting longest-first (e.g. axa-button-link comes before axa-button)
+    .sort((a, b) => b.length - a.length);
```
danyball, line 21: Is sorting safe enough?  Or is this just one more step to be more safe beside the "{"- replace below?

raphaellueckl: This is actually necessary, as you need to start with the longest names, to really make sure you do not process long names with shorter ones.

Example:
`<axa-button` HAS TO BE processed after `<axa-button-link` with this approach (otherwise the wrong component will be versioned) and this sorting makes sure of that.

raphaellueckl:
This doesn't seems to work 100%:
Untitled

src/utils/component-versioning.js
markus-walther: Raphael is right here.

raphaellueckl (approved):

PR #2063, opened 2020-12-07
update materials readme with typescript example

### LINKED ISSUE #2054: Add "typings" to svg files
SVG's should be shipped with a typescript "typings" (not actually, but outplay the system). Could also be made with real types for each icon, generated by a script.

File:
```
declare module '*.svg' {
    const ReactComponent: React.FC<React.SVGProps<SVGSVGElement>>;  
    export default ReactComponent; 
}
```

Enables easy SVG imports in react:
`import ThumbDownSvg from '../../assets/thumb_down.svg';`

raphaellueckl:
Fixes #2054

Typings are not needed, but with the import statement that is recommended by the [Create React Team](https://create-react-app.dev/docs/adding-images-fonts-and-files/#adding-svgs) and also seem to use the same methods we already suggest with SVGR.

# Merge Checklist:
- [x] Tests are sufficient.

### DIFF
--- a/src/components/00-materials/README.md
+++ b/src/components/00-materials/README.md
@@ -12,17 +12,25 @@ npm install @axa-ch/materials
 
 ```js
 import { svg } from 'lit-html';
-import { ArrowRightSvg } from '@axa-ch/materials/icons';
+import { Arrow_forwardSvg } from '@axa-ch/materials/icons';
 
-<span>${svg(ArrowRightSvg)}</span>;
+<span>${svg(Arrow_forwardSvg)}</span>;
 ```
 
 ### React
 
 ```js
-import ArrowRightSvg from '@axa-ch/materials/icons-raw/arrow-right.svg';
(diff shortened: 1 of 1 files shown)


src/components/00-materials/README.md
```diff
@@ -12,17 +12,25 @@ npm install @axa-ch/materials
 
 ```js
 import { svg } from 'lit-html';
-import { ArrowRightSvg } from '@axa-ch/materials/icons';
+import { Arrow_forwardSvg } from '@axa-ch/materials/icons';
```
01231, line 15: Icons like `backspace.svg` gets exported as `BackspaceSvg`. 
We should consider exporting icons as `ArrowForwardSvg` (PascalCase) instead of `Arrow_forwardSvg`. This is just very inconsistent and not intuitiv.

01231 (approved):

src/components/00-materials/README.md
raphaellueckl: I agree.

PR #2093, opened 2021-02-08
Feature/npmci

danyball:
Fixes: faster and harder build. We could discuss if we also remove `rm -rf node_modules` because `ci` is doing that like the docu says. But i think rm is faster.

> In short, the main differences between using npm install and npm ci are:
> 
> The project must have an existing package-lock.json or npm-shrinkwrap.json.
> If dependencies in the package lock do not match those in package.json, npm ci will exit with an error, instead of updating the package lock.
> npm ci can only install entire projects at a time: individual dependencies cannot be added with this command.
> If a node_modules is already present, it will be automatically removed before npm ci begins its install.
> It will never write to package.json or any of the package-locks: installs are essentially frozen.

PS:

If that PR is approved i will install that also on midgard.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "watch-components": "lerna run watch --parallel",
     "setup-npm-links": "node scripts/setup-npm-link.js",
     "postinstall": "lerna bootstrap --hoist && cross-env NODE_ENV=production npm run build-components && lerna run --parallel genPackagelock",
-    "release": "npm whoami && npx lerna clean && rm -rf node_modules && npm ci && npm run test && lerna version --exact && cross-env NODE_ENV=production npm run build-components && lerna publish from-package",
+    "release": "npm whoami && npx lerna clean && npm ci && npm run test && lerna version --exact && cross-env NODE_ENV=production npm run build-components && lerna publish from-package",
     "start": "npm run storybook",
     "storybook": "start-storybook -p 6006 -s src/static",
     "build-storybook": "build-storybook -s ./src/static",
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "watch-components": "lerna run watch --parallel",
     "setup-npm-links": "node scripts/setup-npm-link.js",
     "postinstall": "lerna bootstrap --hoist && cross-env NODE_ENV=production npm run build-components && lerna run --parallel genPackagelock",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -10,7 +10,7 @@
     "watch-components": "lerna run watch --parallel",
     "setup-npm-links": "node scripts/setup-npm-link.js",
     "postinstall": "lerna bootstrap --hoist && cross-env NODE_ENV=production npm run build-components && lerna run --parallel genPackagelock",
-    "release": "npm whoami && npx lerna clean && rm -rf node_modules && npm install && npm run test && lerna version --exact && cross-env NODE_ENV=production npm run build-components && lerna publish from-package",
+    "release": "npm whoami && npx lerna clean && rm -rf node_modules && npm ci && npm run test && lerna version --exact && cross-env NODE_ENV=production npm run build-components && lerna publish from-package",
```
raphaellueckl: As long as the `rm -rf` command stays there, i don't see the point in running npm ci. Because the whole purpose seems to be that it automatically does that for you.

And what I'm not sure about: It ALWAYS expects a package-lock file. Is that a good idea? What if you want to update all minor versions at once? By thinking about it, that is probably not a big concern, as npm install can still be run manually.

I would approve if you remove the `rm -rf node_modules`

raphaellueckl (changes requested):

package.json
markus-walther: I had to read up on things here: https://medium.com/better-programming/npm-ci-vs-npm-install-which-should-you-use-in-your-node-js-projects-51e07cb71e26 

@danyball A brief explainer would have been helpful.

"
- `npm ci` will delete your node_modules folder to ensure a clean state.
- It will look in your package-lock.json to install all the dependencies with the exact version.
- Unlike npm install, `npm ci` will never modify your package-lock.json. It does however expect a package-lock.json file in your project — if you do not have this file, npm ci will not work and you have to use npm install instead.

If you use npm ci, you’ll get reliable builds."

Convincing to me, and since it is restricted to `release`, OK. We can still update package-lock.json separately, using `npm i`.

markus-walther (dismissed):
LGTM, see feedback comment.

package.json
danyball: @markus-walther i have given exact your explainer at my description? 

Jep, will remove that.

raphaellueckl (approved):

PR #2103, opened 2021-03-10
[WIP] axa-carousel extensions

### LINKED ISSUE #2102: axa-carousel extensions
For substituting the existing carousel solution on axa.ch, a few missing features need to be implemented:

- **arrow colours** must be configurable (now: hardwired to white)
- at a viewport width &gt;= 980px, **all** the child elements should be visible, and presented in a in horizontal layout that fills the viewport width (configurable, default: off). In that mode, all carousel navigation should be disabled.

markus-walther:
Fixes #2102.

# Merge Checklist:
- [ ] Code selfreview has been performed ([Best Practices](https://github.com/axa-ch/patterns-library/blob/develop/CONTRIBUTION.md#best-practices)).
- [ ] Tests are sufficient.
- [ ] Modifications still work in IE.
- [ ] Dependencies to other components still work.


src/components/10-atoms/carousel/README.md
```diff
@@ -1,21 +1,34 @@
...
 
 ### keysenabled
 
-This boolean attribute specifies if the keyboard navigation (left/right keys) should be enabled.
+This Boolean attribute specifies whether keyboard navigation via left/right arrow keys should be enabled (default: false).
+
+### arrowstyle
+
+This string-valued attribute adds custom inline styling to the arrow buttons of the carousel
+(e.g. `border-color:grey` for grey arrows, or `display:none` to suppress arrows).
```
danyball: The user can do anything with css now. Isnt it too risky that we get carousels that have "crazy" arrows?

src/components/10-atoms/carousel/child.scss
```diff
@@ -15,3 +15,25 @@ axa-carousel {
...
+axa-carousel[visible="0"] > *:first-child,
+axa-carousel[visible="1"] > *:first-child + *,
+axa-carousel[visible="2"] > *:first-child + * + *,
+axa-carousel[visible="3"] > *:first-child + * + * + *,
+axa-carousel[visible="4"] > *:first-child + * + * + * + *,
+axa-carousel[visible="5"] > *:first-child + * + * + * + * + *,
+axa-carousel[visible="6"] > *:first-child + * + * + * + * + * + *,
+axa-carousel[visible="7"] > *:first-child + * + * + * + * + * + * + *,
+axa-carousel[visible="8"] > *:first-child + * + * + * + * + * + * + * + *,
+axa-carousel[visible="9"] > *:first-child + * + * + * + * + * + * + * + * + * {
```
danyball, line 28: Oh Tannenbaum, oh Tannenbaum,.... :-D

danyball: Thats ok for a quick solution. But to hard-limit it to 10 because of these css rules feels wrong. Is there no solution to be dynamic? On the other hand.... it is the carousel. The ugliest Webelement. So why not ;)

src/components/10-atoms/carousel/index.js
```diff
@@ -45,7 +50,7 @@ class AXACarousel extends InlineStyles {
     // Internal
     this.autoRotateTimerID = null;
     this.slides = null;
-    this.visibleSlideIndex = 0;
+    this.visible = 0;
```
danyball, line 57: Why renamed it?

src/components/10-atoms/carousel/index.js
```diff
@@ -109,57 +114,41 @@ class AXACarousel extends InlineStyles {
...
-
-    this._setSlideVisibleWithAnimation(nextSlideIndex, ANIMATION_RIGHT_CLASS);
+  _nextSlide(increment = 1) {
+    let numSlides = this.slides.length;
+    // wrap around after last slide...
+    let nextSlideIndex = (this.visible + increment) % numSlides;
+    this._setSlideVisibleWithAnimation(
+      nextSlideIndex < 0 ? numSlides - 1 : nextSlideIndex, // ... and before first slide
+      increment > 0 ? ANIMATION_RIGHT_CLASS : ANIMATION_LEFT_CLASS
+    );
```
danyball, line 132: For me the old solution is better readable. There are just simple mathematical operations like ">" and less ifs.

src/components/10-atoms/carousel/index.js
```diff
@@ -109,57 +114,41 @@ class AXACarousel extends InlineStyles {
...
-
-      if (node.clientHeight > this._carouselMinHeight) {
-        this._carouselMinHeight = node.clientHeight;
+    (async () => {
+      const savedVisible = this.visible;
+      // we need to set carousel min height in case there are elements with different heights.
+      for (let nodes = this.slides, i = 0, n = nodes.length, node; i < n; i++) {
+        this.visible = i;
+        node = nodes[i];
+        await sleep(0); // force event-loop
```
danyball, line 149: The old solution do not need a setTimeout.

src/other/pages/showcases/axa-ch-main/carousel.js
```diff
@@ -0,0 +1,231 @@
...
+      </axa-picture>
+      <div class="m-related-content__content">
+        <p class="m-related-content__caption">
+          Grund- und Zusatzversicherung: Hilft separat versichern beim Sparen?
+        </p>
+        <h3 class="m-related-content__text">Gesundheit</h3>
+      </div>
+    </a>
+  </axa-carousel>
+`;
```
danyball, line 316: Not checked that file in detail

danyball (changes requested):

src/components/10-atoms/carousel/README.md
raphaellueckl: I also don't like that we hand down CSS properties as HTML properties. This is a bit sandboxy and allows for too much variants. Button, Button-Link, Link, Hero Banner, ... almost everywhere we have a set of colors or "modes" that we allow.

src/components/10-atoms/carousel/README.md
```diff
@@ -1,21 +1,34 @@
...
+
+This string-valued attribute forces the carousel into a specific mode of behaviour (default: no specific mode of behaviour).
+
+Currently, one mode value is recognized: `all` switches to horizontal layout of child elements for viewports &gt;= 992px.
+In that layout, _all_ child elements are visible, they fill the viewport width seamlessly, and carousel navigation is suppressed.
 
 ## Child Elements
 
-Do not set more than ~100 child elements. Otherwise, because of height-calculation costs performance will suffer.
+There is an absolute limit of 10 child elements.
```
raphaellueckl, line 40: Only 10? Why is this needed? As long as we don't have to "show all at once", there should not be a strict limit, just a recommendation, no?

This would be a massive breaking change, if we limit our users like that. They might have to add additional logic just to work with this limitation.

src/components/10-atoms/carousel/child.scss
```diff
@@ -1,6 +1,6 @@
 @mixin defaultAttributesChildElements {
   @include text(t4);
-  display: flex;
+  display: none !important;
```
raphaellueckl, line 3: If that is needed, ok. If you just forgot it, please remove.

I'm writing this comment not because I do not trust you, but because this happened to me. :D

src/components/10-atoms/carousel/child.scss
raphaellueckl: I agree with danyball. I'd rather prefer a dynamic solution to generate this CSS, than limit it to 10 just because we don't want an ugly stylesheet.

src/components/10-atoms/carousel/index.js
```diff
@@ -33,8 +35,11 @@ class AXACarousel extends InlineStyles {
       autorotatedisabled: { type: Boolean },
       autorotatetime: { type: Number, defaultValue: 5000 },
       keysenabled: { type: Boolean },
+      visible: { type: Number, reflect: true },
+      arrowstyle: { type: String, reflect: true },
+      mode: { type: String, reflect: true },
       _animationWrapperClass: { type: String },
-      _carouselMinHeight: { type: Number },
+      _carouselMinHeight: { type: Number, defaultValue: 0 },
```
raphaellueckl, line 46: Why an "internal prop" and not just a class property in the constructor?

src/components/10-atoms/carousel/index.js
```diff
@@ -109,57 +114,41 @@ class AXACarousel extends InlineStyles {
...
-
-    if (nextSlideIndex > this.slides.length - 1) {
-      nextSlideIndex = 0;
-    }
-
-    this._setSlideVisibleWithAnimation(nextSlideIndex, ANIMATION_RIGHT_CLASS);
+  _nextSlide(increment = 1) {
+    let numSlides = this.slides.length;
+    // wrap around after last slide...
+    let nextSlideIndex = (this.visible + increment) % numSlides;
```
raphaellueckl: If `numSlides` is 0, `nextSlideIndex` will be `NaN`.

src/components/10-atoms/carousel/index.js
```diff
@@ -109,57 +114,41 @@ class AXACarousel extends InlineStyles {
...
+    this._setSlideVisibleWithAnimation(
+      nextSlideIndex < 0 ? numSlides - 1 : nextSlideIndex, // ... and before first slide
```
raphaellueckl, line 130: `nextSlideIndex` cannot be smaller than 0, no?

src/components/10-atoms/carousel/index.js
```diff
@@ -229,18 +224,18 @@ class AXACarousel extends InlineStyles {
   _handleKeyUp = ev => {
     const e = ev || window.event;
 
-    if (e.keyCode === 37) {
+    if (e.keyCode === ARROW_LEFT) {
       this.handlePreviousButtonClick();
-    } else if (e.keyCode === 39) {
+    } else if (e.keyCode === ARROW_RIGHT) {
```
raphaellueckl, line 234: Nice! No magic numbers.

src/components/10-atoms/carousel/index.js
markus-walther: As you can see in the diff, this was not introduced by me, so ask my predecessor ;-)

src/components/10-atoms/carousel/README.md
markus-walther: Our users will like it _because_ it gives them flexibility. This flexibility is indeed necessary, to both allow backwards compatibility and new uses, see the mocked AXA main page in the demos.

To provide this with only one tiny change is arguably elegant, so I would defend it here.

src/components/10-atoms/carousel/child.scss
markus-walther: There is no unlimited dynamic solution because CSS does not have variables in selectors.

The change to controlling visibility via a single numeric attribute is well-motivated, because the mode=all functionality could not be realized efficiently with the old clumsy way of controlling slide visibility.

src/components/10-atoms/carousel/index.js
markus-walther: Easy - because it is now externally visible as an attribute, unlike how it was before, and those tend to be short and to the point!

src/components/10-atoms/carousel/index.js
danyball: Only props are watched by lit-element and are updated automatically the template. I think that was the reason.

PR #2119, opened 2021-03-25
introducing a spinner component

### LINKED ISSUE #2066: Add Spinner component
The spinner is being used everywhere and was also being discussed a longer time ago, but faded out due to other priorities.

This should be reevaluated with the designers. Especially the ones that rely on "position: fixed".

marcwelz:
Fixes #2066

# Merge Checklist:
- [x] Tests are sufficient.


src/components/10-atoms/spinner/README.md
```diff
@@ -0,0 +1,30 @@
+# Spinner
+
+Adds a loading spinner
```
raphaellueckl: Well, this does not need an explanation. The explanation should rather be, what that component is and what it can be used for.

src/components/10-atoms/spinner/README.md
```diff
@@ -0,0 +1,30 @@
...
+| `size=default`              | Size of the spinner  |
+| `color=inverted-blue-ocean` | Color of the spinner |
+
+### Size
+
+Declaires the size of the spinner.
+
+### Color
+
+Sets the color of the spinner. If `color` is set to `default`, it has the color ocean blue.
```
raphaellueckl: If this is true, this should be refactored. Default-colors should not be named and `color="default"` does not make sense, since it is default. If you do not add the `color` attribute, you expect the default to happen. You don't expect, that you have to define attributes just to set them to defaults.

src/components/10-atoms/spinner/index.js
```diff
@@ -0,0 +1,59 @@
...
+    return {
+      size: { type: String, defaultValue: 'default' },
+      color: { type: String, defaultValue: 'inverted-blue-ocean' },
+    };
+  }
+
+  render() {
+    const { size = '', color = '' } = this;
+    const classes = {
+      'a-spinner--default': true,
```
raphaellueckl: Please remove this class, as it is not needed.

src/components/10-atoms/spinner/index.js
```diff
@@ -0,0 +1,59 @@
...
+      'a-spinner--small': size === 'small',
+      'a-spinner--inverted-blue-ocean': true,
```
raphaellueckl: Please remove this class, as it is not needed. If the variable is not dynamic anyways (as it is always `true`), there is no need to declare it here. Better put it in the template. But you should remove the class completely. We don't need it.

src/components/10-atoms/spinner/index.js
```diff
@@ -0,0 +1,59 @@
...
+          ></span>
+          <span
+            class="${classMap(classes)} a-spinner__dot a-spinner__dot-3"
+          ></span>
+        </div>
+      </article>
+    `;
+  }
+
+  disconnectedCallback() {
```
raphaellueckl: This method can be deleted.

src/components/10-atoms/spinner/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@axa-ch/spinner",
+  "version": "0.0.0-beta.0",
+  "description": "The spinner component for the AXA Pattern Library",
+  "author": "Pattern Warriors",
+  "homepage": "https://github.com/axa-ch/patterns-library/tree/develop/src/components/10-atoms/spinner#readme",
+  "license": "Copyright 2019 AXA Versicherungen AG",
```
raphaellueckl: Hm... Maybe we should remove the year here? I know, that this will be the same problem in all components.

src/components/10-atoms/spinner/pw.ui.test.js
```diff
@@ -0,0 +1,17 @@
...
+    );
+    await page.waitForSelector(tag);
+
+    expect(
+      (await page.isVisible('.a-spinner__dot-1')) &&
+        (await page.isVisible('.a-spinner__dot-2')) &&
+        (await page.isVisible('.a-spinner__dot-3'))
+    ).toBeTruthy();
+  });
+});
```
raphaellueckl, line 29: You could add a test that validates the following:

- Open spinner, check if it is displayed
- Close the spinner, check that is is not displayed
- Open spinner, check if it is displayed

And maybe another test that validates the colors of the dots.

src/components/10-atoms/spinner/story.js
```diff
@@ -0,0 +1,47 @@
...
+export const Spinner = () => {
+  const sizes = radios('size', sizeOptions, 'default');
+  const colors = radios('color', invertedBgs, 'inverted-blue-ocean');
+
+  const wrapper = document.createElement('div');
+
+  const template = html`
+    <style>
+      body {
+        background-color: ${colors === 'inverted-white' ? '#ccc' : '#fff'};
```
raphaellueckl, line 34: nice

src/components/10-atoms/spinner/unit.test.js
```diff
@@ -0,0 +1,13 @@
...
+  test('firstUpdated() should not call another method', () => {
+    AXASpinner.prototype.methodThatShouldNotBeCalled = jest.fn();
+
+    AXASpinner.prototype.firstUpdated();
+
+    expect(
+      AXASpinner.prototype.methodThatShouldNotBeCalled
+    ).not.toHaveBeenCalled();
+  });
+});
```
raphaellueckl: Remove this file completely. This test is only wasting time.

src/components/30-organisms/modal/index.scss
```diff
@@ -21,7 +21,7 @@
     flex-flow: column;
   }
 
-  &__upper-close-container {
+  &__upper-close-container {
```
raphaellueckl: What happened here?

src/components/10-atoms/spinner/pw.ui.test.js
marcwelz: There is currently no option to turn on/off the spinner so it would make less sense to implement a test for it.

src/components/10-atoms/spinner/index.js
Milchmaa: I'm not 100% sure what you mean but we need these classes

src/components/30-organisms/modal/index.scss
marcwelz: It seems that I made a mistake copying code from it... Is there a way to undo this commit?

src/components/10-atoms/spinner/README.md
```diff
@@ -0,0 +1,30 @@
...
+```
+
+## Properties
+
+### Variant
+
+| Attribute                   | Details              |
+| --------------------------- | -------------------- |
+| `size=false`                | Size of the spinner  |
+| `color=inverted-blue-ocean` | Color of the spinner |
```
raphaellueckl: This does no longer exist right? Because it is the default

src/components/10-atoms/spinner/README.md
```diff
@@ -0,0 +1,30 @@
...
+<axa-spinner></axa-spinner>
+```
+
+## Properties
+
+### Variant
+
+| Attribute                   | Details              |
+| --------------------------- | -------------------- |
+| `size=false`                | Size of the spinner  |
```
raphaellueckl: This is a boolean attribute, so we should just write `size` and explain in the details, what it's doing. Anyways, since you made a boolean out of it, you should rename it to `small` in the code base. So that the variable tells you what it's going to do. `Size` does not tell you that it is a boolean. When I read that I expect that I could set some numbers.

src/components/10-atoms/spinner/index.js
```diff
@@ -0,0 +1,53 @@
...
+
+  static get styles() {
+    return css`
+      ${unsafeCSS(styles)}
+    `;
+  }
+
+  static get properties() {
+    return {
+      size: { type: Boolean, defaultValue: false },
```
raphaellueckl: Like I wrote already: Size = false... what does that mean? No Size? So rename it to `small`.

src/components/10-atoms/spinner/index.js
```diff
@@ -0,0 +1,53 @@
...
+      color: { type: String, defaultValue: 'inverted-blue-ocean' },
+    };
+  }
+
+  render() {
+    const { size = '', color = '' } = this;
```
raphaellueckl: size (oder `small`) is a boolean, so you should add `false` as default value to be consistent with types.

src/components/10-atoms/spinner/index.js
```diff
@@ -0,0 +1,53 @@
...
+          <span
+            class="a-spinner__dot a-spinner__dot-3 ${classMap(classes)}"
+          ></span>
+        </div>
+      </article>
+    `;
+  }
+}
+
+/* eslint-disable no-undef */
```
raphaellueckl: This line can be removed. If that yields an error for you, you can rebase with develop. I fixed that recently.

src/components/10-atoms/spinner/index.scss
```diff
@@ -0,0 +1,56 @@
+.a-spinner {
+  display: block;
```
raphaellueckl: This can be removed, as it is the default anyways for  `article`.

src/components/10-atoms/spinner/story.js
```diff
@@ -0,0 +1,41 @@
...
+  parameters: {
+    readme,
+    usage: {
+      componentName: 'spinner',
+    },
+    changelog,
+  },
+};
+
+const invertedBgs = {
```
raphaellueckl: What does that mean? We should only use abbrevations if they are "common". So better write it out.

src/components/10-atoms/spinner/README.md
```diff
@@ -0,0 +1,30 @@
...
+<axa-spinner></axa-spinner>
+```
+
+## Properties
+
+### Variant
+
+| Attribute     | Details                   |
+| ------------- | ------------------------- |
+| `small=false` | Sets the spinner to small |
```
raphaellueckl: Just call it `small` (without the false) and write the default value in the details section.

raphaellueckl: Also mention that this is a boolean attribute.

src/components/10-atoms/spinner/README.md
```diff
@@ -0,0 +1,30 @@
...
+| `color`       | Sets the color            |
+
+### small
+
+If `small` is set to true, it has the small size.
+
+### Color
+
+Sets the color of the spinner. If `color` is not set, it has the color ocean blue.
```
raphaellueckl, line 30: Probably mention the different possible variants.

src/components/10-atoms/spinner/index.react.d.ts
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+export interface AXASpinnerProps {
+  className?: string;
+  size?: string;
```
raphaellueckl: This has to be renamed to `small` too.

src/components/10-atoms/spinner/README.md
```diff
@@ -28,3 +28,4 @@ If `small` is set to true, it has the small size.
 ### Color
 
 Sets the color of the spinner. If `color` is not set, it has the color ocean blue.
+The other color options are: dark-grey and white.
```
raphaellueckl: "The other color options are: `dark-grey` and `white`."

src/components/10-atoms/spinner/README.md
```diff
@@ -16,10 +16,10 @@ import '@axa-ch/spinner';
 
 ### Variant
 
-| Attribute     | Details                   |
-| ------------- | ------------------------- |
-| `small=false` | Sets the spinner to small |
-| `color`       | Sets the color            |
+| Attribute | Details                                                          |
+| --------- | ---------------------------------------------------------------- |
+| `small`   | Boolean. Sets the spinner to small, the default value is `false` |
```
raphaellueckl: "Smaller version of the spinner (boolean, default: false)"

src/components/10-atoms/spinner/index.react.d.ts
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+export interface AXASpinnerProps {
+  className?: string;
+  size?: Boolean;
```
raphaellueckl: This is still `size`

raphaellueckl (approved):

PR #2133, opened 2021-04-20
add 3 style sizes from figma

### LINKED ISSUE #2110: 3 standard sizes for icons?
From Design Gilde:
Discuss how to set 3 standard sizes for icons/svgs

image

danyball:
Fixes #2110 

# Merge Checklist:
- [x] Tests are sufficient.


src/components/10-atoms/icon/ui.test.js
```diff
@@ -42,9 +42,26 @@ test('should not render "undefined" string', async t => {
...
+);
+test('should set size to original svg dimensions', async t => {
+  await t.expect(svgImageItem.getStyleProperty('width')).eql('100px');
+  await t.expect(svgImageItem.getStyleProperty('height')).eql('100px');
```
raphaellueckl: Hm we once decided to only write playwright tests and not use or adapt testcafe tests. :/

danyball: Ups :(

src/components/00-materials/README.md
```diff
@@ -99,6 +99,10 @@ body {
 
 ## Icon/Image set
 
+### Size classes
```
raphaellueckl: I would prefer "Sizes" over "Size classes". I know what you mean but I think it is confusing.

src/components/10-atoms/icon/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 6.0.0
+
+- BREAKING CHANGE: The new default for `size` of icons loaded by URL or SVG string is now 24x24 px instead of `auto`.
+- Functional icons should have only 3 sizes. You can now set these dimensions via `size` property. #2110
+- We renamed the size value `auto` to `original`.
```
raphaellueckl: This is also a breaking change imho.

src/components/10-atoms/icon/index.d.ts
```diff
@@ -1,23 +1,3 @@
...
-  | 'expand'
-  | 'mobile'
-  | 'phone'
-  | 'search'
-  | 'upload'
-  | 'cloud-upload'
-  | 'axa-logo'
-  | 'axa-logo-open'
-  | String;
+export type Icon = String;
```
raphaellueckl: I'm against removing the possible values. This is very valuable when you have autocomplete in your IDE, as you see directly what you can choose from.

src/components/10-atoms/icon/index.js
```diff
@@ -55,26 +55,25 @@ class AXAIcon extends LitElement {
...
-      size: { type: String, reflect: true },
+      size: { type: String, reflect: true, defaultValue: 'medium' },
       _loadedSvg: { type: String },
     };
   }
 
   constructor() {
     super();
     applyDefaults(this);
+    this.size = 'medium';
```
raphaellueckl: This is redundant to line 58. I would anyway delete both lines and add a default parameter to the render function.

raphaellueckl: Is it needed that you declare it as 'medium'? I don't see where that value is used in here.

src/components/10-atoms/icon/index.scss
```diff
@@ -1,4 +1,15 @@
-:host(:not([size=auto])) svg {
+:host([size='small']) svg {
+  width: 20px;
+  height: 20px;
+}
+
+:host(:not([size])) svg,
+:host([size='medium']) svg {
```
raphaellueckl: This line could be changed to:

`:host([size]) svg {`

so we would not need an explicit size "medium" at all.

raphaellueckl: Advantage of that is: If a user writes `<axa-icon size="hello"`, it would still fallback to the default value. also if `size=""` is being set.

src/components/10-atoms/icon/pw.ui.test.js
```diff
@@ -0,0 +1,58 @@
...
+      );
+      const axaIconSvg = await page.$('svg');
+
+      expect((await axaIconSvg.boundingBox()).width).toBe(100);
+      expect((await axaIconSvg.boundingBox()).height).toBe(100);
+    });
+  });
+});
```
raphaellueckl, line 60: nice!

src/components/10-atoms/icon/index.js
danyball: redundant: there are cases that the class is instanciated without the default values of the props (marco told me that).

medium: this prop is reflected. So, the user can see the attribute on his dom node.

src/components/10-atoms/icon/CHANGELOG.md
danyball: jep, now it is

src/components/10-atoms/icon/index.js
raphaellueckl: But shouldn't we then add the default value only in the constructor? Or do we need both?

src/components/10-atoms/icon/index.scss
```diff
@@ -1,4 +1,14 @@
-:host(:not([size=auto])) svg {
+:host(:not([size='original'])) svg {
```
raphaellueckl, line 1: nice!

src/components/10-atoms/icon/story.js
```diff
@@ -28,7 +28,14 @@ export default {
 
 export const Icon = () => {
   const showAll = boolean('Show list of available icons', false);
-  const noIcon = boolean('Do not set icon on component', false);
+  const noIcon = boolean('Do not set attribute "icon"', false);
+  const noSize = boolean('Do not set attribute "size"', false);
```
raphaellueckl, line 32: What are those for? To explicitly remove the dom attributes?

src/components/10-atoms/icon/index.js
danyball: I saw that only `axa-carousel` is setting these values at the constructor. Maybe markos issue is fixed by `applyDefaults()`. I will remove it from constructor.

src/components/10-atoms/icon/story.js
danyball: Yes. Needed to test the default behaviour.

raphaellueckl (approved):
nice!

PR #2172, opened 2021-05-11
Feature/modal children without margin

raphaellueckl:
# Merge Checklist:
- [x] Tests are sufficient.

### DIFF
--- a/src/components/30-organisms/modal/child.scss
+++ b/src/components/30-organisms/modal/child.scss
@@ -1,0 +1,19 @@
+axa-modal > * {
+  &:first-child {
+    // !important reason: An axa-heading child element sets margins on the host
+    // element. This can only be overridden by !important.
+    // This is harmless because: There is no scenario where it would make sense
+    // for the first element to have any margin-top, except if they work with
+    // negative margins, which is an anti-pattern.
+    margin-top: 0 !important;
+  }
+
+  &:last-child {
+    // !important reason: An axa-heading child element sets margins on the host
+    // element. This can only be overridden by !important.
+    // This is harmless because: There is no scenario where it would make sense
+    // for the first element to have any margin-top, except if they work with
+    // negative margins, which is an anti-pattern.
+    margin-bottom: 0 !important;
+  }
+}
--- a/src/components/30-organisms/modal/child.scss
+++ b/src/components/30-organisms/modal/child.scss
@@ -1,4 +1,4 @@
-axa-modal > * {
+axa-modal[forced] > * {
   &:first-child {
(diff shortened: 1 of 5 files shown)


src/components/30-organisms/modal/child.scss
```diff
@@ -0,0 +1,19 @@
...
+    margin-top: 0 !important;
+  }
+
+  &:last-child {
+    // !important reason: An axa-heading child element sets margins on the host
+    // element. This can only be overridden by !important.
+    // This is harmless because: There is no scenario where it would make sense
+    // for the first element to have any margin-top, except if they work with
+    // negative margins, which is an anti-pattern.
+    margin-bottom: 0 !important;
```
danyball, line 17: This do not have any effect at SB example?

src/components/30-organisms/modal/child.scss
```diff
@@ -0,0 +1,19 @@
+axa-modal[forced] > * {
+  &:first-child {
+    // !important reason: An axa-heading child element sets margins on the host
+    // element. This can only be overridden by !important.
+    // This is harmless because: There is no scenario where it would make sense
+    // for the first element to have any margin-top, except if they work with
+    // negative margins, which is an anti-pattern.
+    margin-top: 0 !important;
```
danyball, line 8: Big change for very small outcome... its ok, i can see a difference.

danyball: This is a axa-heading thing.. so maybe something like `&:first-child & axa-heading`? Ah, is that possible with versioning?!

raphaellueckl: No this is also a thing for normal headings. It anyway makes sense to only use margin-top 0 in all cases, to prevent weird and unequal modals.

In my opinion, this change is not so big, because we use childelements, so it should be naturally, that we have styling for them in place.

src/components/30-organisms/modal/child.scss
raphaellueckl: Yeah but does that matter? The button does not have any margins by default. That is ok too if you ask me. If it would have, this css property would erase it.

danyball: This is a fix for something wo do not know if it will occur in future. I would just set margin-top.

src/components/30-organisms/modal/child.scss
danyball: Or the modal do not use any spacings for themselfes. And the customer have to style the content. But jea, its ok.

danyball (approved):

PR #2198, opened 2021-06-18
give checkbox an outline

### LINKED ISSUE #2080: AxaCheckbox lacks focused outline
@axa-ch/checkbox 5.1.9

When using keyboard to navigate through form elements axa checkbox doesn't have any visual indication that it is being focused.

Further reading:
https://www.a11yproject.com/posts/2013-01-25-never-remove-css-outlines/

raphaellueckl:
Fixes #2080

# Merge Checklist:
- [x] Tests are sufficient.

### DIFF
--- a/src/components/10-atoms/checkbox/index.js
+++ b/src/components/10-atoms/checkbox/index.js
@@ -114,6 +114,14 @@ class AXACheckbox extends NoShadowDOM {
     onChange(event);
   }
 
+  handleSpacebar(e) {
+    // Key: Space
+    if (e.keyCode === 32) {
+      this.checked = !this.checked;
+      this.state.native = this.checked;
+    }
+  }
+
   // throttle re-rendering to once per animation frame
   // (helps e.g. with static websites using this component together with static children
   // (like <axa-checkbox><span>my label</span></axa-checkbox>),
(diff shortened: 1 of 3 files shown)


src/components/10-atoms/checkbox/index.js
```diff
@@ -218,7 +232,12 @@ class AXACheckbox extends NoShadowDOM {
           </label>
         `
       : html`
-          <label for="${refId}" class="a-checkbox__wrapper">
+          <label
+            for="${refId}"
+            class="a-checkbox__wrapper"
+            tabindex="0"
```
marcwelz, line 238: Is this needed?

I thought using `tabindex` is dangerous since it can lead to unwanted inconvenience in combination with other components who can be addressed with the tabulator (You can skip whole elements, [example](https://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_global_tabindex) ).

I might be wrong but I would at least test it with other components

marcwelz (changes requested):

01231 (approved):
Is this per AXA design?

src/components/10-atoms/checkbox/index.js
raphaellueckl: Yeah it can be misused. In this case it is fine, because the label contains the input element and because I use a tabindex of 0 instead of any positive value (that is mostly being warned about).

marcwelz (approved):

PR #2201, opened 2021-06-22
File-upload-onchange

### LINKED ISSUE #2159: File Upload - onChange support
We should be able to deliver change events to the user with out file upload component. `onChange` should probably have the current list of files in it, ready to ship it over the network.

01231:
Fixes #2159

# Merge Checklist:
- [x] Tests are sufficient.

### DIFF
--- a/src/components/20-molecules/file-upload/README.md
+++ b/src/components/20-molecules/file-upload/README.md
@@ -129,3 +129,7 @@ The attribute `onFileRemove` function executes, as soon as the user removed a fi
 ### allowedFileTypes
 
 In the `allowedFileTypes` attribute, you can declare the allowed file types that a user can upload (ex: only pdf's are allowed). If the attribute is empty, every file type is allowed.
+
+### onChange
+
+The attribute `onChange` function executes, as soon as the user removes or adds new files. Adding new files include files from drag & drop or the `<axa-input-file>`.
--- a/src/components/20-molecules/file-upload/README.md
+++ b/src/components/20-molecules/file-upload/README.md
@@ -132,4 +132,4 @@ In the `allowedFileTypes` attribute, you can declare the allowed file types that
 
 ### onChange
 
-The attribute `onChange` function executes, as soon as the user removes or adds new files. Adding new files include files from drag & drop or the `<axa-input-file>`. Custom data (files) gets passed and can be accessed over `e.detail`.
+The attribute `onChange` function executes, as soon as the user removes or adds new files. Adding new files includes files from drag & drop or the `<axa-input-file>`. Custom data (files) gets passed and can be accessed over `e.detail`.
--- a/src/components/20-molecules/file-upload/README.md
+++ b/src/components/20-molecules/file-upload/README.md
@@ -132,4 +132,4 @@ In the `allowedFileTypes` attribute, you can declare the allowed file types that
 
 ### onChange
 
-The attribute `onChange` function executes, as soon as the user removes or adds new files. Adding new files includes files from drag & drop or the `<axa-input-file>`. Custom data (files) gets passed and can be accessed over `e.detail`.
+The attribute `onChange` function executes as soon as the user removes or adds new files. Adding new files includes files from drag & drop or the `<axa-input-file>`. Custom data (files) gets passed and can be accessed over `e.detail`.
--- a/src/components/20-molecules/file-upload/README.md
+++ b/src/components/20-molecules/file-upload/README.md
@@ -132,4 +132,4 @@ In the `allowedFileTypes` attribute, you can declare the allowed file types that
 
 ### onChange
 
-The attribute `onChange` function executes, as soon as the user removes or adds new files. Adding new files include files from drag & drop or the `<axa-input-file>`.
+The attribute `onChange` function executes, as soon as the user removes or adds new files. Adding new files include files from drag & drop or the `<axa-input-file>`. Custom data (files) gets passed and can be accessed over `e.detail`.
--- a/src/components/20-molecules/file-upload/demo.js
+++ b/src/components/20-molecules/file-upload/demo.js
@@ -37,13 +37,56 @@ storiesOf('Examples/File Upload/Pure HTML', module)
 
     const template = html`
       <div style="width:455px;">
-        <axa-file-upload class="js-file-upload__file-upload"
+        <axa-file-upload
+          maxSizeOfSingleFileKB="500"
+          class="js-file-upload__file-upload"
+          @change="${e => console.log('change 123')}"
           >These files are going to be uploaded</axa-file-upload
         >
       </div>
       <form @click="${handleSubmit}" style="margin-top:40px;">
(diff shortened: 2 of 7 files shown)


src/components/20-molecules/file-upload/demo.js
```diff
@@ -37,7 +37,9 @@ storiesOf('Examples/File Upload/Pure HTML', module)
 
     const template = html`
       <div style="width:455px;">
-        <axa-file-upload class="js-file-upload__file-upload"
+        <axa-file-upload
+          maxSizeOfSingleFileKB="500"
```
raphaellueckl, line 42: Why 500? This line seems a bit random to me, or is there a usecase for that?

src/components/20-molecules/file-upload/index.js
```diff
@@ -158,6 +160,13 @@ class AXAFileUpload extends LitElement {
     }
   }
 
+  fireCustomChangeEvent() {
+    this.onChange({
+      detail: this.files,
+    });
+    fireCustomEvent('change', this.files, this);
```
raphaellueckl, line 172: Why doing it like this? Better register a 'change' listener on the axa-input-file tag and just pass on that event. Or is that not possible?

raphaellueckl (changes requested):
Maybe add a test too.

src/components/20-molecules/file-upload/demo.js
01231: This is just that not all files are displayed with an error message (default is 100kb). The number is indeed random.

src/components/20-molecules/file-upload/index.js
01231: The change listener would only be called if the file was uploaded via the input file. And this happens max. one time (the first time a file gets uploaded)

01231: All other usecases like drag & drop or file deletion would be ignored.

src/components/20-molecules/file-upload/README.md
```diff
@@ -124,3 +124,7 @@ The attribute `onFileDrop` function executes, as soon as the user dropped one or
 ### onFileRemove
 
 The attribute `onFileRemove` function executes, as soon as the user removed a file from the file-upload.
+
+### onChange
+
+The attribute `onChange` function executes, as soon as the user removes or adds new files. Adding new files include files from drag & drop or the `<axa-input-file>`. Custom data (files) gets passed and can be accessed over `e.detail`.
```
marcwelz: I'm not an english expert but I think it should be includes instead of include

01231: done

src/components/20-molecules/file-upload/demo.js
raphaellueckl: Then why not 99999999?

raphaellueckl (changes requested):
This has to work: ```document.querySelector('axa-file-upload').addEventListener('change', ()=>{console.log('hello')})```

Whenever a file gets dropped, added via button or removed, a change event should be fired.

And the change event should contain the stuff that a normal change event would contain for the native element `<input type="file"....`

src/components/20-molecules/file-upload/index.js
raphaellueckl: Best would be to artificially reroute the changes to the input-file button. The input file button would always be there (visible or not) and contain the current state of the file upload. This keeps it simple in terms of single source of truth.

src/components/20-molecules/file-upload/demo.js
01231: Because then no file would be invalid.

src/components/20-molecules/file-upload/index.js
01231: But the file upload as a component is now the singel source of truth. In my opinion it's not nice that a input file for example triggers an event onDeletion, because nowbody expects it to do so.

01231:
> This has to work: `document.querySelector('axa-file-upload').addEventListener('change', ()=>{console.log('hello')})`
> 
> Whenever a file gets dropped, added via button or removed, a change event should be fired.
> 
> And the change event should contain the stuff that a normal change event would contain for the native element `<input type="file"....`

This works. Check out the latest deployment. I made demos for react and native that show how it works

src/components/20-molecules/file-upload/README.md
```diff
@@ -124,3 +124,7 @@ The attribute `onFileDrop` function executes, as soon as the user dropped one or
 ### onFileRemove
 
 The attribute `onFileRemove` function executes, as soon as the user removed a file from the file-upload.
+
+### onChange
+
+The attribute `onChange` function executes, as soon as the user removes or adds new files. Adding new files includes files from drag & drop or the `<axa-input-file>`. Custom data (files) gets passed and can be accessed over `e.detail`.
```
markus-walther: no comma after 'executes'!

src/components/20-molecules/file-upload/index.js
markus-walther: I agree with Noah, this is fine and readable. But since there has been some discussion, a comment motivating this way of doing things by mentioning drag/drop and deletion would help!

markus-walther (changes requested):
Noah, I like it in general. Few small findings. Didn't  actually test things though, trusting you on that!

src/components/20-molecules/file-upload/demo.js
raphaellueckl: But does the file need to be invalid at 500? I mean, if this is just a random number, then no file should ever become invalid right?

01231: Why did I changed it to 500kb? Because most of my files are bigger than 100kb and when I want to test I also want to test with valid files.
Why haven't I set it to 99999999kb? Because then all files would have a valid size, but I want to test both valid and ivalid ones...

And also it makes no sense to set it to 99999999kb because i can't imagine a use case to allow a 100gb file.

raphaellueckl: Oh god... I totally missed that this is a demo and not the index.js file, sorry.

src/components/20-molecules/file-upload/index.js
raphaellueckl: Ok, I understand that my idea would not be a proper solution. Sad that `input type=file` has such a bad api.

src/components/20-molecules/file-upload/ui.test.js
```diff
@@ -252,11 +252,10 @@ fixture('File upload - Remove file event').page(
   `${host}/iframe.html?id=examples-file-upload-react--story&viewMode=story`
```
raphaellueckl, line 252: We once decided that we won't touch testcafe tests anymore. Rather we migrate to playwright, if any test is failing. :)

raphaellueckl (dismissed):

markus-walther (approved):
LGTM

PR #2208, opened 2021-07-01
add invalid on file upload

### LINKED ISSUE #2196: Invalid flag in file upload component
@asalinifugoch had a requirement, where he wants to check if the file upload reports a "valid" state. He has to consider this state before making a submit possible.

Since we have that functionality on our other form elements, I don't see why we shouldn't apply the same logic to it, as this component has validation implemented, just not an api to check the state.

marcwelz:
Fixes #2196

# Merge Checklist:
- [x] Tests are sufficient.

### DIFF
--- a/src/components/20-molecules/file-upload/README.md
+++ b/src/components/20-molecules/file-upload/README.md
@@ -60,7 +60,7 @@ if (files.length > 0) {
 | `orText="or"`                               | Specify the text on the start view box                                   |
 | `wrongFileTypeStatusText="Wrong file-type"` | Specify the error message for dropped files with wrong file-type         |
 | `allowedFileTypes="image/jpg"`              | Determindes the allowed file types. If empty, every file type is allowed |
-| `invalid`                                   | Checks if the file upload reports a "valid" state. (type: Boolean)       |
+| `invalid`                                   | Checks if the file upload reports a "invalid" state. (type: Boolean, read only)       |
 
 ### inputFileText
 
@@ -133,4 +133,4 @@ In the `allowedFileTypes` attribute, you can declare the allowed file types that
 
 ### invalid
 
-Checks if the file upload reports a "valid" state. You can consider this state before making a submit possible.
+Checks if the file upload reports a "invalid" state (read-only). You can consider this state before making a submit possible.
--- a/src/components/20-molecules/file-upload/README.md
+++ b/src/components/20-molecules/file-upload/README.md
@@ -43,24 +43,24 @@ if (files.length > 0) {
 
 ### Variant
 
-| Attribute                                   | Details                                                                  |
-| ------------------------------------------- | ------------------------------------------------------------------------ |
-| `inputFileText="Upload file"`               | Text in the input-file component                                         |
-| `maxSizeOfSingleFileKB="10"`                | Maximal size of a single, already compromised file in KB                 |
-| `maxSizeOfAllFilesKB="15"`                  | Maximal size of all compromised files together in KB                     |
-| `maxNumberOfFiles="10"`                     | Maximal number of files                                                  |
-| `showFileOverview="false"`                  | Switch between normal view and file overview                             |
-| `icon="cloud-upload"`                       | Specify the upload icon in input-file component                          |
-| `fileTooBigStatusText="Error occurred"`     | Specify the error message for too big files                              |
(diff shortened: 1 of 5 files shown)


src/components/20-molecules/file-upload/README.md
```diff
@@ -60,6 +60,7 @@ if (files.length > 0) {
 | `orText="or"`                               | Specify the text on the start view box                                   |
 | `wrongFileTypeStatusText="Wrong file-type"` | Specify the error message for dropped files with wrong file-type         |
 | `allowedFileTypes="image/jpg"`              | Determindes the allowed file types. If empty, every file type is allowed |
+| `invalid`                                   | Checks if the file upload reports a "valid" state. (type: Boolean)       |
```
01231: Would go for:
Checks if the file upload reports a "invalid" state. (type: Boolean)

src/components/20-molecules/file-upload/README.md
```diff
@@ -129,3 +130,7 @@ The attribute `onFileRemove` function executes, as soon as the user removed a fi
 ### allowedFileTypes
 
 In the `allowedFileTypes` attribute, you can declare the allowed file types that a user can upload (ex: only pdf's are allowed). If the attribute is empty, every file type is allowed.
+
+### invalid
+
+Checks if the file upload reports a "valid" state. You can consider this state before making a submit possible.
```
01231: Same here: would use "invalid"

src/components/20-molecules/file-upload/story.js
```diff
@@ -79,6 +79,7 @@ export const FileUpload = () => {
     'The following files are being transferred:'
   );
   const preventFileCompression = boolean('preventFileCompression', false);
+  const invalid = boolean('invalid', false);
```
01231: I personally wouldn't put the new attribute into the story because it's read only. The `outputdate` attribute from the datepicker isn't is the story either (you can look it up).

src/components/20-molecules/file-upload/README.md
01231: Also mention that it is intended for read only. You can look in the readme of the datepicker and search for the `outputdate` attribute. It's read only too.

src/components/20-molecules/file-upload/story.js
01231: Rather make a good demo that shows the value of the `invalid` attribute when the user uploads a file

01231 (changes requested):
Logic looks good. Consider making a demo and tests.

src/components/20-molecules/file-upload/story.js
```diff
@@ -102,6 +116,13 @@ export const FileUpload = () => {
         >${headerText}</axa-file-upload
       >
     </div>
+
+    <axa-button style="margin: 10px 0">
```
01231: would move this (your whole example) to the demos to keep the stories clean.

src/components/20-molecules/file-upload/demo.js
```diff
@@ -90,6 +91,55 @@ storiesOf('Examples/File Upload/Pure HTML', module)
...
+      });
+    });
+
+    const template = html`
+      <div style="width: 455px">
+        <axa-file-upload
+          inputFileText="Upload file"
+          maxSizeOfSingleFileKB="100"
+          maxSizeOfAllFilesKB="500"
+          maxNumberOfFiles="10"
```
01231: I think you can delete most of the attributes because they are already set as default

src/components/20-molecules/file-upload/demo.js
```diff
@@ -90,6 +91,55 @@ storiesOf('Examples/File Upload/Pure HTML', module)
     `;
 
     const wrapper = document.createElement('div');
+    render(template, wrapper);
+    return wrapper;
+  })
+  .add('invalid', () => {
+    const wrapper = document.createElement('div');
+
+    setTimeout(() => {
```
01231: I'd prefer using litElement event listeners. You could just add `@click=${handleSubmit}` just like the `In a form` demo from the file upload does. This would be much cleaner.
https://lit-element.polymer-project.org/guide/events

01231 (changes requested):
Nice demo. I have found some changes that would make the code even cleaner in my opinion.

01231 (approved):
LGTM

PR #2218, opened 2021-07-28
Feature/lit migration

### LINKED ISSUE #2207: Migrate to lit
https://lit.dev/ has arrived. It is the official new lit-element, better, smaller, faster.

We should migrate everything to it.

After we did that, we should attempt to add the AXA logo to the "Many of the world's most forward-looking organizations are building with Lit" section...

01231:
Fixes #2207

# Merge Checklist:
- [x] Tests are sufficient.
- [x] Documentation is up to date. -> #2220 
- [x] Changelog is up to date. -> #2220 


01231:
## IE not working 
### Message from browser stack when testing feature branch in IE
Screenshot 2021-08-16 at 13 35 38

### Thing I tried
- research (articles don't seem to be relevant, but they all have to do with polyfills): https://github.com/angular/angular/issues/21202 and https://github.com/facebook/create-react-app/issues/6924
- Add polyfills (listed [here](https://lit.dev/docs/tools/requirements/#polyfills)) -> didn't work

01231:
I removed `@ungap/global-this `and `lit/polyfill-support` polyfill, but the error remains the same when testing in browser stack..

markus-walther:
The syntax error thrown is line 2, character position 7092 of https://axa-ch-webhub-cloud.github.io/plib-feature/feature/lit-migration/vendors~main.1e0ee724f0266d5bb054.manager.bundle.js.

It reads like this: `class css_tag_s{....`. ES6 classes are a syntax error in IE11.

This is part of the ShadyCSS polyfill apparently.

The investigation now needs to focus on why this was not transpiled.

LucaMele (commented):
I see that lit-element node module name changed. Pattern library exports always in ES6 via rollup. To make new framework work on the webhub, you need to add new node module to transpilation whitelist in midgard: https://github.com/axa-ch-webhub-cloud/midgard/blob/7f6ef211625945db5fd7126947c20a0e0c4741be/scripts/templates/webpack-conf.js#L15

LucaMele:
> The syntax error thrown is line 2, character position 7092 of https://axa-ch-webhub-cloud.github.io/plib-feature/feature/lit-migration/vendors~main.1e0ee724f0266d5bb054.manager.bundle.js.
> 
> It reads like this: `class css_tag_s{....`. ES6 classes are a syntax error in IE11.
> 
> This is part of the ShadyCSS polyfill apparently.
> 
> The investigation now needs to focus on why this was not transpiled.

See my comment

markus-walther:
The saga continues... New error is in https://axa-ch-webhub-cloud.github.io/plib-feature/feature/lit-migration/vendors~main.be4a5b57.iframe.bundle.js, line 2, character 150158. And it's the fact that https://developer.mozilla.org/en-US/docs/Web/API/Element/append is not supported in IE at all. Polyfills exists.

Now the next question is: Does lit.dev account for this already, and we just made a mistake with our polyfills, or did this slip through? It is likely we made a mistake, as always. 

Time to investigate more...

markus-walther:
With the 2 latest commits Storybooks loads without console errors on IE11.

It was necessary to disable jest unit tests, because all kinds of errors appeared (recall that it uses jsdom and runs on node, and a faithful DOM simulation is difficult, esp. in combination with polyfills. The latter combination apparently changed with the polyfills version upgrade...).

However, the datepicker does not visibly react to user interaction, i.e. mouseclicks.

Debugging in minified code in IE 11 is very painful...

https://raygun.com/blog/debug-javascript-internet-explorer/ helps a little with a way to intercept control after a mouse click.

The observable difference I could come up with sofar is that in IE the control flow sometimes ends up in the general background-click handler instead of a specific foreground handler such as the ones for month navigation.

Also visible is the fact that event handlers are heavily patched/wrapped by e.g. ShadyCSS.

The whole situation does not exactly inspire trust...

markus-walther:
Resuming work on this. The requirement that it must work under IE is gone, since IE is deprecated also in an AXA context.

src/components/00-materials/README.md
```diff
@@ -11,10 +11,10 @@ npm install @axa-ch/materials
 ```
 
 ```js
-import { svg } from 'lit-html';
+import { unsafeHTML } from 'lit/directives/unsafe-html';
 import { Arrow_forwardSvg } from '@axa-ch/materials/icons/material-design';
 
-<span>${svg(Arrow_forwardSvg)}</span>;
+<span>${unsafeHTML(Arrow_forwardSvg)}</span>;
```
amir-ag, line 17: Apparently 'unsafeSVG' does also exist. Would this be an alternative?

markus-walther: no, because it is provided for SVG **parts**, but we're passing an entire SVG with top-level `<svg>`!

amir-ag (approved):

PR #2251, opened 2021-12-27
Top content bar

### LINKED ISSUE #2252: axa-top-content-bar: icons, mobile, closable
As a result of https://jira.axa.com/jira/browse/WEBHUB-1100, the axa-top-content-bar component needs to be extended with new functionality. 

It should allow for icons, be more usable on mobile, and closable.

Four new properties would allow to activate and control the new functionality:

`icon: { type: String },
      stickymobile: { type: Boolean },
      closable: { type: Boolean },
      closed: { type: Boolean }`

The design changes are already specified by the designers.

bozojovicic:
Fixes #2252.

Top content bar

# Merge Checklist:
- [x] Tests are sufficient.

### DIFF
--- a/src/components/00-materials/styles/00-colors.scss
+++ b/src/components/00-materials/styles/00-colors.scss
@@ -33,6 +33,7 @@ $color-tosca: #914146 !default;
 $color-azalea: #e196aa !default;
 $color-cotton-candy: #fad6de !default;
 $color-apache: #ddbe65 !default;
+$color-underbrush: #bc9d45 !default;
 $color-dune: #fcd385 !default;
 $color-acid: #f0ff93 !default;
 $color-logan: #9190ac !default;
--- a/src/components/10-atoms/icon/index.js
+++ b/src/components/10-atoms/icon/index.js
@@ -14,6 +14,11 @@ import {
   File_uploadSvg,
   Cloud_uploadSvg,
   Check_circleSvg,
+  CheckSvg,
+  Info_outlineSvg,
+  WarningSvg,
+  Cloud_queueSvg,
+  CloseSvg,
 } from '@axa-ch/materials/icons/material-design';
 import { ArrowRightSvg, ArrowLeftSvg } from '@axa-ch/materials/icons';
 import { AxaLogoSvg, AxaLogoOpenSvg } from '@axa-ch/materials/images';
@@ -51,6 +56,11 @@ class AXAIcon extends LitElement {
       'axa-logo': AxaLogoSvg,
       'axa-logo-open': AxaLogoOpenSvg,
       'check-circle': Check_circleSvg,
+      check: CheckSvg,
+      'info-outline': Info_outlineSvg,
+      warning: WarningSvg,
+      'cloud-queue': Cloud_queueSvg,
+      close: CloseSvg,
     };
   }
 
--- a/src/components/20-molecules/top-content-bar/README.md
+++ b/src/components/20-molecules/top-content-bar/README.md
@@ -32,7 +32,7 @@ Set the boolean attribute `stickymobile` to `true` to display the bar at the bot
 
 ### overlaydesktop
 
-Set the boolean attribute `overlaydesktop` to `true` to display the bar on top of the DOM element that comes after this element. Usually that would be the Keyvisual CTA component. This feature applies only for the screen width larger than 411px. For mobile devices the bar will either be displayed as usually with `position:static` or at the bottom of the screen if the attribute `stickymobile` is enabled. 
+Set the boolean attribute `overlaydesktop` to `true` to display the bar on top of the DOM element that comes after this element. Usually that would be the Keyvisual CTA component. This feature applies only for the screen width larger than 411px. For mobile devices the bar will either be displayed as usually with `position:static` or at the bottom of the screen if the attribute `stickymobile` is enabled.
 
 ### closable
 
@@ -40,7 +40,7 @@ Set the boolean attribute `closable` to `true` to display the Close button. Once
 
 ### initiallyclosed
 
-Set the boolean attribute `initiallyclosed` to `true` to make the bar is initially closed (on page load). By default it is initially opened ie. displayed. If the bar is initially closed it will be displayed after the check that the bar is not closed by the user during that browser session. Also if the bar is initially opened it will be automatically closed after the check that the bar is closed by the user during that browser session. 
+Set the boolean attribute `initiallyclosed` to `true` to make the bar is initially closed (on page load). By default it is initially opened ie. displayed. If the bar is initially closed it will be displayed after the check that the bar is not closed by the user during that browser session. Also if the bar is initially opened it will be automatically closed after the check that the bar is closed by the user during that browser session.
 
 ### onClick
 
--- a/src/components/20-molecules/top-content-bar/README.md
+++ b/src/components/20-molecules/top-content-bar/README.md
@@ -42,6 +42,10 @@ Set the boolean attribute `closable` to `true` to display the Close button. Once
 
 Set the boolean attribute `initiallyclosed` to `true` to make the bar is initially closed (on page load). By default it is initially opened ie. displayed. If the bar is initially closed it will be displayed after the check that the bar is not closed by the user during that browser session. Also if the bar is initially opened it will be automatically closed after the check that the bar is closed by the user during that browser session.
 
+### closed
+
+The boolean property `closed` can be used to hide or show the bar on the page. We use this property after we check the sessionStorage item `top-content-bar-closed`.
+
 ### onClick
 
 The function-valued attribute `onClick` can be used as a callback prop for React and other frameworks.
--- a/src/components/20-molecules/top-content-bar/README.md
+++ b/src/components/20-molecules/top-content-bar/README.md
@@ -28,7 +28,7 @@ Using the string-valued attribute `icon` as icon name, an icon will be rendered.
 
 ### stickymobile
 
-Set the Boolean attribute `stickymobile` to `true` to display the bar at the bottom of the screen on mobile devices with width < 411px.
+Set the Boolean attribute `stickymobile` to `true` to display the bar at the bottom of the screen on mobile devices with width < 576px.
 
 ### closable
 
--- a/src/components/20-molecules/top-content-bar/README.md
+++ b/src/components/20-molecules/top-content-bar/README.md
@@ -28,23 +28,15 @@ Using the string-valued attribute `icon` as icon name, an icon will be rendered.
 
 ### stickymobile
 
-Set the boolean attribute `stickymobile` to `true` to display the bar at the bottom of the screen on mobile devices with width < 411px.
-
-### overlaydesktop
-
-Set the boolean attribute `overlaydesktop` to `true` to display the bar on top of the DOM element that comes after this element. Usually that would be the Keyvisual CTA component. This feature applies only for the screen width larger than 411px. For mobile devices the bar will either be displayed as usually with `position:static` or at the bottom of the screen if the attribute `stickymobile` is enabled.
+Set the Boolean attribute `stickymobile` to `true` to display the bar at the bottom of the screen on mobile devices with width < 411px.
 
 ### closable
 
-Set the boolean attribute `closable` to `true` to display the Close button. Once the bar is closed it will not be displayed any more during that browser session.
-
-### initiallyclosed
-
-Set the boolean attribute `initiallyclosed` to `true` to make the bar is initially closed (on page load). By default it is initially opened ie. displayed. If the bar is initially closed it will be displayed after the check that the bar is not closed by the user during that browser session. Also if the bar is initially opened it will be automatically closed after the check that the bar is closed by the user during that browser session.
+Set the Boolean attribute `closable` to `true` to display the Close button. Once the bar is closed it will not be displayed any more during that browser session.
 
 ### closed
 
-The boolean property `closed` can be used to hide or show the bar on the page. We use this property after we check the sessionStorage item `top-content-bar-closed`.
+The Boolean property `closed` can be used to hide or show the bar on the page. We use this property after we check the sessionStorage item `top-content-bar-closed`.
 
 ### onClick
 
--- a/src/components/20-molecules/top-content-bar/README.md
+++ b/src/components/20-molecules/top-content-bar/README.md
@@ -1,7 +1,7 @@
 # Top content bar
 
 Used as top of the page warning or info box that can show text and have a call to action button.
-It can also be display at the bottom of the page on mobile devices.
+It can also be display at the bottom of the screen on mobile devices (see below for the stickymobile attribute).
 
 There is also the possibility to add child elements.
 
@@ -24,19 +24,23 @@ The attribute `href` specifies the link, that should be opened by pressing on th
 
 ### icon
 
-Using the string attribute `icon` as icon name, an icon will be rendered. To see the full list of possible icons, see the [axa-icon](https://github.com/axa-ch-webhub-cloud/pattern-library/blob/develop/src/components/10-atoms/icon/README.md) README.
+Using the string-valued attribute `icon` as icon name, an icon will be rendered. To see the full list of possible icons, see the [axa-icon](https://github.com/axa-ch-webhub-cloud/pattern-library/blob/develop/src/components/10-atoms/icon/README.md) README.
 
 ### stickymobile
 
-Set the attribute `stickymobile` to `true` to display the bar at the bottom of the screen on mobile devices with width < 411px.
+Set the boolean attribute `stickymobile` to `true` to display the bar at the bottom of the screen on mobile devices with width < 411px.
 
 ### overlaydesktop
 
-Set the attribute `overlaydesktop` to `true` to display the bar on top of DOM element that comes after this element. Usually that would be the Keyvisual CTA component.
+Set the boolean attribute `overlaydesktop` to `true` to display the bar on top of the DOM element that comes after this element. Usually that would be the Keyvisual CTA component. This feature applies only for the screen width larger than 411px. For mobile devices the bar will either be displayed as usually with `position:static` or at the bottom of the screen if the attribute `stickymobile` is enabled. 
 
-### closeable
+### closable
 
-Set the attribute `closeable` to `true` to display the Close button. Once the bar is closed it will not be displayed any more during that browser session.
+Set the boolean attribute `closable` to `true` to display the Close button. Once the bar is closed it will not be displayed any more during that browser session.
+
+### initiallyclosed
+
+Set the boolean attribute `initiallyclosed` to `true` to make the bar is initially closed (on page load). By default it is initially opened ie. displayed. If the bar is initially closed it will be displayed after the check that the bar is not closed by the user during that browser session. Also if the bar is initially opened it will be automatically closed after the check that the bar is closed by the user during that browser session. 
 
 ### onClick
 
--- a/src/components/20-molecules/top-content-bar/README.md
+++ b/src/components/20-molecules/top-content-bar/README.md
@@ -1,6 +1,7 @@
 # Top content bar
 
 Used as top of the page warning or info box that can show text and have a call to action button.
+It can also be display at the bottom of the page on mobile devices.
 
 There is also the possibility to add child elements.
 
@@ -20,6 +21,22 @@ If no attribute `href` is set, it will render an `axa-button` and the `onClick` 
 ### href
 
 The attribute `href` specifies the link, that should be opened by pressing on the button with the text defined in `ctatext`. This is mainly for semantic reasons, you still need to forward the user to the new location, by listening to the `onClick` callback. Just providing the `href` attribute will not do anything by itself.
+
+### icon
+
+Using the string attribute `icon` as icon name, an icon will be rendered. To see the full list of possible icons, see the [axa-icon](https://github.com/axa-ch-webhub-cloud/pattern-library/blob/develop/src/components/10-atoms/icon/README.md) README.
+
+### stickymobile
+
+Set the attribute `stickymobile` to `true` to display the bar at the bottom of the screen on mobile devices with width < 411px.
+
+### overlaydesktop
+
+Set the attribute `overlaydesktop` to `true` to display the bar on top of DOM element that comes after this element. Usually that would be the Keyvisual CTA component.
+
+### closeable
+
+Set the attribute `closeable` to `true` to display the Close button. Once the bar is closed it will not be displayed any more during that browser session.
 
 ### onClick
 
--- a/src/components/20-molecules/top-content-bar/child.scss
+++ b/src/components/20-molecules/top-content-bar/child.scss
@@ -1,14 +1,14 @@
 axa-top-content-bar {
-  axa-link {
+  > axa-link {
     padding: 5px;
   }
 
   a {
-    color: white;
+    color: $color-prim-white;
     text-decoration: underline;
 
     &:hover {
-      color: white;
+      color: $color-prim-white;
       text-decoration: none;
     }
   }
--- a/src/components/20-molecules/top-content-bar/child.scss
+++ b/src/components/20-molecules/top-content-bar/child.scss
@@ -1,5 +1,17 @@
-axa-top-content-bar > axa-link {
-  padding: 5px;
+axa-top-content-bar {
+  axa-link {
+    padding: 5px;
+  }
+
+  a {
+    color: white;
+    text-decoration: underline;
+
+    &:hover {
+      color: white;
+      text-decoration: none;
+    }
+  }
 }
 
 @include breakpoint($mediaquery-md-up) {
--- a/src/components/20-molecules/top-content-bar/index.js
+++ b/src/components/20-molecules/top-content-bar/index.js
@@ -138,9 +138,17 @@ class AXATopContentBar extends InlineStyles {
   }
 
   render() {
-    const { variant, icon, stickymobile, overlaydesktop, closable, closed, initiallyclosed } = this;
+    const {
+      variant,
+      icon,
+      stickymobile,
+      overlaydesktop,
+      closable,
+      closed,
+      initiallyclosed,
+    } = this;
 
-    if (closed || initiallyclosed) return html ``;
+    if (closed || initiallyclosed) return html``;
 
     const btnHtml = this.getButtonHtml();
 
@@ -199,10 +207,10 @@ class AXATopContentBar extends InlineStyles {
             </div>
             ${
               closable
-                ? html ``
+                ? html``
                 : html`
-                  <div class="m-top-content-bar-empty"></div>
-                `
+                    <div class="m-top-content-bar-empty"></div>
+                  `
             }
           </div>
 
--- a/src/components/20-molecules/top-content-bar/index.js
+++ b/src/components/20-molecules/top-content-bar/index.js
@@ -38,9 +38,7 @@ class AXATopContentBar extends InlineStyles {
       variant: { type: String },
       icon: { type: String },
       stickymobile: { type: Boolean },
-      overlaydesktop: { type: Boolean },
       closable: { type: Boolean },
-      initiallyclosed: { type: Boolean },
       closed: { type: Boolean },
       onClick: { type: Function, attribute: false },
     };
@@ -121,20 +119,9 @@ class AXATopContentBar extends InlineStyles {
   }
 
   render() {
-    const {
-      variant,
-      icon,
-      stickymobile,
-      overlaydesktop,
-      closable,
-      closed,
-    } = this;
+    const { variant, icon, stickymobile, closable, closed } = this;
 
-    if (!closed) {
-      this.initiallyclosed = false;
-    }
-
-    if (closed || this.initiallyclosed) return html``;
+    if (closed) return html``;
 
     const btnHtml = this.getButtonHtml();
 
@@ -145,7 +132,6 @@ class AXATopContentBar extends InlineStyles {
     };
 
     const rootClasses = {
-      'm-top-content-bar__overlay-desktop': overlaydesktop,
       'm-top-content-bar__bottom-mobile': stickymobile,
     };
 
--- a/src/components/20-molecules/top-content-bar/index.js
+++ b/src/components/20-molecules/top-content-bar/index.js
@@ -171,7 +171,7 @@ class AXATopContentBar extends InlineStyles {
             
               <div class="m-top-content-bar__children">
                 <div class="m-top-content-bar__icon ${classMap(iconClasses)}">
-                  <axa-icon icon="${icon}"></axa-icon>
+                  <axa-icon size="large" icon="${icon}"></axa-icon>
                 </div>
                 
                 <div class="m-top-content-bar__content-text">
--- a/src/components/20-molecules/top-content-bar/index.js
+++ b/src/components/20-molecules/top-content-bar/index.js
@@ -1,4 +1,4 @@
-import { css, unsafeCSS } from 'lit-element';
+import { css, html, unsafeCSS } from 'lit-element';
 import { classMap } from 'lit-html/directives/class-map';
 /* eslint-disable import/no-extraneous-dependencies */
 import AXAContainer from '@axa-ch/container';
@@ -43,6 +43,7 @@ class AXATopContentBar extends InlineStyles {
       icon: { type: String },
       stickymobile: { type: String },
       overlaydesktop: { type: String },
+      closeable: { type: String },
       onClick: { type: Function, attribute: false },
     };
   }
@@ -127,7 +128,7 @@ class AXATopContentBar extends InlineStyles {
   }
 
   render() {
-    const { variant, icon, stickymobile, overlaydesktop } = this;
+    const { variant, icon, stickymobile, overlaydesktop, closeable } = this;
 
     const btnHtml = this.getButtonHtml();
 
@@ -143,6 +144,18 @@ class AXATopContentBar extends InlineStyles {
       hide: elementHidden,
     };
 
+    const iconClasses = {
+      'hide-icon': !icon,
+    };
+
+    const closeClasses = {
+      'hide-close': closeable !== 'true',
+    };
+
+    const buttonClasses = {
+      'hide-button': !btnHtml,
+    };
+
     return versionedHtml(this)`
       <article class="m-top-content-bar ${classMap(rootClasses)}">
         <div class="m-top-content-bar__container ${classMap(variantClasses)}">
@@ -152,25 +165,34 @@ class AXATopContentBar extends InlineStyles {
             <div class="m-top-content-bar__container-component">
             
               <div class="m-top-content-bar__children">
-                <div class="m-top-content-bar__icon">
+                <div class="m-top-content-bar__icon ${classMap(iconClasses)}">
                   <axa-icon icon="${icon}"></axa-icon>
                 </div>
                 
                 <div class="m-top-content-bar__content-text">
                   <slot></slot>
                 </div>
-                <div class="m-top-content-bar__content-button">
+                <div class="m-top-content-bar__content-button ${classMap(
+                  buttonClasses
+                )}">
                   ${btnHtml}
                 </div>
               </div>
 
             </div>
         
-            <div class="m-top-content-bar__close-button" @click="${
-              this.onClose
-            }">
+            <div class="m-top-content-bar__close-button ${classMap(
+              closeClasses
+            )}" @click="${this.onClose}">
               <axa-icon icon="close"></axa-icon>
-            </div> 
+            </div>
+            ${
+              closeable !== 'true'
+                ? html`
+                    <div class="m-top-content-bar-empty"></div>
+                  `
+                : null
+            }
           </div>
 
           <div class="m-top-content-bar__content-button-mobile">
--- a/src/components/20-molecules/top-content-bar/index.js
+++ b/src/components/20-molecules/top-content-bar/index.js
@@ -54,14 +54,14 @@ class AXATopContentBar extends InlineStyles {
     super();
     applyDefaults(this);
 
-    window.addEventListener('axa-top-content-bar-open', () => {
+    this.addEventListener('axa-top-content-bar-open', () => {
       if (this.initiallyclosed) {
         this.closed = false;
         this.initiallyclosed = false;
       }
     });
 
-    window.addEventListener('axa-top-content-bar-close', () => {
+    this.addEventListener('axa-top-content-bar-close', () => {
       if (!this.initiallyclosed) {
         this.closed = true;
       }
@@ -83,9 +83,8 @@ class AXATopContentBar extends InlineStyles {
     // item to sessionStorage.
     // The similar counts for the event 'axa-top-content-bar-open', we check the
     // sessionStorage in core module and then fire the event to open the bar
-    // if it's initially closed. Since we cannot read the elements in shadow dom,
-    // we have to use the window object as a target for the events.
-    fireCustomEvent('axa-top-content-bar-close', null, window);
+    // if it's initially closed.
+    fireCustomEvent('axa-top-content-bar-close', null, this);
   }
 
   firstUpdated() {
--- a/src/components/20-molecules/top-content-bar/index.js
+++ b/src/components/20-molecules/top-content-bar/index.js
@@ -4,6 +4,7 @@ import { classMap } from 'lit-html/directives/class-map';
 import AXAContainer from '@axa-ch/container';
 import AXAButton from '@axa-ch/button';
 import AXAButtonLink from '@axa-ch/button-link';
+import AXAIcon from '@axa-ch/icon';
 import InlineStyles from '../../../utils/inline-styles';
 import childStyles from './child.scss';
 
@@ -13,13 +14,18 @@ import {
 } from '../../../utils/component-versioning';
 import { applyDefaults } from '../../../utils/with-react';
 import styles from './index.scss';
+import fireCustomEvent from '../../../utils/custom-event';
 
+const elTagName = 'axa-top-content-bar';
+const elRootSelector = '.m-top-content-bar';
+const hideClass = 'hide';
+let elementHidden = true;
 /**
  *  We need InlineStyles to give children some margins.
  */
 class AXATopContentBar extends InlineStyles {
   static get tagName() {
-    return 'axa-top-content-bar';
+    return elTagName;
   }
 
   static get styles() {
@@ -34,6 +40,9 @@ class AXATopContentBar extends InlineStyles {
       ctatext: { type: String },
       href: { type: String },
       variant: { type: String },
+      icon: { type: String },
+      stickymobile: { type: String },
+      overlaydesktop: { type: String },
       onClick: { type: Function, attribute: false },
     };
   }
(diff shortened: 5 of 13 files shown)


src/components/20-molecules/top-content-bar/index.js
```diff
@@ -46,13 +55,31 @@ class AXATopContentBar extends InlineStyles {
...
     );
   }
 
+  onClose() {
+    const contBarEl = this.shadowRoot.querySelector(elRootSelector);
+    contBarEl.style.maxHeight = 0;
+    setTimeout(() => {
+      contBarEl.classList.add(hideClass);
+      fireCustomEvent('axa-top-bar-close', null, window);
+    }, 300);
```
Ben-Zahler: why do we need a 300ms delay?

src/components/20-molecules/top-content-bar/index.scss
```diff
@@ -9,18 +95,20 @@
 
     min-height: 50px;
     vertical-align: middle;
-
-    text-align: center;
```
Ben-Zahler, line 13: do these changes impact the behaviour of the "existing" notification bar? That may be already in use , if our new designs collide with the existing implementation, this must be discussed.

bozojovicic: @Ben-Zahler 
If none of the top-content-bar attributes are set the only difference between new and old top-content-bar is the close functionality. The close icon is displayed and of course once you close it it will remain closed during that session.

Screenshot 2021-12-28 at 11 21 27

If we should not display the close button for existing instances of top-content-bar then we need to know when to display it. In that case it will be hidden by default and we need some new attribute like "closable=true"
Ok like that or any other idea?

src/components/20-molecules/top-content-bar/index.js
bozojovicic: @Ben-Zahler 
I tried to add some animation into closing process although that was not the requirement and also it doesn't work for desktop-overlay so I will just remove this timeout and do the simple close.

src/components/20-molecules/top-content-bar/index.scss
Ben-Zahler: that sounds all ver good, I think adding closeable=true is a good idea

bozojovicic: The attribute closeable added

Ben-Zahler (dismissed):

src/components/20-molecules/top-content-bar/README.md
```diff
@@ -1,6 +1,7 @@
 # Top content bar
 
 Used as top of the page warning or info box that can show text and have a call to action button.
+It can also be display at the bottom of the page on mobile devices.
```
markus-walther: Please detail the conditions under which it will switch to bottom display! Here, it would suffice to say "(see below for the _stickymobile_ attribute).".

src/components/20-molecules/top-content-bar/README.md
```diff
@@ -21,6 +22,22 @@ If no attribute `href` is set, it will render an `axa-button` and the `onClick`
 
 The attribute `href` specifies the link, that should be opened by pressing on the button with the text defined in `ctatext`. This is mainly for semantic reasons, you still need to forward the user to the new location, by listening to the `onClick` callback. Just providing the `href` attribute will not do anything by itself.
 
+### icon
+
+Using the string attribute `icon` as icon name, an icon will be rendered. To see the full list of possible icons, see the [axa-icon](https://github.com/axa-ch-webhub-cloud/pattern-library/blob/develop/src/components/10-atoms/icon/README.md) README.
```
markus-walther: Better: string-valued attribute...

src/components/20-molecules/top-content-bar/README.md
```diff
@@ -21,6 +22,22 @@ If no attribute `href` is set, it will render an `axa-button` and the `onClick`
...
+
+### stickymobile
+
+Set the attribute `stickymobile` to `true` to display the bar at the bottom of the screen on mobile devices with width < 411px.
+
+### overlaydesktop
+
+Set the attribute `overlaydesktop` to `true` to display the bar on top of DOM element that comes after this element. Usually that would be the Keyvisual CTA component.
```
markus-walther: better: Boolean attribute.

Grammar: on top of **the** DOM element....

Question: What restrictions come with this feature, why does it have 'desktop' in the name? This may need further explaining here.

src/components/20-molecules/top-content-bar/README.md
```diff
@@ -21,6 +22,22 @@ If no attribute `href` is set, it will render an `axa-button` and the `onClick`
...
+
+### closeable
```
markus-walther: Google search of 'closeable' vs 'closable' shows the latter is far more frequent => adjust spelling!

src/components/20-molecules/top-content-bar/index.js
```diff
@@ -34,6 +40,10 @@ class AXATopContentBar extends InlineStyles {
       ctatext: { type: String },
       href: { type: String },
       variant: { type: String },
+      icon: { type: String },
+      stickymobile: { type: String },
```
markus-walther: Type error: should be **Boolean**

src/components/20-molecules/top-content-bar/index.js
```diff
@@ -34,6 +40,10 @@ class AXATopContentBar extends InlineStyles {
...
+      overlaydesktop: { type: String },
```
markus-walther: ditto

src/components/20-molecules/top-content-bar/index.js
```diff
@@ -34,6 +40,10 @@ class AXATopContentBar extends InlineStyles {
...
+      closeable: { type: String },
```
markus-walther: ditto, plus 'closable' property name change.

src/components/20-molecules/top-content-bar/index.react.d.ts
```diff
@@ -4,7 +4,11 @@ export type Variant = 'warning';
 
 export interface AXATopContentBarProps {
   variant?: Variant;
-  ctatext: string;
+  icon?: string;
+  stickymobile?: string;
```
markus-walther: string -> boolean, I think.

src/components/20-molecules/top-content-bar/index.react.d.ts
```diff
@@ -4,7 +4,11 @@ export type Variant = 'warning';
...
+  overlaydesktop?: string;
```
markus-walther: ditto.

src/components/20-molecules/top-content-bar/index.react.d.ts
```diff
@@ -4,7 +4,11 @@ export type Variant = 'warning';
...
+  closeable?: string;
```
markus-walther: ditto, plus closable.

src/components/20-molecules/top-content-bar/index.js
```diff
@@ -13,13 +14,18 @@ import {
...
+const elRootSelector = '.m-top-content-bar';
+const hideClass = 'hide';
+let elementHidden = true;
 /**
  *  We need InlineStyles to give children some margins.
  */
 class AXATopContentBar extends InlineStyles {
   static get tagName() {
-    return 'axa-top-content-bar';
+    return elTagName;
```
markus-walther: please revert this change - it's best practice to have a readable string here for all our components!

src/components/20-molecules/top-content-bar/index.js
```diff
@@ -13,13 +14,18 @@ import {
 } from '../../../utils/component-versioning';
 import { applyDefaults } from '../../../utils/with-react';
 import styles from './index.scss';
+import fireCustomEvent from '../../../utils/custom-event';
 
+const elTagName = 'axa-top-content-bar';
```
markus-walther: Please delete this line, cf. line 28.

src/components/20-molecules/top-content-bar/index.js
```diff
@@ -46,13 +56,28 @@ class AXATopContentBar extends InlineStyles {
...
     defineVersioned(
-      [AXAButton, AXAButtonLink, AXAContainer],
+      [AXAButton, AXAButtonLink, AXAContainer, AXAIcon],
       __VERSION_INFO__,
       this
     );
   }
 
+  onClose() {
+    const contBarEl = this.shadowRoot.querySelector(elRootSelector);
```
markus-walther: With lit-html/element, we have an _efficient_, reactive HTML (re)rendering library, so we should not resort to CSS-based hiding. It's also bad for mobile performance, where high DOM-node count is critical (in your present solution, no UI is shown when closed, yet all the DOM nodes remain underlyingly).

A better approach (after deleting lines 76-77):
-  define a property `closed: {type: Boolean, reflect: true}`
-  `this.closed = true` here
- in line 131: `..., closed } = this;`
- line 132: `if (closed) return html ``;`

markus-walther: If, using the better approach, one needs to refer to the closed state in component CSS, `:host([closed])` can be used.

See HTML (click on the triangle!) <details> element and its `open` attribute for an analogous case.

src/components/20-molecules/top-content-bar/index.js
```diff
@@ -46,13 +56,28 @@ class AXATopContentBar extends InlineStyles {
...
+    contBarEl.classList.add(hideClass);
+    fireCustomEvent('axa-top-bar-close', null, window);
```
markus-walther: this needs a comment: `// let *others* know that this component is now closed`

And the name should start with the exact component name, so it would be findable via `git grep`: `axa-top-content-bar-close`.

Third, our best practice is to fire this at the host element itself, and let event bubbling take it from there: `fireCustomEvent('axa-top-content-bar-close', null, this);`

src/components/20-molecules/top-content-bar/index.scss
```diff
@@ -1,5 +1,95 @@
...
+
 .m-top-content-bar {
   display: block;
+  line-height: 24px;
+
+  .m-top-content-bar__container {
+    line-height: 24px;
+  }
+
+  &.hide {
```
markus-walther: delete when better approach (see above) is used.

src/components/20-molecules/top-content-bar/index.js
```diff
@@ -103,23 +128,76 @@ class AXATopContentBar extends InlineStyles {
...
+    const variantClasses = {
       'm-top-content-bar__container--warning': variant === 'warning',
+      'm-top-content-bar__container--success': variant === 'success',
+      'm-top-content-bar__container--attention': variant === 'attention',
+    };
+
+    const rootClasses = {
+      'm-top-content-bar__overlay-desktop': overlaydesktop === 'true',
+      'm-top-content-bar__bottom-mobile': stickymobile === 'true',
+      hide: elementHidden,
```
markus-walther: I suggest to delete this line, and use the better approach instead.

src/components/20-molecules/top-content-bar/index.js
```diff
@@ -103,23 +128,76 @@ class AXATopContentBar extends InlineStyles {
...
 
-    const classes = {
+    const variantClasses = {
       'm-top-content-bar__container--warning': variant === 'warning',
+      'm-top-content-bar__container--success': variant === 'success',
+      'm-top-content-bar__container--attention': variant === 'attention',
+    };
+
+    const rootClasses = {
+      'm-top-content-bar__overlay-desktop': overlaydesktop === 'true',
```
markus-walther: After making overlaydesktop a Boolean property, you should simplify this to `overlaydesktop`.

src/components/20-molecules/top-content-bar/index.js
```diff
@@ -103,23 +128,76 @@ class AXATopContentBar extends InlineStyles {
...
+      'm-top-content-bar__bottom-mobile': stickymobile === 'true',
```
markus-walther: ditto for stickymobile.

src/components/20-molecules/top-content-bar/index.js
```diff
@@ -103,23 +128,76 @@ class AXATopContentBar extends InlineStyles {
...
+      hide: elementHidden,
+    };
+
+    const iconClasses = {
+      'hide-icon': !icon,
+    };
+
+    const closeClasses = {
+      'hide-close': closeable !== 'true',
```
markus-walther: ditto, `!closable`.

src/components/20-molecules/top-content-bar/index.js
```diff
@@ -103,23 +128,76 @@ class AXATopContentBar extends InlineStyles {
...
+            </div>
+        
+            <div class="m-top-content-bar__close-button ${classMap(
+              closeClasses
+            )}" @click="${this.onClose}">
+              <axa-icon icon="close"></axa-icon>
             </div>
-          </axa-container>
+            ${
+              closeable !== 'true'
```
markus-walther: Simplified and more idiomatic: `closable ? html ``: html ``<div class="m-top-content-bar-empty"></div>`.

src/components/20-molecules/top-content-bar/index.scss
```diff
@@ -1,5 +1,95 @@
...
+  }
+
+  &.m-top-content-bar__overlay-desktop,
+  &.m-top-content-bar__overlay-desktop.m-top-content-bar__bottom-mobile {
+    @include breakpoint($mediaquery-mobile-up) {
+      position: relative;
+      top: 60px;
+      max-height: 0px;
+      width: 100%;
+      width: -moz-available;
```
markus-walther: What's the purpose of lines 23-25?? I would suggest to either drop them, or add justifying comments.

src/components/20-molecules/top-content-bar/index.scss
```diff
@@ -1,5 +1,95 @@
...
+      width: -webkit-fill-available;
+      width: stretch;
+    }
+    z-index: 2;
+  }
+
+  &.m-top-content-bar__bottom-mobile:not(.m-top-content-bar__overlay-desktop) {
+    @include breakpoint($mediaquery-mobile-up) {
+      position: static;
```
markus-walther: This is the first use of `position: static` in the entire pattern-library. Why needed?

src/components/20-molecules/top-content-bar/index.scss
```diff
@@ -9,18 +99,20 @@
 
     min-height: 50px;
     vertical-align: middle;
-
-    text-align: center;
-
-    padding-top: 5px;
-    padding-bottom: 5px;
+    padding: 16px 16px;
```
markus-walther: Isn't that equivalent to the shorter: `padding: 16px` ?!

src/components/20-molecules/top-content-bar/index.scss
markus-walther: I notice @bozojovicic that you haven't really answered @Ben-Zahler 's original question here. You are deleting CSS that presumably served some purpose. Can you demonstrate why it has no impact on the look'n feel of the original top-content-bar? Or is this change blessed by the designers somehow? Or are you silently fixing some previously undetected shortcomings? Please address these questions here, so we know ;-)

src/components/20-molecules/top-content-bar/story.js
```diff
@@ -7,7 +7,32 @@ import readme from './README.md';
 
 const variantOptions = {
   none: '',
-  warning: 'warning',
+  success: 'success',
+  attention: 'attention',
+  'warning / error': 'warning',
+};
+
+const stickyMobileOptions = {
```
markus-walther: A bit overkill perhaps to name a single-use options bundle, but OK...

src/components/20-molecules/top-content-bar/story.js
```diff
@@ -29,16 +54,38 @@ export const TopContentBar = () => {
...
   const href = text('href', '');
   const textValue = text(
     'Text',
     'Unidentified flying object detected in your region. People are panicking. Stay calm!'
   );
   const link = text('Add axa-link', '');
 
+  setTimeout(() => {
+    if (!window.sessionStorage.getItem('top-content-bar-closed')) {
+      window.dispatchEvent(new window.Event('axa-top-bar-open'));
```
markus-walther: The mantra for components is: Props down, events up. So, I suggest to invert the sessionStorage test, and if the top-content-bar-closed item exists in it, say this **after render(template,wrapper)** (assuming the 'better approach'):
`wrapper.querySelector('axa-top-content-bar').closed = true;`.

src/components/20-molecules/top-content-bar/index.js
```diff
@@ -46,13 +56,28 @@ class AXATopContentBar extends InlineStyles {
...
 
+  onClose() {
+    const contBarEl = this.shadowRoot.querySelector(elRootSelector);
+    contBarEl.classList.add(hideClass);
+    fireCustomEvent('axa-top-bar-close', null, window);
+  }
+
   firstUpdated() {
     this.inlineStyles('childStyles');
     const links = Array.prototype.slice.call(this.querySelectorAll('axa-link'));
```
markus-walther: In 2022, we don't need to take care of IE11 anymore, so could do an en-passant legacy fix here:
`const links = [...this.querySelectorAll('axa-link')];`

src/components/20-molecules/top-content-bar/index.scss
```diff
@@ -1,5 +1,95 @@
+$mediaquery-mobile-up: 411px;
+$mediaquery-tablet-up: 991px;
```
markus-walther: New breakpoints and/or bypassing the existing SASS definitions need extra justifications.

Here, your 991px breakpoint is just 1 px off the existing $mediaquery-md-up, so the latter should be used.

src/components/20-molecules/top-content-bar/index.scss
```diff
@@ -1,5 +1,95 @@
+$mediaquery-mobile-up: 411px;
```
markus-walther: Has this new breakpoint been discussed with the designers? The existing system, blessed by designers at the time AFAIK, has as smallest breakpoint $mediaquery-xs-up: // (min-width: 576px). What motivates going smaller here?

Ben-Zahler:
@amir-ag @markus-walther can you review this? thanks!

src/components/20-molecules/top-content-bar/index.scss
```diff
@@ -56,9 +156,52 @@
...
+}
 
 .m-top-content-bar__container--warning {
   background-color: $color-shy-tomato;
 }
+.m-top-content-bar__container--success {
+  background-color: $color-malachite;
+}
+.m-top-content-bar__container--attention {
+  background-color: #BC9D45;
```
markus-walther: Best practice is to use only symbolic colours. So, if you can find a nearby existing colour in 00-colors.scss, use it. If not, and the new colour is designer-approved, it would be best to add it to the /// Background and Illustration section of said 00-colors.scss (consider a name and enter it to https://colornames.org/color/bc9d45, perhaps).

src/components/20-molecules/top-content-bar/README.md
```diff
@@ -21,6 +22,22 @@ If no attribute `href` is set, it will render an `axa-button` and the `onClick`
 
 The attribute `href` specifies the link, that should be opened by pressing on the button with the text defined in `ctatext`. This is mainly for semantic reasons, you still need to forward the user to the new location, by listening to the `onClick` callback. Just providing the `href` attribute will not do anything by itself.
 
+### icon
+
+Using the string attribute `icon` as icon name, an icon will be rendered. To see the full list of possible icons, see the [axa-icon](https://github.com/axa-ch-webhub-cloud/pattern-library/blob/develop/src/components/10-atoms/icon/README.md) README.
+
+### stickymobile
+
+Set the attribute `stickymobile` to `true` to display the bar at the bottom of the screen on mobile devices with width < 411px.
```
markus-walther: better: Boolean attribute

src/components/20-molecules/top-content-bar/index.scss
```diff
@@ -1,5 +1,95 @@
...
+
+    @include breakpoint($mediaquery-mobile-up) {
+      position: unset;
+      top: 0;
+      bottom: unset;
+    }
+  }
+
+  .m-top-content-bar__content-text {
+    margin: 1em 0px;
```
markus-walther: Can simplify 0px to 0.

src/components/20-molecules/top-content-bar/index.scss
```diff
@@ -1,5 +1,95 @@
...
+
+  .m-top-content-bar__container {
+    line-height: 24px;
+  }
+
+  &.hide {
+    display: none;
+  }
+
+  &.m-top-content-bar__overlay-desktop,
```
markus-walther: The README talks 'on top of DOM element that comes after this element' - how is that guaranteed here, what's the mechanism? Looking at the CSS that implements this, I don't understand how it works.

src/components/20-molecules/top-content-bar/README.md
```diff
@@ -21,6 +22,22 @@ If no attribute `href` is set, it will render an `axa-button` and the `onClick`
...
+
+### overlaydesktop
+
+Set the attribute `overlaydesktop` to `true` to display the bar on top of DOM element that comes after this element. Usually that would be the Keyvisual CTA component.
+
+### closeable
+
+Set the attribute `closeable` to `true` to display the Close button. Once the bar is closed it will not be displayed any more during that browser session.
+
```
markus-walther, line 40: When following the 'better approach' (see below), describe an additional Boolean property/attribute `closed` here.

src/components/20-molecules/top-content-bar/child.scss
```diff
@@ -1,5 +1,17 @@
-axa-top-content-bar > axa-link {
-  padding: 5px;
+axa-top-content-bar {
+  axa-link {
```
markus-walther: Why no longer immediate-dominance '>' ?

src/components/20-molecules/top-content-bar/child.scss
```diff
@@ -1,5 +1,17 @@
...
+    padding: 5px;
+  }
+
+  a {
```
markus-walther, line 6: What motivates this additional child CSS? Could it break backwards compatibility?

src/components/20-molecules/top-content-bar/child.scss
```diff
@@ -1,5 +1,17 @@
...
+    color: white;
```
markus-walther: That's not in our 0-colors.scss, but $color-prim-white is.

src/components/20-molecules/top-content-bar/index.js
```diff
@@ -46,13 +56,28 @@ class AXATopContentBar extends InlineStyles {
     super();
     applyDefaults(this);
 
+    window.addEventListener('axa-top-bar-open', () => {
```
markus-walther: Using the better approach discussed at line 76, this event listener seems superfluous. top-component-bar-open is the default **state** of the component, doesn't feel semantically correct to model it as an **event**. I suggest to delete it.

src/components/10-atoms/icon/index.js
```diff
@@ -14,6 +14,11 @@ import {
   File_uploadSvg,
   Cloud_uploadSvg,
   Check_circleSvg,
+  CheckSvg,
+  Info_outlineSvg,
+  WarningSvg,
+  Cloud_queueSvg,
+  CloseSvg,
```
markus-walther, line 21: Question: Does this increase the bundle size wherever axa-icon is used? By how much?

markus-walther: I asked because axa-icon also allows bring-your-own SVGs (see section 'Icon as a SVG string' of the README), and so the alternative would be to put these new icons into axa-top-content-bar (which is rarely used). Knowing the size increase, if any, would help make a rational decision here.

markus-walther (changes requested):
Hello @bozojovicic ,

thanks a lot for your first contribution to the pattern library!

Found a bunch of things, not unexpected for first time - so, don't worry and don't take it personal ;-)

Feel free to ask me on Slack or make a meeting if things are too unclear anywhere, although I tried my best.

Let's continue to collaborate to get this merged soon!

Cheers, Markus

src/components/20-molecules/top-content-bar/index.scss
bozojovicic: @markus-walther 
In the design https://www.figma.com/file/6zurYk3bJpzUg0H2THSxGF/AXA-UI-Kit?node-id=8744%3A5711 we have breakpoint: mobile < 411px and tablet < 991px
I agree that I can use existing $breakpoint-md:992px;
but for mobile I had to set the new breakpoint.
Should I use existing $breakpoint-xs:576px; instead of 411px ?

src/components/20-molecules/top-content-bar/index.scss
bozojovicic: According to https://www.figma.com/file/6zurYk3bJpzUg0H2THSxGF/AXA-UI-Kit?node-id=8744%3A5711 the bar will be placed after the header. The top content bar with default configuration would be just displayed there and push the other content down. But if "overlay on dekstop" is enabled the bar would be just displayed over the component that comes after the header. That is usually some hero component, or keyvisual.
Here are some examples. Imagine that this is our page without the bar displayed
page
When bar is displayed with overlaydesktop=false it will be displayed between the header and subheader
page with bar
When bar is displayed with overlaydesktop=true it will be displayed after the header but on top of the subheader so when we close the bar the page will look exactly as on the first image without pushing any content up or down.
page with bar overlayed

I'm also not happy with the CSS code that implements this. When I push the new code I will come back to this comment.

src/components/10-atoms/icon/index.js
bozojovicic: I measured the size of icon bundle with and without new icons
/src/components/10-atoms/icon/dist/index.js

- from develop branch 56391 byte
- from this branch with all new icons 57693 bytes

bozojovicic: @markus-walther let me know if this is OK or I should use <svg> directly
I also don't understand how using <svg> will reduce the build size when at the end the whole code ends up in the same build file /etc/clientlibs/axa/hub/podaempatternslibrary.js
Using separated <svg> strings would actually increase the size of this file at the end (in my opinion).

src/components/20-molecules/top-content-bar/index.scss
bozojovicic: After all the changes I made they are useless. Will remove them.

src/components/20-molecules/top-content-bar/index.scss
bozojovicic: Absolutely useless now. Will remove it.

src/components/20-molecules/top-content-bar/story.js
bozojovicic: Could not decide if I should name it warning or error :) Will rename it to simple warning

src/components/20-molecules/top-content-bar/index.scss
bozojovicic: I found here, and only here, https://hexcolorpedia.com/color/bc9d45/ that this color is called "underbrush" so I will add  $color-underbrush: #bc9d45 to colors.scss

src/components/20-molecules/top-content-bar/index.js
bozojovicic: I did it as described in the first comment and it works fine. The 2nd comment totally confuses me, I don't see why we need :host([closed]) when it works well also without it.

src/components/20-molecules/top-content-bar/index.scss
bozojovicic: Top content bar used to be a super-simple component that displays a text and a button. Now it has the close icon, the left icon, both of them are optional and it also has different design for mobile devices. I had to change some of the existing code to merge the new functionality into the existing component. But the text is still center aligned, not with text-align:center, but with justify-content:center, but visually the text is centered as requested in the design. Also per the new design the component is now 64px high and also the font-size is 18px, Ben said that is agreed with the designer. All in all I had to change some of the old code to support the new design.

src/components/20-molecules/top-content-bar/README.md
bozojovicic: I did my best to explain it in the updated description.

src/components/20-molecules/top-content-bar/README.md
bozojovicic: I searched for the same thing and I've seen that both are used but did not check which version is used more. Anyway now it is "closable".

src/components/20-molecules/top-content-bar/child.scss
bozojovicic: No idea why I changed this. Rollbacked.

src/components/20-molecules/top-content-bar/child.scss
bozojovicic: The text displayed in the bar is entered in page properties in rich text area where the user can also add the links. That link is not axa-link but the ordinary a[href] so I had to add some CSS here to make it look like the link in the design. This will not affect the axa-link-aem child components, if there are any.

src/components/20-molecules/top-content-bar/index.js
bozojovicic: I have to explain couple of things. The component's visibility depends on the sessionStorage ie. once the user closed it, we need to set some flag in sessionStorage to keep it closed during that browser session. We can access sessionStorage only from the axa-ch-core module where we listen for the event 'axa-top-content-bar-close' to set the sessionStorage flag
https://github.com/axa-ch-webhub-cloud/axa-ch-core/pull/76
The bar is initially opened, by default. We don't have any requirement if the bar should be initially opened or closed (before we close/open it with new "closed" property) so I added one more boolean attribute **initiallyclosed** to make this configurable.

src/components/20-molecules/top-content-bar/index.js
bozojovicic: reverted

src/components/20-molecules/top-content-bar/index.js
bozojovicic: deleted

src/components/20-molecules/top-content-bar/index.js
bozojovicic: Event names are now 'axa-top-content-bar-open' and 'axa-top-content-bar-close'.
I added couple of comments in this method to explain why these events are used.
I changed **window** with **this** for the events.

src/components/20-molecules/top-content-bar/index.scss
bozojovicic: **hide** deleted

src/components/20-molecules/top-content-bar/index.js
bozojovicic: done

src/components/20-molecules/top-content-bar/index.js
bozojovicic: done

src/components/20-molecules/top-content-bar/index.scss
bozojovicic: yes :) fixed

src/components/20-molecules/top-content-bar/story.js
bozojovicic: I explained in one of my previous comments why we need to use events to open or close the bar from axa-ch-core module

src/components/20-molecules/top-content-bar/index.js
bozojovicic: done

src/components/20-molecules/top-content-bar/index.scss
bozojovicic: done

src/components/20-molecules/top-content-bar/index.scss
bozojovicic: fixed

src/components/20-molecules/top-content-bar/child.scss
bozojovicic: fixed

src/components/20-molecules/top-content-bar/index.js
bozojovicic: I changed stickymobile, overlaydesktop and closable to Boolean

src/components/20-molecules/top-content-bar/index.js
bozojovicic: Event name is now 'axa-top-content-bar-close'.
Changed 'window' with 'this'
Added a detailed comment to onClose() method.

src/components/20-molecules/top-content-bar/story.js
bozojovicic: Done

src/components/20-molecules/top-content-bar/README.md
bozojovicic: Done

src/components/20-molecules/top-content-bar/index.js
bozojovicic: And yes, I deleted the 'axa-top-bar-open' event and I use the "closed" property instead to show/hide the bar

bozojovicic:
@markus-walther @Ben-Zahler I think I answered and fixed all comments by now. For couple of things I first answered that it cannot be fixed because of this and that but in all such situations at the end I realized that it actually can be done exactly as Markus explained so I did it. I hope I deleted all such "can't do it" comments.
If you find some comments where I did not answer anything you can count that I fixed it. Just didn't want to write DONE 30 times.
One thing that is out of my power are these tests that fail randomly. After every commit I checked and every time some different test is failing. When I run them locally nothing fails.

src/components/00-materials/styles/00-colors.scss
```diff
@@ -33,6 +33,7 @@ $color-tosca: #914146 !default;
 $color-azalea: #e196aa !default;
 $color-cotton-candy: #fad6de !default;
 $color-apache: #ddbe65 !default;
+$color-underbrush: #bc9d45 !default;
```
markus-walther, line 36: Good!

src/components/10-atoms/icon/index.js
markus-walther: OK, so that's a 2.3 % increase or + 1.27 KB. That's acceptable to me - thanks for supplying the data, so a data-driven decision is possible!

On the podaempatternslibrary.js question: this bundle is just an artifact of our current inability to deliver truly component-by-component bundles, which is a planned improvement to further reduce page weight. Then, each per-component bundle matters separately. Plus, and this is what was probably missing in your understanding, the icon bundle can get independently bundled in pods, which are not forced to use the same version from podaempatternslibrary.js. So, one can easily have multiple copies of the icon bundle on 1 page!

src/components/20-molecules/top-content-bar/README.md
```diff
@@ -1,6 +1,7 @@
 # Top content bar
 
 Used as top of the page warning or info box that can show text and have a call to action button.
+It can also be display at the bottom of the screen on mobile devices (see below for the stickymobile attribute).
```
markus-walther, line 4: Good!

src/components/20-molecules/top-content-bar/README.md
```diff
@@ -21,6 +22,30 @@ If no attribute `href` is set, it will render an `axa-button` and the `onClick`
 
 The attribute `href` specifies the link, that should be opened by pressing on the button with the text defined in `ctatext`. This is mainly for semantic reasons, you still need to forward the user to the new location, by listening to the `onClick` callback. Just providing the `href` attribute will not do anything by itself.
 
+### icon
+
+Using the string-valued attribute `icon` as icon name, an icon will be rendered. To see the full list of possible icons, see the [axa-icon](https://github.com/axa-ch-webhub-cloud/pattern-library/blob/develop/src/components/10-atoms/icon/README.md) README.
+
+### stickymobile
+
+Set the boolean attribute `stickymobile` to `true` to display the bar at the bottom of the screen on mobile devices with width < 411px.
```
markus-walther: Spelling: boolean -> Boolean (after Georg Boole, i.e. an adjective derived from a _name_). Please change everywhere.

src/components/20-molecules/top-content-bar/README.md
```diff
@@ -21,6 +22,30 @@ If no attribute `href` is set, it will render an `axa-button` and the `onClick`
...
+
+Set the boolean attribute `overlaydesktop` to `true` to display the bar on top of the DOM element that comes after this element. Usually that would be the Keyvisual CTA component. This feature applies only for the screen width larger than 411px. For mobile devices the bar will either be displayed as usually with `position:static` or at the bottom of the screen if the attribute `stickymobile` is enabled.
+
+### closable
+
+Set the boolean attribute `closable` to `true` to display the Close button. Once the bar is closed it will not be displayed any more during that browser session.
+
+### initiallyclosed
+
+Set the boolean attribute `initiallyclosed` to `true` to make the bar is initially closed (on page load). By default it is initially opened ie. displayed. If the bar is initially closed it will be displayed after the check that the bar is not closed by the user during that browser session. Also if the bar is initially opened it will be automatically closed after the check that the bar is closed by the user during that browser session.
```
markus-walther: Grammar: "to make the bar is initially closed (on page load)." -> "to force the notification bar to appear closed on page load."

markus-walther: I honestly don't understand the 2 sentences that follow. Please rework or drop them if not truly necessary.

Since closed state seems to sync with session state, I also have questions regarding the UX (flicker? how can users understand a closed bar coming back to life without user action, isn't the only sensible transition open->closed?).

src/components/20-molecules/top-content-bar/index.js
```diff
@@ -47,15 +55,25 @@ class AXATopContentBar extends InlineStyles {
...
+    // sessionStorage can be used only in the core module, so the event will be
+    // fired here and handled in core module by adding the 'top-content-bar-closed'
+    // item to sessionStorage.
+    fireCustomEvent('axa-top-content-bar-close', null, this);
+  }
+
   firstUpdated() {
     this.inlineStyles('childStyles');
-    const links = Array.prototype.slice.call(this.querySelectorAll('axa-link'));
+    const links = [...this.querySelectorAll('axa-link')];
```
markus-walther, line 74: Good!

src/components/20-molecules/top-content-bar/index.js
```diff
@@ -103,23 +121,88 @@ class AXATopContentBar extends InlineStyles {
...
+    const {
+      variant,
+      icon,
+      stickymobile,
+      overlaydesktop,
+      closable,
+      closed,
+    } = this;
+
+    if (!closed) {
```
markus-walther: Now - from reading lines 133-137 I can understand initiallyclosed (which does not participate in any logic elsewhere) - and prove that it is _not_ needed.

Here is a truth table of `closed` and the consequences for `render`:

closed   render
-----------------
false       content
true        empty

Thus, initiallyclosed can be eliminated in favour of <axa-top-content-bar closed ...>. 

Is there any situation where this alternative expression of the initially-closed state cannot be used?

If not, please delete initiallyclosed in code and documentation.

src/components/20-molecules/top-content-bar/index.scss
markus-walther: I can't say what motivated the designers to specify an even smaller mobile breakpoint - can you ask if that was deliberate or an oversight?

If oversight, use $breakpoint-xs, otherwise your new breakpoint (which you'd keep local to this component).

Please record the designer's decision in the PR description, so we know.

src/components/20-molecules/top-content-bar/index.scss
markus-walther: I understand it better now. But technically, nothing guarantees the overlay property, as seen in this screenshot where I added some CSS to the subheader:
Screen Shot 2022-01-06 at 09 38 13
:

markus-walther: Apart from the problem with the overlay implementation, a more fundamental question is why a PL component should be given the power to suppress another, neighbouring component - doesn't this break the fundamental contract of a "component", to not draw outside of the box it has been given?

src/components/20-molecules/top-content-bar/README.md
markus-walther: (See below discussion in the code where this is used for why I think this can be eliminated alltogether).

src/components/20-molecules/top-content-bar/README.md
```diff
@@ -21,6 +22,30 @@ If no attribute `href` is set, it will render an `axa-button` and the `onClick`
...
+
+Using the string-valued attribute `icon` as icon name, an icon will be rendered. To see the full list of possible icons, see the [axa-icon](https://github.com/axa-ch-webhub-cloud/pattern-library/blob/develop/src/components/10-atoms/icon/README.md) README.
+
+### stickymobile
+
+Set the boolean attribute `stickymobile` to `true` to display the bar at the bottom of the screen on mobile devices with width < 411px.
+
+### overlaydesktop
+
+Set the boolean attribute `overlaydesktop` to `true` to display the bar on top of the DOM element that comes after this element. Usually that would be the Keyvisual CTA component. This feature applies only for the screen width larger than 411px. For mobile devices the bar will either be displayed as usually with `position:static` or at the bottom of the screen if the attribute `stickymobile` is enabled.
```
markus-walther: Grammar, wording: "This feature applies only **to** non-mobile screen widths." + usually -> usual

src/components/20-molecules/top-content-bar/index.scss
```diff
@@ -56,9 +139,52 @@
...
+}
 
 .m-top-content-bar__container--warning {
   background-color: $color-shy-tomato;
 }
+.m-top-content-bar__container--success {
+  background-color: $color-malachite;
+}
+.m-top-content-bar__container--attention {
+  background-color: $color-underbrush;
```
markus-walther, line 177: Good!

src/components/20-molecules/top-content-bar/index.scss
markus-walther: One could argue that the orchestration of such suppression of neighbouring components belongs one level higher, to page CSS - and that it has no place in component CSS, since arguably a generic top-content-bar component should not encode narrowly specific usage scenarios.

markus-walther (changes requested):
Hey Bojo / Ben,

great progress!

Some issues remain, the most fundamental one being the question of whether overlaydesktop is a good idea.

Looking forward to the final version, hopefully soon!

--Markus

src/components/20-molecules/top-content-bar/README.md
bozojovicic: Fixed

src/components/20-molecules/top-content-bar/README.md
bozojovicic: I removed the initiallyclosed attribute. "closed" is used now for this

src/components/20-molecules/top-content-bar/index.js
bozojovicic: initiallyclosed removed

src/components/20-molecules/top-content-bar/README.md
bozojovicic: I removed the "overlaydesktop" totally from the code. From JS, CSS and readme. Only stickymobile remains. If I understood it well, you agreed with Ben that this should be implemented in axa-ch-core CSS.

bozojovicic:
@markus-walther I don't understand why I can't reply on some comments so I will write it here. I removed initiallyclosed and overlaydesktop "features" from code. @Ben-Zahler will ask the designer if mobile breakpoint 411px is mandatory or we can use the standard $breakpoint-xs:576px;
I think all other comments are fixed/answered.

bozojovicic:
@markus-walther @Ben-Zahler the designer said that we can use the common mobile breakpoint.
The one for 411px is deleted.

src/components/20-molecules/top-content-bar/README.md
markus-walther: Great!

src/components/20-molecules/top-content-bar/README.md
markus-walther: "you agreed with Ben that this should be implemented in axa-ch-core CSS." Yes, that was the recommendation.

markus-walther:
> @markus-walther @Ben-Zahler the designer said that we can use the common mobile breakpoint. The one for 411px is deleted.

Great.

markus-walther:
> @markus-walther I don't understand why I can't reply on some comments so I will write it here. I removed initiallyclosed and overlaydesktop "features" from code.

That's great.

src/components/20-molecules/top-content-bar/child.scss
markus-walther: Good!

src/components/20-molecules/top-content-bar/index.js
markus-walther: The 2nd comment was proactive - if you don't need it _here_, that's fine.

src/components/20-molecules/top-content-bar/index.scss
markus-walther: Great!

markus-walther (dismissed):

Ben-Zahler (dismissed):

Ben-Zahler (approved):

PR #2291, opened 2022-06-05
Feature/remove babel

MKaHead:
- I was able to reduce the package size of each component by 50%. (Terser 40% and Babel kick 10%)
- Fix for unintended import of `skatejs/val` from our non-wrapped react cps. `applyDefaults` function had to be in a separate file.
- I replaced the rollup sass plugin - now all sass files are resolved accordingly, even nested ones. This allowed the use of index.js for dist build instead of lib/index -> that's not in this PR
- I have removed Babel from our build. With this we drop IE 11 support for the next generation of major version components. The components use ES2019 features at most and linting ensures that no developer accidentally uses ES2020+ features. Now the support looks like this:
  - Chrome >=73 current version 101
  - Safari >=12.1 current version 15.5
  - Firefox >=63 current version 101
  - Edge >=79 current version 101
 - node js upgrade
 - lerna upgrade
 - testcafe upgrade

# Merge Checklist:
- [x] Tests are sufficient.


package.json
```diff
@@ -23,25 +23,25 @@
...
+    "blueimp-canvas-to-blob": "^3.29.0",
+    "date-fns": "^2.28.0",
     "lit": "^2.2.1"
   },
   "devDependencies": {
-    "@axa-ch/materials": "latest",
     "@babel/core": "^7.3.3",
-    "@babel/eslint-parser": "^7.17.0",
     "@babel/plugin-transform-runtime": "^7.4.0",
     "@babel/preset-env": "^7.3.1",
```
amir-ag, line 41: are these babel dependencies still necessary?

amir-ag (dismissed):

package.json
MKaHead: good catch, some of them are not needed anymore

amir-ag (approved):

PR #2303, opened 2022-06-27
Feature/accordion

### LINKED ISSUE #2260: Accordeon - missing dependencies
I spent some time with your Accordeon component (btw. I really believe, that it should be called Accordion in english, not Accordeon 😄  ). I had issue with property `small` which change title size. It didn't work in myaxa project, but it worked in your storybook. We didn't use `@axa-ch/text` in myaxa project and that was the issue. I think you should mention in readme that this package is required and/or add axa text to depedencies array in package.json (you used axa text and materials in source code, but they are not mentioned in package.json as depedencies).

## Expected Behavior

Add `@axa-ch/text` and `@axa-ch/materials` as dependency of `@axa-ch/accordeon` component.

## Current Behavior

If you don't have `@axa-ch/text` in your project, Accordeon won't work correctly.

## Steps to Reproduce

mentioned above

## Context (Environment)

mentioned above; Issue was found in pod-myaxa.

## Possible Solution

mentioned above

### LINKED ISSUE #2279: Accordeon - wrong content height when changing viewports
When you open Accordeon and then change size of window, then some content is missing because height of content didn't change. Also height is calculated too soon - our styles are probably not considered because it will cut out some content. (see screenshots in slack channel or let me know and I'll show you. I can't add them here from security reasons)

## Expected Behavior

Content should be still visible when user change window size.

## Current Behavior

Content is cut when changing from bigger viewport to smaller.

## Steps to Reproduce

1. Go to your storyboook: https://patterns.axa.ch/?path=/story/components-accordeon--accordeon 
2. Open any accordeon.
image

3. Change viewport to XS
image

4. Some content is missing now:
image

5. Close and open Accordeon again and it appears.
image

## Context (Environment)

Use implemented your component at myaxa and now lot of content is missing sometimes. Interesting is that on "tablet" viewport is from page init (I don't have to change viewport). **I guess there is another issue and content height is calculated before content renders.**

## Possible Solution

- recalculate content height when viewport changes ( maybe event listnener on `resize` event? I didn't use it yet )
- wait with calculation of content so it's fully loaded. it's possible that accordeonContent.scrollHeight is evaluated before content is fully rendered/our styles are applied.

### LINKED ISSUE #2290: axa-accordeon: wrong hight in certain cases because width changes when opened.
Wrong hight of accordion in certain cases because width changes when opened. Therefore some content is sometimes not visible.

image

## Expected Behavior
Hight of accordeon-content slot must match content hight.

## Current Behavior
When the accordeon is opened, the margin of the contents changes from margin: auto; to margin: 0px 4px 16px.

Because the width of the content changes by -8px the content might use more vertical space (e.g. text flow to a new line). But the hight of the accordeon-content slot is not changing, some content is not visible.

## Steps to Reproduce
1. Put text inside an accordeon that wraps if the accordeon width is reduced by 8px (see example screenshot)
2. Open the accordeon --> last line of text is not visible

## Context (Environment)
Issue potentially occurs on any OS / Browser. I tested on Edge on Windows (AXA Switzerland).

## Possible Solution
- Preferred solution: do not add margin to the accordeon-content when it is opened. This should be an easy fix.
- Alternative solution: dynamically recalculate the hight of the accordeon-content slot when it is opened ( might be fixed together with https://github.com/axa-ch-webhub-cloud/pattern-library/issues/2279 )

MKaHead:
Should also fix these issues:
fix #2290
fix #2279
fix #2260

- Rename Accordeon to Accordion that now a new component
- When the window is resized, the height of the container is recalculated and set in css
- Remove axa-text use styles for texts 
- Add axa materials to dependencies
- Adapt Variant `small` now default instead of `large` -> see figma
- Remove ie11 hack

# Merge Checklist:
- [ ] Code selfreview has been performed ([Best Practices](https://github.com/axa-ch-webhub-cloud/pattern-library/blob/develop/CONTRIBUTION.md#best-practices)).
- [ ] Tests are sufficient.
- [ ] Dependencies to other components still work.


src/components/20-molecules/accordion/index.js
```diff
@@ -8,12 +7,13 @@ import { classMap } from 'lit/directives/class-map.js';
...
 import { sanitizeSVG } from '../../../utils/sanitize';
+import debounce from '../../../utils/debounce';
 
 import styles from './index.scss';
 
-class AXAAccordeon extends LitElement {
+class AXAAccordion extends LitElement {
   static get tagName() {
-    return 'axa-accordeon';
+    return 'axa-accordion';
```
markus-walther, line 14: This requires a **big fat warning** when announcing the new version on Slack's patterns-lib-devs!!! What if somebody has baked the old name into his app?

Also, we need to release as semver major version.

src/components/20-molecules/accordion/index.js
```diff
@@ -24,7 +24,7 @@ class AXAAccordeon extends LitElement {
 
   static get properties() {
     return {
-      small: { type: Boolean },
+      large: { type: Boolean },
```
markus-walther: Another backwards-incompatible change = double reason for big fat warning.

src/components/20-molecules/accordion/index.js
```diff
@@ -35,6 +35,7 @@ class AXAAccordeon extends LitElement {
   constructor() {
     super();
     applyDefaults(this);
+    this.debouncedHandleViewportCheck = debounce(this.toggleAnimation, 500);
```
markus-walther: Requires a comment saying what problem is solved here by debouncing and explaining the magical 500 milliseconds!

src/components/20-molecules/accordion/package.json
```diff
@@ -1,9 +1,9 @@
 {
-  "name": "@axa-ch/accordeon",
-  "version": "3.0.0",
-  "description": "The accordeon component for the AXA Pattern Library",
+  "name": "@axa-ch/accordion",
+  "version": "1.0.0",
```
markus-walther, line 3: Ah, I see. You are treating it as a new npm package. That would work ;-)

Can we then mark @axa-ch/accordeon as deprecated on npm? If needed, by publishinhg 3.0.1 with big fat deprecation warning?

src/components/20-molecules/accordion/index.js
markus-walther: this.debouncedHandleViewportCheck is not a great name. That it's debounced can be inspected locally at the method, it's a minor detail.

How about this.recalculateDimensionsOnResize?

markus-walther (changes requested):
Hi Marco,


great work! Some small findings only.

Would it be possible to add a Playwright test demonstrating that resizing works now?

Cheers, Markus

src/components/20-molecules/accordion/index.js
```diff
@@ -46,62 +47,66 @@ class AXAAccordeon extends LitElement {
...
           <div
-            class="m-accordeon__title-container ${classMap(titleClasses)}"
+            class="${classMap(titleContainerClasses)}"
             @click="${this.toggleAccordion}"
           >
-            <div class="m-accordeon__title-container-title">
+            <div class="m-accordion__title-container-title">
               <div class="${iconClass}">${iconHTML}</div>
-              <axa-text variant="${small ? 'size-3' : ''}"> ${title} </axa-text>
+              <span class="${classMap(titleClasses)}">${title}</span>
```
markus-walther, line 98: much leaner, I like it :-)

src/components/20-molecules/accordion/package.json
MKaHead: We can do this easily on npm. [see here](https://docs.npmjs.com/deprecating-and-undeprecating-packages-or-package-versions)

src/components/20-molecules/accordion/index.js
```diff
@@ -35,6 +35,9 @@ class AXAAccordeon extends LitElement {
   constructor() {
     super();
     applyDefaults(this);
+    this.toggleAnimation = this.toggleAnimation.bind(this);
+    // When changing the window size, we have to calculate and set the max-height of m-accordion__content again.
+    this.recalculateDimensionsOnResize = debounce(this.toggleAnimation, 500);
```
markus-walther: // See https://bencentra.com/code/2015/02/27/optimizing-window-resize.html for why we debounce here.

src/components/20-molecules/accordion/index.js
```diff
@@ -46,62 +49,66 @@ class AXAAccordeon extends LitElement {
...
-      '.m-accordeon__content'
+    const accordionContent = this.shadowRoot.querySelector(
+      '.m-accordion__content'
     );
 
+    if (!accordionContent) {
+      return;
+    }
+
     if (this.open) {
```
markus-walther: While youre' at it, you might as well DRYify it more here:

`accordionContent.style.maxHeight = this.open ? `${accordionContent.scrollHeight}px` : 0;`

src/components/20-molecules/accordion/package.json
markus-walther: Learned something new, great!

markus-walther (changes requested):
Hi Marco,

almost there - 2 tiny things, and you're good.

You can self-approve after that.

--Markus

markus-walther (approved):
Hi Marco,

LGTM.

--Markus

PR #2326, opened 2022-08-11
[WIP]Feature/button rework

MKaHead:
Changes:
- Put `button-link` functionality into `button` and remove button link -> later we can deprecate the `button-link`
- Write all test in playwright and sort them
- I have completely revised the clickhandler. The general idea is to add fake button only onclick I need your feedback here @markus-walther
- I also removed inline styles because you can style the :host with `appearance: none` and child css is obsolete, too?

Feedback:
- Actually I wanted to spread some props in lit html, but that is not possible yet. That would also be nice for the stories [see here](https://github.com/lit/lit/pull/1960) 
-  I removed a css test -> Maybe we could add this to the wrapperReact tests, as general?

# Merge Checklist:
- [x] Tests are sufficient.


src/components/10-atoms/button/index.js
```diff
@@ -64,90 +56,25 @@ class AXAButton extends InlineStyles {
...
+  handleClick(ev) {
+    if (this.type !== 'button' && hasShadowDom(this)) {
+      // this button wants to specifically submit a form
+      // climb up the dom to see if we're in a <form>
+      // and if so, then use JS to submit it
+      const form = this.closest('form');
+      if (form) {
+        ev.preventDefault();
+
+        const fakeButton = document.createElement('button');
```
markus-walther, line 68: What about this alternative to lines  68-73: `form[this.type]();` ?!

src/components/10-atoms/button/index.js
```diff
@@ -64,90 +56,25 @@ class AXAButton extends InlineStyles {
...
-          this._observer = new MutationObserver(() => this.attachFakeButton());
-        }
-        this._observer.observe(this, {
-          childList: true,
-        });
-        break;
-      default:
-        break;
+  handleClick(ev) {
+    if (this.type !== 'button' && hasShadowDom(this)) {
```
markus-walther, line 60: You have 3 special cases, that could be handled by **3 early returns**, then it's more readable and less nested:
- type=button
- no shadowDOM
- no form

src/components/10-atoms/button/index.js
markus-walther: In whatever final solution you must link to https://www.tjvantoll.com/2013/01/01/enter-should-submit-forms-stop-messing-with-that/ in a comment.

The implicit-form-submission-via-enter algorithm causes internal synthetic click on a submit button if found. Said button cannot be discovered when hidden under shadowDOM. Therefore, your current solution is incorrect insofar as the fakeButton should _not_ be removed and that the fakeButton should be constructed in firstUpdated() since the algorithm can strike at any time, unconnected to a local click on this instance. Also, the disabled state of such a fakeButton should equal the disabled state of its host, I suspect that a disabled submit button will cause no form submission on Enter (test this!).

markus-walther: An alternative is to _also_ modify axa-input-text: detect Enter keypress event, find closest form, submit  form (maybe with added conditions from the linked article).

markus-walther: Another alternative is to rewrite button(-link) to use NoShadowDOM, and drop all complications like here. Isn't this the simplest,most HTML-conformant approach?

markus-walther (changes requested):
WIP-review:

- I would not drop axa-button-link in customElements.define(....), just share its implementation (using AXAButtonLink extends AXAButton {} hack). Reason: backwards compatibility (so important!)
- the implicit-form-submission-by-Enter aspect needs more work, existing solution has a bug in my view (see discussion)
- I like the unification effort per se!

PR #2342, opened 2022-09-12
Release

### LINKED ISSUE #2230: axa-link reason for onClick's event.preventDefault() ?
Axa-link is preventing default event on click. This is documented.

As this is counter-intuitive (compared to dom's anchor element), I am missing reason for this behavior.

From my consumer point-of-view, this is a downside, because it forces me  to use both `href` attribute (for usability) and call in onClick handler the history.push() again, so that the navigation even works.  (As I am attaching other functionality in onClick handler also).

Ideally, there would no e.preventDefault() be called in axa-link at all. As I can do that myself when I need to. As the event is passed down already since v 4.2.23,  see:

Thanks!

image

### LINKED ISSUE #2265: Radio disabled implies checked
**IS:**
`<Radio disabled={true} checked={false} />` renders a disabled checked radio button

**SHOULD**
`<Radio disabled={true} checked={false} />` should render a disabled *unchecked* radio button

**Demo**
https://patterns.axa.ch/?path=/story/components-radio--radio

image

Maybe related to https://github.com/axa-ch-webhub-cloud/pattern-library/issues/2238

Thanks!

### LINKED ISSUE #2278: axa-file-upload: allow uploading .msg files
We have the [file-upload](https://github.com/axa-ch-webhub-cloud/pattern-library/tree/develop/src/components/20-molecules/file-upload) component in use and want to enable external users/customers to upload emails (.msg)  to us, for example. But no matter whether I leave "allowedFileTypes" empty or enter "application/vnd.ms-outlook", for example, it does not work. Can the component be extended accordingly?

PS: Files with the formats tif or tiff cannot be uploaded either. And why is a png file displayed as a jpg after uploading?

Thx

### LINKED ISSUE #2328: axa-file-upload: fix logic to determine if uploaded file was an image
File-upload molecule cannot handle (non image/* and non application/*) files properly. e.g. text/csv

## Expected Behavior
If I upload a file which is not an image and doesn't contain 'application' in its MIME-Type e.g. "text/csv", I'd like to see the usual paperclip icon in the list of added files. Also, I'd expect the readme to match the change logs description that all filetypes are now allowed (since 5.1.0)

## Current Behavior
Currently the component thinks it has to handle an image since the check for non-image-files looks something like this:
`const isNonImageFile = file.type.indexOf('application') > -1;`
Therefore it wants to create a preview, which obviously doesn't work. 

## Steps to Reproduce
1. Open [Codesandbox](https://codesandbox.io/s/objective-pond-72lk2g?fontsize=14&hidenavigation=1&theme=dark)
2. Try to drag and drop a random csv file into the file-upload component
3. ????
4. Profit

## Possible Solution
fix [this](https://github.com/axa-ch-webhub-cloud/pattern-library/blob/a300c12b25826323d805bc156d4a48c05dedc717/src/components/20-molecules/file-upload/index.js#L535) line

### LINKED ISSUE #2336: Simplify axa heading api
There is only one other font family in the AXA world and no more are likely to be added. Therefore a simplification variant="secondary" to `secondary` boolean would make sense.

MKaHead:
Fixes #2336
Fixes #2278
Fixes #2328
Fixes #2265
Fixes #2265
Fixes #2230

# Merge Checklist:
- [x] Tests are sufficient.

### DIFF
--- a/src/components/00-materials/package-lock.json
+++ b/src/components/00-materials/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@axa-ch/materials",
-  "version": "16.0.4",
+  "version": "16.0.5",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
--- a/src/components/00-materials/package-lock.json
+++ b/src/components/00-materials/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "@axa-ch/materials",
-      "version": "16.0.4",
+      "version": "16.0.5",
       "license": "Copyright 2019 AXA Versicherungen AG",
       "dependencies": {
         "lit": "^2.2.1"
--- a/src/components/00-materials/package.json
+++ b/src/components/00-materials/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@axa-ch/materials",
-  "version": "16.0.4",
+  "version": "16.0.5",
   "description": "The materials component for the AXA Pattern Library",
   "author": "Pattern Warriors",
   "homepage": "https://github.com/axa-ch-webhub-cloud/pattern-library/tree/develop/src/components/00-materials/#readme",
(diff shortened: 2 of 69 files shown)


markus-walther (approved):
Marco,

LGTM.

--Markus

Issue #2373, opened 2022-11-11
Input-Text React-18 can't edit text, cursor jumps to end


MarekLacoAXA:
## Expected Behavior
- type in text, move cursor back, can edit text in the middle or on the start
- https://axa-ch-webhub-cloud.github.io/plib-feature/develop/?path=/story/examples-input-text-react--input-text

## Current Behavior
- when under React-18, cursor jumps to end of text after the keypress
- probably because of a blur/focus cycle
- all versions of axa-input-text are affected
- only React-18 version is affected
- (works with React-17)

## Steps to Reproduce

1. https://axa-ch-webhub-cloud.github.io/plib-feature/develop/?path=/story/examples-input-text-react--input-text
2. type in text, move cursor back, type a char  cursor jumps to end of text

## Context (Environment)
- Chrome/Windows

MKaHead:
@MarekLacoAXA I have looked at it and already have a fix. But we should keep an eye on the input text afterwards to see if the fix has side effects or if there are other problems.

MarekLacoAXA:
Still reproducible when "typing quicky": can by simulated by pressing several (two/three) keys at once repeatedly on the begin of text. Cursor jumps to end. (browser-dependent, doesn't occur in Safari/MacOs, does on Chrome/Win).

MarekLacoAXA:
Just to mention: PRIOR fixing this issue (with the cursor jumping to the end in React-18):
In the following browsers the text field was not unusable at all. The whole content was "randomly jumping" with each keypress:
Apple: Safari/MacOs, Chrome+Safari/iOS+iPadOS.

PR #2374, opened 2022-11-12
Fix input-text cursor jump react 18

MKaHead:
Fix input-text cursor jump react 18 #2373 

# Merge Checklist:
- [x] Tests are sufficient.

### DIFF
--- a/src/components/10-atoms/input-text/CHANGELOG.md
+++ b/src/components/10-atoms/input-text/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 6.3.0
 
-Feat: Add boolean attribute `autocomplete` to put the HTML autocomplete attribute to "on". The default is "off"
+Feat: Add boolean attribute `autocomplete` to put the HTML autocomplete attribute to "on". The default is "off" #2364
 Fixes: react 18 cursor jump #2373
 
 ## 6.2.1
--- a/src/components/10-atoms/input-text/CHANGELOG.md
+++ b/src/components/10-atoms/input-text/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 6.3.0
 
-Add boolean attribute `autocomplete` to put the HTML autocomplete attribute to "on". The default is "off" 
+Feat: Add boolean attribute `autocomplete` to put the HTML autocomplete attribute to "on". The default is "off"
+Fixes: react 18 cursor jump #2373
 
 ## 6.2.1
 
--- a/src/components/10-atoms/input-text/index.js
+++ b/src/components/10-atoms/input-text/index.js
@@ -216,7 +216,7 @@ class AXAInputText extends AXAPopupMixin(NoShadowDOM) {
   }
 
   handleInput(ev) {
-    const { selectionStart } = ev.target;
+    const { selectionStart = null } = ev.target;
 
     // sets this.invalidFormat
     this._formatCurrency(ev.target.value);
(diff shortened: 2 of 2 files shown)


src/components/10-atoms/input-text/index.js
```diff
@@ -215,13 +216,19 @@ class AXAInputText extends AXAPopupMixin(NoShadowDOM) {
   }
 
   handleInput(ev) {
+    this.cursor = ev.target.selectionStart;
```
MarekLacoAXA: I suggest to reuse the existing naming: selectionStart.
(maybe alternative: targetSelectionStart)

```suggestion
    this.selectionStart = ev.target.selectionStart;
```

MarekLacoAXA (commented):
Thanks for fixing this issue! 👍

src/components/10-atoms/input-text/index.js
MKaHead: Thank you, the variable can also be function scoped.

src/components/10-atoms/input-text/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## 6.3.0
 
-Add boolean attribute `autocomplete` to put the HTML autocomplete attribute to "on". The default is "off" 
+Feat: Add boolean attribute `autocomplete` to put the HTML autocomplete attribute to "on". The default is "off" #2364
```
markus-walther, line 3: Interesting - but where is the implementation?

src/components/10-atoms/input-text/index.js
```diff
@@ -216,13 +216,19 @@ class AXAInputText extends AXAPopupMixin(NoShadowDOM) {
...
     // sets this.invalidFormat
     this._formatCurrency(ev.target.value);
     this.onChange(ev, this.invalidFormat);
     // are we a 'controlled' input in the React sense?
     if (this.isControlled) {
       // yes, set UI from model state
       this.nativeInput.value = this.modelValue;
+
+      requestAnimationFrame(() => {
+        this.nativeInput.setSelectionRange(selectionStart, selectionStart);
```
markus-walther, line 230: OK, this is a creative solution, that seems to be usable in all browsers we care about - for the narrow case of React controlled components with a trivial selection. Just curious: what if the selection is a non-trivial range spanning a few letters? Is it acceptable that the range is lost here? (I genuinely don't know).

markus-walther (changes requested):
Hi Marco,

two questions remain, otherwise good work!

Cheers, Markus

src/components/10-atoms/input-text/CHANGELOG.md
MKaHead: I have develop rebased, feature autocomplete is already in develop and affects the same component.

src/components/10-atoms/input-text/index.js
MKaHead: We would have to find out. That's why I wrote Marek directly that we have to observe the component. I think at the moment is the best fix.

MarekLacoAXA:
Do you have an explanation why does this issue happen? What is different in React-18 compared to React-17? 
I would like to understand what is the root cause of this.

src/components/10-atoms/input-text/CHANGELOG.md
markus-walther: I see, good.

src/components/10-atoms/input-text/index.js
markus-walther: Ok, then good for now.

markus-walther (approved):
Hi Marco,

LGTM.

--M.

MarekLacoAXA:
Thanks for fixing Marco!
Works quite good, in 99% cases.
(However I was still able to experience this cursor jump "sometimes": in 1% of keypresses or so.)

MarekLacoAXA:
Still reproducible when "typing quicky": can by simulated by pressing several (two/three) keys at once repeatedly on the begin of text. Cursor jumps to end.

PR #2405, opened 2023-01-30
allow html min and max properties

ahoehn:
Fixes #

# Merge Checklist:
- [x ] Code selfreview has been performed ([Best Practices](https://github.com/axa-ch-webhub-cloud/pattern-library/blob/develop/CONTRIBUTION.md#best-practices)).
- [x ] Tests are sufficient.
- [x ] Documentation is up to date.
- [x ] Changelog is up to date.
- [x ] Typescript typings for properties are up to date.
- [x ] Designers approved changes or no approval needed.
- [ x] Dependencies to other components still work.

### DIFF
--- a/src/components/10-atoms/input-text/CHANGELOG.md
+++ b/src/components/10-atoms/input-text/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 6.5.0
+
+Add HTML `min` and `max`props
+
 ## 6.4.0
 
 Add `onPaste` callback prop #2391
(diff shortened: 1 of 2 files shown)


markus-walther:
Hello Adrian Höhn,

thanks for submitting a PR! 

Can you please open an issue that clearly describes the problem that your PR desires to fix?
Then you can actually fill in the Fixes # ! I think that issue is a Feature request, so the label should be set that way.
Then, you might want to review lit-html template syntax here: https://lit.dev/docs/v1/lit-html/template-reference/#binding-types. Seems your min/max properties are using the syntax for Boolean attributes, which is probably not what you want.
Since you did not add any Playwright tests, you might want to add some to test your new feature - and only then check the checkbox above. Same for README.md documentation.

Thanks for working with us to get this into good shape.

Cheers, Markus

